Issue #10, opened 2016-11-20
Add `curryN` for more explicit partial application 


evilsoft:
While the current recursive `curry` function is great in most cases, it would be nice to have the ability to explicitly specify the expected arity of a function. This is handy for complicated compositions with curried mutli-ary functions.

flintinatux:
Wanted to add some details to this one.  I tried an operation like this on the `ramda` repl, and was disappointed by [the result](https://goo.gl/ObnDGO).  A simplified version of the use case I was aiming for looks like this:
```js
const f = curryN(3, op => (dt, s) => [op, dt, s])
f('del', 250, 'foo')  // => should return ['del', 250, 'foo']
```
The best that `ramda` was able to do was curry the outer function, and once passed three arguments, only return the inner function:
```js
function (dt, s) {
  return [op, dt, s];
}
```

evilsoft:
Submitted a [PR#20](https://github.com/evilsoft/crocks/pull/20) to address this issue.
This function cannot have the magic of `curry` auto currying until a non function value is returned.
Instead it is to be used with the various n-ary functions that exist in the wild.

Please see the PR for some examples of usage.

evilsoft:
Merged in [PR#20](https://github.com/evilsoft/crocks/pull/20).
Closing.

PR #20, opened 2017-01-08
Add `curryN` to the helper functions

evilsoft:
## Limit Dem 

This PR address [issue#10](https://github.com/evilsoft/crocks/issues/10).

It adds a curryN function to limit the number of arguments to a function or deal with N-ary functions or defaulted arguments.

```javascript
// limit to default
const times10 = curryN(1,
  (x, y=10) => x * y
)
times10(23) // 230

// arg spread
const scaleSum = curryN(4,
  (s, ...args) => args.reduce((x, y) => x + y, 0) * s
)
const doubleSum = scaleSum(2)

scaleSum(10, 3)(4, 5) // 120
doubleSum(3, 4)(5) // 24

// N-ary
const maxOf3 = curryN(3, Math.max)
maxOf3(3)(20)(0) // 20
```

This function will not keep currying until a non-function is returned like `curry` does.

```javascript
const f =  curryN(3,
  (x, y) => z => [ x, y, z ]

// the last argument is applied to the outer function and
// therefore thrown away, resulting in the inner function
// being returned.
f(3, 4, 5) // z => ...
```


flintinatux:
I like the look of this.  Since it's OSS, you oughta hook up some free [Travis](https://travis-ci.org/) for your specs to check PR's.

evilsoft:
@flintinatux 🍬 idea!

Issue #14, opened 2016-12-12
Need to correct `first` and `second` for Arrow


evilsoft:
No idea what I was thinking with that, but first and second should NOT take a function. It should take nothing and just apply the wrapped function to either the `first` or `second` on a pair that the Arrow is ran with.

evilsoft:
Will need to implement a Proper `first` and `second` for `Star` once this change gets in.

evilsoft:
Submitted [PR#30](https://github.com/evilsoft/crocks/pull/30) to address this.

evilsoft:
Merged [PR#30](https://github.com/evilsoft/crocks/pull/30). We should be GTG on this one now.

PR #30, opened 2017-01-21
Fix up `Arrow` `first` and `second` and add them to `Star`

evilsoft:
## You are my shining Star (up Star anyway)

This PR addresses [Issue#14](https://github.com/evilsoft/crocks/issues/14) and removes the silly function function params on `first` and `second` functions for `Arrow`. 

Now that they are all fixed up, I also implemented them on `Star`.

:rotating_light: **BONUS** :tada: :boom: 
Added `concat` to `Star` for composition of Stars.


evilsoft:
Also added a the `fanout` function to make it easier to set up `first` and `second` calls.

PR #31, opened 2017-01-22
Adds custom crocks.js graphic and revises part of readme

rstegg:
I felt an issue wasn't necessary since it's so minor, but it adds some :eyes: to the intro of the repo


evilsoft:
Please squash your commits

rstegg:
yessir

rstegg:
moved to cdn for svg, good to go 👍

README.md
```diff
@@ -11,16 +16,16 @@ $ npm install crocks -S
 This will pull down `crocks` into your project's `node_modules` folder and can be accessed by adding something like the following in the file that needs it:
 
 ```javascript
-// node require syntax
+// node require syntax - "es5"
 const crocks = require('crocks')
 
-// Javascript modules (if you are transpiling or using rollup.js)
+// Javascript modules (if you are transpiling or using rollup.js) - "es6"
```
evilsoft: I hate the term ES6. would rather not make a distinction.
It is Javascript Modules.

README.md
```diff
@@ -11,16 +16,16 @@ $ npm install crocks -S
 This will pull down `crocks` into your project's `node_modules` folder and can be accessed by adding something like the following in the file that needs it:
 
 ```javascript
-// node require syntax
+// node require syntax - "es5"
```
evilsoft: It is not really es5, it is common JS modules.
would rather not make distinctions like this at they become brittle  and stale very quickly.

README.md
```diff
@@ -1,5 +1,10 @@
-# crocks.js
-`crocks` is a collection of popular *Algebraic Data Types (ADTs)* that are all the rage in functional programming. You have heard of things like `Maybe` and `Either` and heck maybe even `IO`, that is what these are. The main goal of `crocks` is to curate and provide not only a common interface between each type (where possible of course), but also all of the helper functions needed to hit the ground running.
+<img style="display: block; width: 360px; margin: 24px auto;" src="https://cdn.rawgit.com/rstegg/crocks/dfdee636/crocks_js.svg" />
```
evilsoft: This should be Markdown. If you need to specify styling, then there may be an issue with the asset.
Also the src will not work for someone viewing offline.
TBH, this is basically just a cartoonish heading. Would rather go back to a simple heading. 
This is an information document, not really a marketing page.

README.md
```diff
@@ -1,5 +1,10 @@
...
+
+# a collection of popular *Algebraic Data Types*
+
+The Algebraic Data Types (ADT)s included in `crocks` are those which are trending in functional programming. If you've ever heard of `Maybe` or `Either`, or even `IO`, you're in the right place! The main goal of `crocks` is to provide a *curated common interface* between each type and include all the helper functions needed to get the wheel spinning.
```
evilsoft: I do not understand the need for this rewrite?
Could you provide your reasoning?

README.md
```diff
@@ -11,16 +16,16 @@ $ npm install crocks -S
...
+// Javascript modules (if you are transpiling or using rollup.js) - "es6"
 import crocks from 'crocks'
 ```
 
-There is no compilation going on here, so as a result, you will need to bundle or run with node `4.x` or higher. This lib *should* work, with no additional compilation in all current browsers (Edge, Safari, Chrome, Firefox), if it does not, please file an issue as I really, really want it to. :smile_cat:.
+There is no compilation going on here, so as a result, you will need to bundle or run with node `4.x` or higher.
 
-Another thing to note is, if you are transpiling, then destructuring in your `import` statement is not going to work as you are thinking (maybe if you are using `babel`, but this will be broken once modules are available in node, so be careful). Basically you should not do this, as `crocks` will not be set up for it until modules are available in node:
+**Precaution**: if you are de-structuring inline your `import` statement, you will need to use `babel` as this will not be supported once modules are available in node. Basically you should not do this, as `crocks` will not be set up for it until modules are available in node:
```
evilsoft: the spelling is wrong for destructuring [as seen here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)

evilsoft: The new wording of this paragraph derails the point of the original. Do no think this change is needed.
Curious the reasoning for this change.

README.md
```diff
@@ -1,5 +1,10 @@
-# crocks.js
-`crocks` is a collection of popular *Algebraic Data Types (ADTs)* that are all the rage in functional programming. You have heard of things like `Maybe` and `Either` and heck maybe even `IO`, that is what these are. The main goal of `crocks` is to curate and provide not only a common interface between each type (where possible of course), but also all of the helper functions needed to hit the ground running.
+<img style="display: block; width: 360px; margin: 24px auto;" src="https://cdn.rawgit.com/rstegg/crocks/dfdee636/crocks_js.svg" />
+
+# a collection of popular *Algebraic Data Types*
```
evilsoft: Maybe this could be a sub-heading, once the image is removed and should probably match the description of the repo for :corn:sistancy

README.md
rstegg: It's easier to read, and with the sub-heading it needed a fair amount of re-write

README.md
rstegg: they are more commonly referred in this aspect.  the quotes entail your point of view on the terms

README.md
rstegg: It's again for readability; it's not very common to detail how `destructuring` works in a library anyways, so it could probably be pulled out altogether

README.md
rstegg: you chose the name 😛 it also fits the cartoon-y style of [fantasyland](https://github.com/fantasyland/fantasy-land) and most functional programming libraries

README.md
evilsoft: For how long though?
When will they just be JS. Then we need to go back and change them.
I do not know people that say 3rd Edition for features adopted back in 1999?
Lets just stick with Javascript :smile:

rstegg: when that happens I'll send another PR to remove them 🆒?

README.md
evilsoft: It is more about destructring it is pointing to the design of this lib.
We are specifically not going to be able to work with some transpiler patterns.
People should know that.

README.md
evilsoft: What makes this easier to read?
How is this easier to read?

README.md
rstegg: markdown?

README.md
rstegg: alrighty, I don't think it's needed then.  how much do you want to keep, or close the PR?

rstegg:
I highly recommend the logo stay, the revision can go; and I can rename pr to "Adds first logo"

README.md
evilsoft: They sure can. but again, we should scrap it anyway.
At least for now, until a branding pass is done for the project.

evilsoft:
The logo is a good idea, do not get me wrong.
I just think this is not the time to add branding bits like this.

README.md
evilsoft: I would keep the grammatical and punctuation fixes and the stuff that was not commented on.

README.md
evilsoft: You can. Just like this:


not the svg itself, but the reference to it, like any other image asset.

rstegg:
Too soon for any flashy logos, revisions are pointless without the logo. Closing~

PR #38, opened 2017-01-26
Adds Identity API to docs

rstegg:
This is a PR for issue #42.

A _WIP_ **API** for crocks; you should review it and change what you'd like, and add to readme; it's modeled after the [most.js api](https://github.com/cujojs/most/blob/master/docs/api.md) 

Not much to dispute here; mostly copied from readme 🧀


evilsoft:
This is an amazing start!
Thanks so much. Would you mind adding this folder to the `.npmignore` so we do not push documentation to the package.

I am a little worried about this getting stale and an extra thing to remember to do. It could get out of date easily. So we may want to update the main readme to lose the hard coupling.

But that is something we can fix up.

evilsoft:
Don't worry about updating the README, we can adjust that after the docs get a bit more fleshed out.

docs/API.md
```diff
@@ -0,0 +1,353 @@
...
+| `State` | `get`, `gets`, `modify` `of`, `put`| `ap`, `chain`, `evalWith`, `execWith`, `map`, `of`, `runWith` |
+| `Unit` | `empty`, `of` | `ap`, `chain`, `concat`, `empty`, `equals`, `map`, `of`, `value` |
+| `Writer`| `of` | `ap`, `chain`, `equals`, `log`, `map`, `of`, `read`, `value` |
+
+[*] based on [this article](https://medium.com/@drboolean/monoidal-contravariant-functors-are-actually-useful-1032211045c4#.polugsx2a)
+
+
+## Monoids
+
+Each `Monoid` provides a means to represent a binary operation and is usually locked down to a specific type. These are great when you need to combine a list of values down to one value. In this library, any ADT that provides both an `empty` and `concat` function can be used as a `Monoid`. There are a few of the `Crocks` that are also monoidial, so be on the look out for those as well. All `Monoids` work with the point-free functions `mconcat`, `mreduce`, `mconcatMap` and `mreduceMap`.
```
evilsoft: Why get rid of the `crocks` description and leave this one?

docs/API.md
```diff
@@ -0,0 +1,353 @@
...
+* `isNumber`: Number that is not a NaN value, Infinity included
+* `isObject`: Plain Old Javascript Object (POJO)
+* `isSemigroup`: an ADT that provides a `concat` function
+* `isSetoid`: an ADT that provides an `equals` function
+* `isString`: String
+* `isTraversable`: an ADT that provides `map` and `traverse` functions
+
+## Point-free
+
+While it can seem natural to work with all these containers in a fluent fashion, it can get cumbersome and hard to get a lot of reuse out of. A way to really get the most out of reusability in Javascript is to take what is called a point-free approach. Below is a small code same to contrast the difference between the two calling styles:
```
evilsoft: Also may want to get rid of this one as well?

docs/API.md
evilsoft: Actually the more I think about it, we should probably keep all descriptions here.
So instead of removing the descriptions, we should probably add the crocks description back.

docs/API.md
evilsoft: Keep this and add back the crocks description.

rstegg: I was not sure whether to add them here or keep them in the README, so they were purposely inconsistent for option; does your new issue #42 makes these suggestions obsolete, or what do you think?

docs/crocks/Identity.md
```diff
@@ -0,0 +1,63 @@
...
+## Instances
+
+### ap
+
+`Function (a -> b) -> Function b`
+
+### chain
+
+`(a -> m b) -> m b`
+
```
rstegg: Different types of instance definitions:  explicit `Function` vs implicit `f`, and implicit `m` vs (not sure what m stands for)

docs/crocks/Identity.md
```diff
@@ -0,0 +1,63 @@
...
+## Constructors
+
+### of
+
+```js
+crocks.Identity.of([1,2,10])
+crocks.Identity([1,2,10])
+```
+
```
rstegg: Different type of definition for constructor;  here is only a definition by example

evilsoft: `of` is a very specific function that signals an `Applicative`.
In many cases the normal constructor function is the same thing as `of`.
We still include `of` in those cases so any functions that rely on calling `of` can call it.

So the constructor parameters should be represented by the type siggy in most cases.
For things like `Writer` it takes a `Monoid` and kicks you back a constructor that is fixed to that Monoid like: `Writer (Sum b) a`

Also for Maybe, it also takes an internal UnionType and if it is not the UnionType, it delegates to `Maybe.of` (which is basically `Maybe.Just`)

Good 👁 tho.

docs/crocks/Identity.md
evilsoft: Here `m` is used to represent any `Monad`:
`Monad m => (a -> m b) -> m b`.

So as this is specifically an Identity, it _could_ (and maybe _should_) be written as:
`(a -> Identity b) -> Identity b` 

Although it would probably look better as: 
`Identity m => (a -> m b) -> m b`

If that makes sense?

docs/API.md
evilsoft: I would add the `crocks` section. We can take it out if we need to once we cut over to these docs.

evilsoft:
@rstegg This is really shaping up nicely.
Thanks so much for jumping into this so enthusiastically. Excited to see what you do with all of this.

docs/crocks/Identity.md
evilsoft: Actually a better example to highlight the difference would be `IO`.
You can only construct an IO with a function. BUT `of` will take anything you pass it and not check it at all and give you back an IO that when run, returns whatever it was you passed `of`

docs/crocks/Identity.md
rstegg: But then should all defs be prefixed by the representing monad?

docs/crocks/Identity.md
```diff
@@ -0,0 +1,61 @@
...
+| Constructor | Instance |
+|:---|:---|
+| [`of`](#of) | [`ap`](#ap), [`chain`](#chain), [`equals`](#equals), [`map`](#map), [`of`](#of), [`sequence`](#sequence), [`traverse`](#traverse), [`value`](#value) |
+
+## Constructors
+
+### of
+
+`Identity a => a -> Identity a`
```
evilsoft: Should be something like:
`Identity m => a -> m a`

docs/crocks/Identity.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Instances
+
+### inspect
```
evilsoft: `inspect` should be along the lines of:
`() -> String`

docs/crocks/Identity.md
```diff
@@ -0,0 +1,61 @@
...
+
+`Identity a => Identity a`
+
+### value
+
+`() -> a`
```
evilsoft: `value`  should show be all like:

`Identity m => m a ~> () => a

docs/crocks/Identity.md
```diff
@@ -0,0 +1,61 @@
...
+
+### type
+
+`() -> Identity`
```
evilsoft: `type` returns a string identifier so something like:
`() -> String`

docs/crocks/Identity.md
```diff
@@ -0,0 +1,61 @@
...
+
+### equals
+
+`a -> Boolean`
+
+### map
+
+`Identity m => m (a -> b) -> m b`
```
evilsoft: So `map` takes a function and lifts it into the context of `Identity`, and because we need to show how the mapping works, we will want to use `~>` here.  So this is how it should be shown IMO:
`Identity m => m a ~> (a -> b) -> m b`

docs/crocks/Identity.md
```diff
@@ -0,0 +1,61 @@
...
+
+### ap
+
+`Identity m => m a ~> m (a -> b) -> m b`
```
evilsoft: This siggy matches the Fantasy Land spec and it just so happens this is the only place in `crocks` that does not match the spec (for reasons). So to get it to show how `crocks` handles `Apply`, it should be something like:
`Identity m => m (a -> b) ~> m a -> m b`.

Notice how it is reversed.

docs/crocks/Identity.md
```diff
@@ -0,0 +1,61 @@
...
+
+### of
+
+`Identity a => a -> Identity a`
```
evilsoft: Should stick with the "shorthand" for 🌽sistancy. Even though this is on the instance for :corn:venience, we do not really need to show the instance, as this function is not dependent on the instance (?)

`Identity m => a -> m a`

docs/crocks/Identity.md
```diff
@@ -0,0 +1,61 @@
...
+
+### chain
+
+`Identity m => m a ~> (a -> m b) -> m b`
```
evilsoft: 🐱fect!! This is exactly right!!

docs/crocks/Identity.md
```diff
@@ -0,0 +1,61 @@
...
+
+### sequence
+
+`Identity m, Applicative f => m (f b) ~> f (m a) -> m (f a)`
```
evilsoft: So for sequence we need an `f` returning function as the argument, but it is independent of the other types in this flow. so we can show that by using `b` like this:
`Identity m, Applicative f => m (f a) ~> (b -> f b) -> f (m a)`

docs/crocks/Identity.md
```diff
@@ -0,0 +1,61 @@
...
+
+### traverse
+
+`Identity m, Applicative f => m a ~> f (m b) -> m (f b)`
```
evilsoft: And this bad boy should be something along the lines of
`Identity m, Applicative f => m a ~> (c -> f c) -> (a -> f b) -> f (m b)`

(The siggy is wrong for the `crocks` pointfree function. I will update that with the next PR. It should take a `m a` and not a `m (f a)` like it says. 🙄 )

docs/crocks/Identity.md
rstegg: `Identity () ->` ?

docs/crocks/Identity.md
rstegg: I do not understand this one, folding `Identity m => m a -> a`?

docs/crocks/Identity.md
rstegg: Sweet!

docs/crocks/Identity.md
rstegg: Curious if `m a ~> (b -> f b) -> f (m a)` works for 🌽 sistency to traverse then?

docs/crocks/Identity.md
rstegg: why do i feel like half of this can be thrown away?

`Identity m => m (a -> b)` or `Identity m => m a -> m b`?

docs/crocks/Identity.md
evilsoft: No need to call out Identity here. this function just maps unit to a string.

docs/crocks/Identity.md
evilsoft: `value` just maps unit to whatever is inside the identity.

```javascript
// `a` would be Number in this case
Identity(3)
  .value()  // returns 3

```

docs/crocks/Identity.md
evilsoft: I don't know what you feel that.

There is a constraint that the instance that this is being called on, MUST contain a function.
So that is why `m (a -> b)` is needed to the left of `~>`...We need to communicate that constraint.

Make sense? Do you have any ideas on how we could better communicate that constraint?
👂

docs/crocks/Identity.md
evilsoft: nope. we need to already contain the `f a` for `sequence`. 

In fact, that is the thing about `sequence`, it must already contain the `Applicative`.
Where `traverse` has a value and takes a function that will lift that value (and apply any additional mapping to it, if it wants) into the `Applicative`, and then `sequence` over that.

Does that make sense?

evilsoft: Good 👁 though! in catching the similarity

docs/crocks/Identity.md
rstegg: throw away was probably not the right way to express my confusion. 🎧

rstegg: well i feel stupid for trying now

rstegg:
On `Either`, only worked on the sigs, nothing above it. next commit after review will have those changes

evilsoft:
Merging...Here is a squirrel

PR #45, opened 2017-01-30
Fix regression in`mreduce` and provide better coverage

evilsoft:
## Thank goodness for Coveralls

This PR fixes a slight regression in how errors would have ben reported in a couple cases for `mreduce`. This was caught by the coverage report. 🎉 

Also updated many specs that were reported as lacking full coverage.


evilsoft:
After this is merged, will cut a release `0.2.2`

Issue #47, opened 2017-02-01
Add some transformation functions


evilsoft:
Was debating for a while on how to do transformations in this lib.

After a bunch of thinking, I think we should go the helper function route with the possible permutations in a `transforms` folder. Each function should follow the form: [x]To[Y] (maybeToEither). We can start off with just the Coproducts for now:

* Either:
  - [ ] `eitherToMaybe : Either b a -> Maybe a`
  - [ ] `eitherToAsync : Either e a -> Async e a`

* Maybe:
  - [ ] `maybeToEither : b -> Maybe a -> Either b a`
  - [ ] `maybeToAsync : e -> Maybe a -> Async e a`

evilsoft:
Submitted [PR#48](https://github.com/evilsoft/crocks/issues/48) to address this issue.

evilsoft:
Merged [PR#48](https://github.com/evilsoft/crocks/issues/48) this is :red_circle: :eye: to :ship:

PR #48, opened 2017-02-04
Add Inital Transformation Functions

evilsoft:
## Robits  in Disguise

This PR addresses [issue#47](https://github.com/evilsoft/crocks/issues/47) and adds all the initial natural transofrmation functions. Really only covers the two main Sum Types `Maybe` and `Either` back and forth and to `Async`. Once `Validation` is added, will add an additional set of functions to account for those transformations.


evilsoft:
Retroactive Squirrel

PR #52, opened 2017-02-10
Adds Either API to docs

rstegg:
Does not conform to other spec library variables [Either l r].  I feel it is easier to follow, but I _might_ be wrong [because !:corn:sistency]

### DIFF
--- a/docs/crocks/Either.md
+++ b/docs/crocks/Either.md
@@ -1,0 +1,103 @@
+# Either
+
+`Either l r`
+
+Crock which takes a `l` and a `r`, or just a `r`, and acts accordingly:
+* taking only `r` will proceed as `Identity`
+* taking `l` and `r` will handle errors (or results defined as `Left`) with `l` and results with `r` (or results defined as `Right`)
+
+The "`Either`" crock creates disjunction with a linear data flow.
+
+```js
+// Imperative flow
+const renderPage = user => {
+  const isAuthenticated = user ? true : false
+  return isAuthenticated ? "Home Page" : "Login Page"
+}
+
+// Declarative flow
+const isAuthenticated = u => u ? Either.Right(u) : Either.Left("Login Page")
+
+const renderPage = user =>
+  isAuthenticated(user)
+  .map(u => "Home Page")
+  .value()
+```
+
+`Either` exposes these constructors and instances:
+
+| Constructor | Instance |
+|:---|:---|
+| [`Left`](#left), [`Right`](#right), [`of`](#of) | [`inspect`](#inspect), [`value`](#value), [`type`](#type), [`equals`](#equals), [`map`](#map), [`ap`](#ap), [`of`](#of), [`chain`](#chain), [`sequence`](#sequence), [`traverse`](#traverse) |
+
+## Constructors
+
+### Left
+
+`Either m => l -> m l _`
+
+### Right
+
+`Either m => r -> m _ r`
+
+### of
+
+`Either m => a -> m _ a`
+
+## Instances
+
+### inspect
+
+`() => String`
+
+### either
+
+`Either m => m l r ~> (l -> c) -> (r -> d) -> m l d`
+
+### value
+
+`Either m => m l r ~> () => a`
+
+### type
+
+`() -> String`
+
+### swap
+
+`Either m => m l r ~> m r l`
+
+### coalesce
+
+`Either m => m l r ~> (l -> a) -> m _ a`
+
+### equals
+
+`a -> Boolean`
+
+### map
+
+`Either m => m l r ~> (r -> d) -> m c d`
+
+### bimap
+
+`Either m => m l r ~> (l -> c) -> (r -> d) -> m c d`
+
+### ap
+
+`Either m => m l (r -> a) ~> m l r -> m l a`
+
+### of
+
+`Either m => a -> m _ a`
+
+### chain
+
+`Either m => m l r ~> (r -> m a) -> m l a`
+
+### sequence
+
+`Either m, Applicative f => m l (f r) ~> (l -> f l) -> (r -> f r) -> f (m l r)`
+
+### traverse
+
+`Either m, Applicative f => m l r ~> (a -> f a) -> (r -> f b) -> f (m l b)`
--- a/docs/crocks/Either.md
+++ b/docs/crocks/Either.md
@@ -1,27 +1,13 @@
 # Either
 
-`Either l r`
+`Either a b`
 
-Crock which takes a `l` and a `r`, or just a `r`, and acts accordingly:
-* taking only `r` will proceed as `Identity`
-* taking `l` and `r` will handle errors (or results defined as `Left`) with `l` and results with `r` (or results defined as `Right`)
+Crock which --
 
 The "`Either`" crock creates disjunction with a linear data flow.
 
 ```js
-// Imperative flow
-const renderPage = user => {
-  const isAuthenticated = user ? true : false
-  return isAuthenticated ? "Home Page" : "Login Page"
-}
-
-// Declarative flow
-const isAuthenticated = u => u ? Either.Right(u) : Either.Left("Login Page")
-
-const renderPage = user =>
-  isAuthenticated(user)
-  .map(u => "Home Page")
-  .value()
+--
 ```
 
 `Either` exposes these constructors and instances:
@@ -34,11 +20,11 @@ const renderPage = user =>
 
 ### Left
 
-`Either m => l -> m l _`
+`Either m => a -> m a _`
 
 ### Right
 
-`Either m => r -> m _ r`
+`Either m => a -> m _ a`
 
 ### of
 
@@ -52,11 +38,11 @@ const renderPage = user =>
 
 ### either
 
-`Either m => m l r ~> (l -> c) -> (r -> d) -> m l d`
+`Either m => m c a ~> (c -> b) -> (a -> b) -> b`
 
 ### value
 
-`Either m => m l r ~> () => a`
+`Either m => m c a ~> () => c | a`
 
 ### type
 
@@ -64,11 +50,11 @@ const renderPage = user =>
 
 ### swap
 
-`Either m => m l r ~> m r l`
+`Either m => m a b ~> m a b`
 
 ### coalesce
 
-`Either m => m l r ~> (l -> a) -> m _ a`
+`Either m => m a b ~> (b -> c) -> m _ c`
 
 ### equals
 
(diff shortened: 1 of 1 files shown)


rstegg:
Do you need `.npmignore` commit on each PR?

evilsoft:
No need for dat `.npmignore`. That first PR is merged into `docs`.

rstegg:
yay! ill wait for review on this pr, i'm prepared for the onslaught :call_me_hand: (how many people use that emoji?)

docs/crocks/Either.md
```diff
@@ -0,0 +1,103 @@
...
+  isAuthenticated(user)
+  .map(u => "Home Page")
+  .value()
+```
+
+`Either` exposes these constructors and instances:
+
+| Constructor | Instance |
+|:---|:---|
+| [`Left`](#left), [`Right`](#right), [`of`](#of) | [`inspect`](#inspect), [`value`](#value), [`type`](#type), [`equals`](#equals), [`map`](#map), [`ap`](#ap), [`of`](#of), [`chain`](#chain), [`sequence`](#sequence), [`traverse`](#traverse) |
```
evilsoft: There is a problem with the dual `of`s. The instance version points to the constructor version.
Leave it for now, will figure that out later.

docs/crocks/Either.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Constructors
+
+### Left
+
+`Either m => l -> m l _`
```
evilsoft: change `l` to `a`

docs/crocks/Either.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Right
+
+`Either m => r -> m _ r`
```
evilsoft: change `r` to `a`

docs/crocks/Either.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Instances
+
+### inspect
+
+`() => String`
+
+### either
+
+`Either m => m l r ~> (l -> c) -> (r -> d) -> m l d`
```
evilsoft: `either` is the `cata` here, it is how you fold values out of the `Either`.

this should be:
`Either m => m c a ~> (c -> d) -> (a -> b) -> d | b`
Notice how it does not return another `Either`, but one of the results from the functions run on the internal value. Technically if this was not JS, both functions would map to the same type, so the | would not be there. But this is JS.

evilsoft: oh and get rid of the `l` & `r` please

docs/crocks/Either.md
```diff
@@ -0,0 +1,103 @@
...
+
+### value
+
+`Either m => m l r ~> () => a`
```
evilsoft: `a` is not defined in this def.
should be something like (after the `l` & `r` are moved to `c` & `a`):

`Either m => m c a ~> () -> c | a`

docs/crocks/Either.md
```diff
@@ -0,0 +1,103 @@
...
+
+### type
+
+`() -> String`
+
+### swap
+
+`Either m => m l r ~> m r l`
```
evilsoft: swap out the `l` and `r` please

docs/crocks/Either.md
```diff
@@ -0,0 +1,103 @@
...
+
+### coalesce
+
+`Either m => m l r ~> (l -> a) -> m _ a`
```
evilsoft: swap out the `l` and `r` please

docs/crocks/Either.md
```diff
@@ -0,0 +1,103 @@
...
+
+### equals
+
+`a -> Boolean`
+
+### map
+
+`Either m => m l r ~> (r -> d) -> m c d`
```
evilsoft: do not know where that `c` came from. once you move away from the `l` and `r` it should look something like this:
`Either m => m c a ~> (a -> b) -> m c b`

docs/crocks/Either.md
```diff
@@ -0,0 +1,103 @@
...
+
+### bimap
+
+`Either m => m l r ~> (l -> c) -> (r -> d) -> m c d`
```
evilsoft: swap out the `l` and `r` please.

docs/crocks/Either.md
```diff
@@ -0,0 +1,103 @@
...
+
+### ap
+
+`Either m => m l (r -> a) ~> m l r -> m l a`
```
evilsoft: swap out the `l` and `r` please

docs/crocks/Either.md
```diff
@@ -0,0 +1,103 @@
...
+
+### of
+
+`Either m => a -> m _ a`
+
+### chain
+
+`Either m => m l r ~> (r -> m a) -> m l a`
```
evilsoft: swap out the `l` and `r` please

and the left is not represented in you function so after you swap out `l` & `r` it should look something like:
`Either m => m c a ~> (a -> m c b) -> m c b`

docs/crocks/Either.md
```diff
@@ -0,0 +1,103 @@
...
+
+### sequence
+
+`Either m, Applicative f => m l (f r) ~> (l -> f l) -> (r -> f r) -> f (m l r)`
```
evilsoft: So `sequence` and `traverse` can be tricky. After you swap out the `l` & `r`, `sequence` should look like:

`Either m, Applicative f => m c (f a) ~> (b -> f b) -> f (m c a)`
It does not take that second function. only works over the rightmost parameter. and that first function is for safety.

docs/crocks/Either.md
```diff
@@ -0,0 +1,103 @@
...
+
+### traverse
+
+`Either m, Applicative f => m l r ~> (a -> f a) -> (r -> f b) -> f (m l b)`
```
evilsoft: You got the feel of `traverse`, when you swap out the `l` & `r` and you are there. should look a little like this:
`Either m, Applicative f => m c a ~> (d -> f d) -> (a -> f b) -> f (m c b)`

docs/crocks/Either.md
rstegg: This should be true for the identity sig as well. keeping it cheap, `#user-content-of-1` could work.

docs/crocks/Either.md
rstegg: I thought `a` through `z` were specification related and not signature related.  it is caustic

docs/crocks/Either.md
evilsoft: again do not worry about it. Continue with this, and we will correct them all in one shot

docs/crocks/Either.md
evilsoft: please change `l` to `a` here, `l` does not make sense.

evilsoft: they are, but you are trying to give them meaning here outside of the containing types.

docs/crocks/Either.md
evilsoft: So I changed my mind on this fold. So while we cannot prevent the `d | b` from happening or check for it in JS, we should probably not teach bad habits to people. Both function, ideally should return the same type. So let's go with something like:

`Either m => m c a ~> (c -> b) -> (a -> b) -> b`

Which would be better for everybody to make sure both return `b`.
Make sense? If not please ask.

docs/crocks/Either.md
evilsoft: Wow, it is good to see these defs. It is amazing what you discover 

Now I realize that this is not valid and that `Either` and `Maybe` cannot/should not implement `value`. I am going to add an Issue for the `0.3.0` release to remove `value` from `Either` and `Maybe`. But keep it in the docs for now.

docs/crocks/Either.md
```diff
@@ -0,0 +1,89 @@
...
+
+Crock which --
+
+The "`Either`" crock creates disjunction with a linear data flow.
+
+```js
+--
+```
+
+`Either` exposes these constructors and instances:
```
evilsoft: Sorry one last thing I missed on the last review, would it be possible to change this to something like:

`Either` exposes the following constructor and instance functions:

docs/crocks/Either.md
```diff
@@ -0,0 +1,89 @@
...
+
+`Either m => m c a ~> () => c | a`
+
+### type
+
+`() -> String`
+
+### swap
+
+`Either m => m a b ~> m a b`
```
evilsoft: This should swap the types, as it negates the or.

so something like:
`Either m => m a b ~> m b a`

docs/crocks/Either.md
```diff
@@ -0,0 +1,89 @@
...
+
+### coalesce
+
+`Either m => m a b ~> (b -> c) -> m _ c`
```
evilsoft: Would be nice to keep the `m c a` for 🌽sistancy.

docs/crocks/Either.md
evilsoft: Would be nice to keep the `m c a` for 🌽sistancy.

docs/crocks/Either.md
```diff
@@ -0,0 +1,89 @@
...
+
+| Constructor | Instance |
+|:---|:---|
+| [`Left`](#left), [`Right`](#right), [`of`](#of) | [`inspect`](#inspect), [`value`](#value), [`type`](#type), [`equals`](#equals), [`map`](#map), [`ap`](#ap), [`of`](#of), [`chain`](#chain), [`sequence`](#sequence), [`traverse`](#traverse) |
+
+## Constructors
+
+### Left
+
+`Either m => a -> m a _`
```
evilsoft: Would be nice to keep the `m c a` for 🌽sistancy.

So in this case, it may make sense to so something like
`Either m => c -> m c _`

docs/crocks/Either.md
```diff
@@ -0,0 +1,89 @@
+# Either
+
+`Either a b`
```
evilsoft: Would be nice to go with the `Either c a` for 🌽sistancy now that I see it in play.

docs/crocks/Either.md
```diff
@@ -0,0 +1,89 @@
...
+
+Crock which --
+
+The "`Either`" crock creates disjunction with a linear data flow.
```
evilsoft: Please remove all description text.

docs/crocks/Either.md
```diff
@@ -0,0 +1,89 @@
...
+
+`a -> Boolean`
+
+### map
+
+`Either m => m c a ~> (a -> b) -> m a b`
+
+### bimap
+
+`Either m => m a b ~> (a -> c) -> (b -> d) -> m c d`
```
evilsoft: Would be nice to keep the `m c a` for 🌽sistancy. So that will look something like:
`Either m => m c a ~> (c -> d) -> (a -> b) -> m d b`

Issue #34, opened 2017-01-22
Create a logic folder


evilsoft:
Helpers is becoming huge, think it is time to take all Logic based functions and move them into a `logic` folder. The functions that need to be moved are:
* `ifElse`
* `not`
* `unless`
* `when`
* ~`has` (to be added from [issue#33](https://github.com/evilsoft/crocks/issues/33)~

evilsoft:
Submitted [PR#53](https://github.com/evilsoft/crocks/issues/53) to address this PR.
Keeping `hasKey` right where it lives for now. Does not make sense moving it here.

evilsoft:
Merged in [PR#53](https://github.com/evilsoft/crocks/issues/53).
Closing.

PR #53, opened 2017-02-10
Move logic functions into new `logic` folder and add `and` and `or` functions

evilsoft:
## It's only Logical...la la logical

This PR addresses [issue#34](https://github.com/evilsoft/crocks/issues/34) and moves over the following functions into a new folder called `logic`:

* `ifElse`
* `not`
* `unless`
* `when`

In addition to moving over those existing functions, it also adds `and` and `or` functions for combining predicates (or `Pred`s) over conjunction or disjunction respectively (mad respekt).


README.md
```diff
@@ -188,6 +184,21 @@ It is hard knowing what is going on inside of some of these ADTs or your wonderf
 #### `tryCatch : (a -> b) -> a -> Either e b`
 Typical try-catch blocks are very imperative in their usage. This `tryCatch` function provides a means of capturing that imperative nature in a simple declarative style. Pass it a function that could fail and it will return you another function wrapping the first function. When called the new function will either return the result in an `Either.Right` if everything was good, or an error wrapped in an `Either.Left` if it fails.
 
+### Logic Functions
+The functions in this section are used to represent simple to complex logical branching in a declarative manner. All of these function will take predicate functions or `Pred`s and either return a resulting `Boolean` or run some function based on the result of the predicates. All of these functions work with functions/`Pred`s instead of values to allow for building out complex, "lazy" evaluated and allow easy composition.
```
rstegg: All of these function[s] will take predicate functions (crocks says`Pred`s), [and either return a `Boolean` or run some function based on the result of the predicates === not informative, readers' deduction should work here].

rstegg: Because they intend to work with `Pred`s instead of values, this allows for building out [complex,] "lazy" evaluation and easy composition.

README.md
```diff
@@ -188,6 +184,21 @@ It is hard knowing what is going on inside of some of these ADTs or your wonderf
...
+
+#### `and : ((a -> Boolean) | Pred) -> ((a -> Boolean) | Pred) -> a -> Boolean`
+Say you have two predicate functions or `Pred`s and would like to combine them into one predicate over conjunction, well you came to the right place, `and` accepts either predicate functions or `Pred`s and will return you a function ready to take a value. Once that value is passed, it will run it through both of the predicates and return the result of combining it over a `logical and`. This is super helpful combined with `or` for putting together reusable, complex predicates. As they follow the general form of `(a -> Boolean)` they are easily combined with other logic functions.
```
rstegg: The Majority of your descriptions read as if you REALLY want to get your word across, but you don't contextualize 🥂  [in essence, instead of trying to sound human with "Say you have two predicate functions", you can just sound like a robot here with "Combinatorial mathematics"

logic/and.js
```diff
@@ -0,0 +1,26 @@
...
+
+const ifElse = require('./ifElse')
+
+const predOrFunc =
+  ifElse(isFunction, identity, x => x.runWith)
+
+// and : (a -> Boolean) | Pred -> (a -> Boolean) | Pred -> a -> Boolean
+function and(f, g) {
+  if(!((isFunction(f) || isSameType(Pred, f)) && (isFunction(g) || isSameType(Pred, g)))) {
+    throw new TypeError('and: Preds or predicate functions required for first two arguments')
```
rstegg: decide on if you want to use `Pred` or `predicate` for pete's sake

README.md
```diff
@@ -57,6 +57,8 @@ There are (6) classifications of "things" included in this library:
 
 * Helper Functions (`crocks/helpers`): All other support functions that are either convenient versions of combinators or not even combinators at all cover this group.
 
+* Logic Functions (`crocks/logic`): A helpful collection of Logic based combinators. All of these functions work with predicate functions and let you combine them in some very interesting ways.
```
evilsoft: A `Pred` is an actual ADT included in the lib. A predicate function is anything that takes a value and returns a `Boolean`. So I do not want to give the impression that they only work with the predicates `crocks` ships with. They will work for all predicates, even user defined ones. 
What should I link to to make it easier to understand?

README.md
evilsoft: Well they work with both predicate functions, functions in the form of `(a -> Boolean)` and also with the ADT `Pred`. Those are two distinct things.

So I like that rewording for that first sentence. Will update.
So are you saying that the typical developer would not know what a predicate is?

README.md
evilsoft: I understand that, but I think I would like to convey personality and conversation here.
Mostly as this is the way the entire doc reads, so if I get all robot here, it will stick out.
If you think this is an issue and you want it to read like a robot, put in an issue, and it can get tackled all at once. Instead of one sore thumb.

logic/and.js
evilsoft: I would like to use both.
a `Pred` is an actual `ADT`, a predicate is a function from `(a -> Boolean)` are you suggesting that there be two function, one for working with only predicate functions, and another with `Pred`?

In case you are wondering, here is the linky to [`Pred`](https://github.com/evilsoft/crocks/blob/master/crocks/Pred.js)

README.md
evilsoft: Ohhh. I gotcha. 
Good call, there are two types of functions here, ones that extend or return new predicates: `and`, `not`and  `or`;
and there are others that will run a function based on the result of a predicate: `ifElse`, `unless` and `when`. 

Maybe the message is not clear there. Can you think of a way that I can reword that to get the point accross.

logic/and.js
evilsoft: Although I may not understand what you mean here, could you explain?

evilsoft: I think I am goin to leave this as both types are accepted.
And this lets you do things like:
[this](https://gist.github.com/evilsoft/0f720f43af2000d2c480a712f7db4ebd)

rstegg:
👍  good to go for merging, and all tests pass.  minor tweaks to readme is recommended, but can be put off. What do you think @evilsoft ?

evilsoft:
Will make those README suggestion and push up tonight!

Do you have any suggestions for making this a little clearer:

 > Ohhh. I gotcha.
Good call, there are two types of functions here, ones that extend or return new predicates: and, not and or;
and there are others that will run a function based on the result of a predicate: ifElse, unless and when.
Maybe the message is not clear there. Can you think of a way that I can reword that to get the point across.

Would like to communicate that it takes either `Pred a` or a `(a -> Boolean)` and by your comment, I do not think that is coming through.

rstegg:
> The functions in this section are used to represent simple to complex logical branching in a declarative manner. All of these function will take predicate functions or `Pred`s and either return a resulting `Boolean` or run some function based on the result of the predicates. All of these functions work with functions/`Pred`s instead of values to allow for building out complex, "lazy" evaluated and allow easy composition.

My suggestion:

> The functions in this section are used to represent logical branching in a declarative manner. Each of these functions require either a `Pred` or a _predicate function_, as dependencies.  Be aware that your `Pred` or _predicate function_ must return an evaluation, most commonly a boolean.  Since these functions work with `Pred`s rather than values, this allows for "lazy" evaluation and easy computation.

evilsoft:
Made the requested change. Did a little editing on your provided copy @rstegg.
But this is getting merged!!

PR #57, opened 2017-02-12
Adds Maybe API

rstegg:
tumblr_mh7h62adzq1qblevio1_500

### DIFF
--- a/docs/crocks/Maybe.md
+++ b/docs/crocks/Maybe.md
@@ -1,0 +1,81 @@
+# Maybe
+
+`Maybe a`
+
+--
+
+--
+
+```js
+--
+```
+
+`Maybe` exposes the following constructor and instances:
+
+| Constructor | Instance |
+|:---|:---|
+| [`Nothing`](#nothing), [`Just`](#just), [`of`](#of) | [`inspect`](#inspect), [`maybe`](#maybe), [`either`](#either), [`option`](#option), [`type`](#type), [`equals`](#equals), [`coalesce`](#coalesce), [`map`](#map), [`ap`](#ap), [`of`](#of), [`chain`](#chain), [`sequence`](#sequence), [`traverse`](#traverse) |
+
+## Constructors
+
+### Nothing
+
+`Maybe m => _ -> m a`
+
+### Just
+
+`Maybe m => a -> m a`
+
+### of
+
+`Maybe m => a -> m a`
+
+## Instances
+
+### inspect
+
+`() => String`
+
+### equals
+
+`a -> Boolean`
+
+### either
+
+`Maybe m => m a ~> (_ -> b) -> (a -> b) -> b`
+
+### option
+
+`Maybe m => m a ~> a -> a`
+
+### type
+
+`() -> String`
+
+### coalesce
+
+`Maybe m => m a ~> (_ -> b) -> (a -> b) -> m b`
+
+### map
+
+`Maybe m => m a ~> (a -> b) -> m b`
+
+### ap
+
+`Maybe m => m (a -> b) ~> m a -> m b`
+
+### of
+
+`Maybe m => a -> m a`
+
+### chain
+
+`Maybe m => m a ~> (a -> m b) -> m b`
+
+### sequence
+
+`Maybe m, Applicative f => m (f a) ~> (b -> f b) -> f (m a)`
+
+### traverse
+
+`Maybe m, Applicative f => m a ~> (c -> f c) -> (a -> f b) -> f (m b)`


docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
+# Maybe
+
+`Maybe c a`
```
evilsoft: While technically the siggy for `Maybe` is: `Maybe () a`,  but because the first parameter is fixed to `Unit`, we just never show it. So typically the siggy for `Maybe` is `Maybe a`.

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+| Constructor | Instance |
+|:---|:---|
+| [`Nothing`](#nothing), [`Just`](#just), [`of`](#of) | [`inspect`](#inspect), [`maybe`](#maybe), [`either`](#either), [`option`](#option), [`type`](#type), [`equals`](#equals), [`coalesce`](#coalesce), [`map`](#map), [`ap`](#ap), [`of`](#of), [`chain`](#chain), [`sequence`](#sequence), [`traverse`](#traverse) |
+
+## Constructors
+
+### Nothing
+
+`Maybe m => c -> m c _`
```
evilsoft: The first parameter is fixed to `()`, so this should be:

`Maybe m => _ -> m a`

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Just
+
+`Maybe m => a -> m _ a`
```
evilsoft: The first parameter is fixed to `()`, so this should be:

`Maybe m => a -> m a`

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+### of
+
+`Maybe m => a -> m _ a`
```
evilsoft: The first parameter is fixed to `()`, so this should be:

`Maybe m => a -> m a`

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Instances
+
+### inspect
+
+`() => String`
+
+### maybe
+
+`Maybe m => m c a ~> a -> m a -> a`
```
evilsoft: This is being removed. So we can remove it from these docs.

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+### either
+
+`Maybe m => m c a ~> (c -> b) -> (a -> b) -> b`
```
evilsoft: The first parameter is fixed to (), so this should be:

`Maybe m => m a ~> (() -> b) -> (a -> b) -> b`

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+### option
+
+`Maybe m => m c a ~> a -> m a -> a`
```
evilsoft: The first parameter is fixed to () and it does NOT take another Maybe as a parameter, so this should be:

`Maybe m => m a ~> a -> a`

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+### value
+
+`Maybe m => m c a ~> () => c | a`
```
evilsoft: `value` is not provided on `Maybe`.

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+### type
+
+`() -> String`
+
+### swap
+
+`Maybe m => m c a ~> m a c`
```
evilsoft: swap is not provided on `Maybe`

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+### coalesce
+
+`Maybe m => m c a ~> (c -> b) -> (a -> b) -> m _ b`
```
evilsoft: The first parameter is fixed to (), so this should be:

`Maybe m => m a ~> (() -> b) -> (a -> b) -> m b`

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+### equals
+
+`a -> Boolean`
```
evilsoft: `equals` should be moved above `either`

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+### map
+
+`Maybe m => m c a ~> (a -> b) -> m a b`
```
evilsoft: The first parameter is fixed to () and the result DOES not move from a `()` to `a`, so this should be:

`Maybe m => m a ~> (a -> b) -> m b`

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+### bimap
+
+`Maybe m => m c a ~> (c -> d) -> (a -> b) -> m d b`
```
evilsoft: `Maybe` is not a bifunctor because the first parameter is fixed to `unit`.
`bimap` is not provided by `Maybe`

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+### ap
+
+`Maybe m => m c (a -> b) ~> m c a -> m c b`
```
evilsoft: The first parameter is fixed to (), so this should be:

`Maybe m => m (a -> b) ~> m a -> m b`

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+### of
+
+`Maybe m => a -> m _ a`
```
evilsoft: The first parameter is fixed to (), so this should be:

`Maybe m => a -> m a`

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+### chain
+
+`Maybe m => m c a ~> (a -> m c b) -> m c b`
```
evilsoft: The first parameter is fixed to (), so this should be:

`Maybe m => m a ~> (a -> m b) -> m b`

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+### sequence
+
+`Maybe m, Applicative f => m c (f a) ~> (b -> f b) -> f (m c a)`
```
evilsoft: The first parameter is fixed to (), so this should be:

`Maybe m, Applicative f => m (f a) ~> (b -> f b) -> f (m a)`

docs/crocks/Maybe.md
```diff
@@ -0,0 +1,97 @@
...
+
+### traverse
+
+`Maybe m, Applicative f => m c a ~> (d -> f d) -> (a -> f b) -> f (m c b)`
```
evilsoft: The first parameter is fixed to (), so this should be:

`Maybe m, Applicative f => m a ~> (c -> f c) -> (a -> f b) -> f (m b)`

docs/crocks/Maybe.md
rstegg: Ah i've made a complete boo-boo then.

docs/crocks/Maybe.md
rstegg: what was the intent?

docs/crocks/Maybe.md
rstegg: option is returning `either(constant(n), identity)`, this isn't expressed in the sig?

docs/crocks/Maybe.md
evilsoft: Well it _was_ a way to extract the value for a [Comonad](https://github.com/fantasyland/fantasy-land#comonad).

But `Maybe` cannot be a `Comonad` because it could be a `Nothing` and it would only be valid for the case:
`Maybe Unit` .

 All other cases would not line up properly, so no `Comonad` for `Maybe`.
(same reason for `Either` and `value` on your last PR)

docs/crocks/Maybe.md
evilsoft: That `a` is what gets passed into `either`...there is no `m a`.

evilsoft: Ah. So THAT `either` is the either defined in `Maybe` and not the pointfree version. If that makes sense.

docs/crocks/Maybe.md
rstegg: Just and of! aha!

Issue #61, opened 2017-02-13
List constructor should behave like `List.of`


evilsoft:
Right now we require an Array to be passed into `List` and this becomes silly when trying to use `List` as a `Monoid` in this framework. We will need to provide a `fromArray` on the constructor, as it is super handy. And while `value` is needed to work with our `Monoid` system we got going on, it makes sense to add a `toArray` to the instance and also add `listToArray` and `arrayToList` transformation functions:

- [x] Remove Array constraint on `List` constructor, and have it work like `of` **breaking**
- [x] Add `fromArray` to `List` constructor
- [x] Add `toArray` to a `List` instance
- [x] Add `listToArray` transformation function.
- [x] Add `arrayToList` transformation function.

evilsoft:
Merged [PR#62](https://github.com/evilsoft/crocks/issues/62).
That addresses the first two non breaking changes.

evilsoft:
Merged [PR#64](https://github.com/evilsoft/crocks/issues/64).
completed the last two items and finishes up the non-breaking changes.

evilsoft:
Submitted [PR#75](https://github.com/evilsoft/crocks/issues/75) to address the final item on this issue.

evilsoft:
[PR#75](https://github.com/evilsoft/crocks/issues/75) merged.
Closing,

PR #62, opened 2017-02-14
List: Adds constructor _fromArray_ and instance _toArray_

rstegg:
squirrel-624939
#61 

Right now we require an Array to be passed into List and this becomes silly when trying to use List as a Monoid in this fra... 😴 

- [ ] Remove Array constraint on List constructor, and have it work like of breaking
- [x] Add fromArray to List constructor
- [x] Add toArray to a List instance
- [ ] Add listToArray transformation function.
- [ ] Add arrayToList transformation function.

### DIFF
--- a/crocks/List.js
+++ b/crocks/List.js
@@ -28,10 +28,10 @@ const _empty =
   () => List([])
 
 function fromArray(xs) {
-  if(!arguments.length || !isArray(xs)) {
+  if(!isArray(xs)) {
     throw new TypeError('List.fromArray: Array required')
   }
-  return xs.reduce((res, x)  => res.concat(List.of(x)), List.empty())
+  return xs.reduce((res, x) => res.concat(List.of(x)), List.empty())
 }
 
 function runSequence(acc, x) {
@@ -59,7 +59,7 @@ function runTraverse(f) {
 }
 
 function List(xs) {
-  if(!arguments.length || !isArray(xs)) {
+  if(!isArray(xs)) {
     throw new TypeError('List: Must wrap an array')
   }
 
--- a/crocks/List.js
+++ b/crocks/List.js
@@ -12,6 +12,8 @@ const constant = require('../combinators/constant')
 
 const _concat = require('../pointfree/concat')
 
+const _mconcat = require('../helpers/mconcat')
+
 const Maybe = require('./Maybe')
 const Pred = require('./Pred')
 
@@ -29,9 +31,9 @@ const _empty =
 
 function fromArray(xs) {
   if(!arguments.length || !isArray(xs)) {
-    throw new TypeError('List.fromArray: Must wrap an array')
+    throw new TypeError('List.fromArray: Array required')
   }
-  return List(xs)
+  return _mconcat(List, [xs])
 }
 
 function runSequence(acc, x) {
@@ -85,7 +87,7 @@ function List(xs) {
     () => xs.slice()
 
   const toArray =
-    () => xs.length && xs.reduce((res, x)  => res.concat(x), [])
+    () => xs.slice()
 
   const empty =
     _empty
--- a/crocks/List.js
+++ b/crocks/List.js
@@ -12,8 +12,6 @@ const constant = require('../combinators/constant')
 
 const _concat = require('../pointfree/concat')
 
-const _mconcat = require('../helpers/mconcat')
-
 const Maybe = require('./Maybe')
 const Pred = require('./Pred')
 
@@ -33,7 +31,7 @@ function fromArray(xs) {
   if(!arguments.length || !isArray(xs)) {
     throw new TypeError('List.fromArray: Array required')
   }
-  return _mconcat(List, [xs])
+  return xs.reduce((res, x)  => res.concat(List.of(x)), List.empty())
 }
 
 function runSequence(acc, x) {
--- a/crocks/List.js
+++ b/crocks/List.js
@@ -85,7 +85,7 @@ function List(xs) {
     () => xs.slice()
 
   const toArray =
-    () => xs.slice()
+    value
 
   const empty =
     _empty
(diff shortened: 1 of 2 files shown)


crocks/List.js
```diff
@@ -77,6 +84,9 @@ function List(xs) {
   const value =
     () => xs.slice()
 
+  const toArray =
+    () => xs.length && xs.reduce((res, x)  => res.concat(x), [])
```
evilsoft: This could just to xs.slice(). As `List` is just a wrapper for an `Array`.

crocks/List.js
```diff
@@ -27,6 +27,13 @@ const _of =
 const _empty =
   () => List([])
 
+function fromArray(xs) {
+  if(!arguments.length || !isArray(xs)) {
+    throw new TypeError('List.fromArray: Must wrap an array')
+  }
+  return List(xs)
```
evilsoft: We will not be able to use this on the same issue this work was defined.

So instead we will have to use a reduction here to get back a`List`.

As `List` is a `Monoid` we could probably use `mconcat` here like this:

```js
function fromArray(xs) {
  if(!arguments.length || !isArray(xs)) {
    throw new TypeError('List.fromArray: Must wrap an array')
  }
  return mconcat(List, xs)
```

crocks/List.spec.js
```diff
@@ -119,6 +120,14 @@ test('List value', t => {
   t.end()
 })
 
+test('List toArray', t => {
+  const a = List([ 'some-thing', ['else', 43], 34 ]).toArray()
+
+  t.same(a, [ 'some-thing', 'else', 43, 34 ], 'folds the wrapped array')
+
+  t.end()
+})
```
rstegg: I did not know whether to flatten the array so I figured not, but perhaps a `toFlatArray` (or something better-efficaciously named) could do that.

And for the `toFlatArray`, would this fly?
```
const emojo = List([ 'some-thing', ['else', [43]], 34])
t.same.(emojo, ['some-thing', 'else', 43, 34]`, 'flatten and folds wrapped array'
t.end()
```

crocks/List.spec.js
```diff
@@ -383,6 +392,28 @@ test('List of', t => {
...
+  t.throws(fromArray(undefined), TypeError, 'throws with undefined')
+  t.throws(fromArray(null), TypeError, 'throws with null')
+  t.throws(fromArray(0), TypeError, 'throws with falsey number')
+  t.throws(fromArray(1), TypeError, 'throws with truthy number')
+  t.throws(fromArray(''), TypeError, 'throws with falsey string')
+  t.throws(fromArray('string'), TypeError, 'throws with truthy string')
+  t.throws(fromArray(false), TypeError, 'throws with false')
+  t.throws(fromArray(true), TypeError, 'throws with true')
+  t.throws(fromArray({}), TypeError, 'throws with an object')
+  t.throws(fromArray(notList), TypeError, 'throws when passed non-List')
```
evilsoft: This test does not matter. we are not checking for a `List` specifically

crocks/List.spec.js
```diff
@@ -119,6 +120,14 @@ test('List value', t => {
   t.end()
 })
 
+test('List toArray', t => {
+  const a = List([ 'some-thing', ['else', 43], 34 ]).toArray()
+
+  t.same(a, [ 'some-thing', 'else', 43, 34 ], 'folds the wrapped array')
```
evilsoft: This is a BAD test and actually shows this is broken. (the issue is your reduce and how you are doing concat), but that will get fixed with `.slice()`

```js
const data = [ 'some-thing', ['else', 43], 34 ]
const a = List(data)

t.same(a, data, 'folds the wrapped array')
```

The inner structure MUST match exactly.

crocks/List.js
```diff
@@ -27,6 +27,13 @@ const _of =
 const _empty =
   () => List([])
 
+function fromArray(xs) {
+  if(!arguments.length || !isArray(xs)) {
+    throw new TypeError('List.fromArray: Must wrap an array')
```
evilsoft: The error should read:
`List.fromArray: Array required`

crocks/List.spec.js
```diff
@@ -383,6 +392,28 @@ test('List of', t => {
   t.end()
 })
 
+test('List fromArray', t => {
+  const fromArray = bindFunc(List.fromArray)
+
+  const notList = { type: constant('List...Not') }
```
evilsoft: This can go away.

crocks/List.spec.js
evilsoft: That is more something you can do after extraction.
No need to add it here.

crocks/List.spec.js
rstegg: Whoops!

crocks/List.js
rstegg: exactly as value?

evilsoft: Yep. It could even be: 
const toArray =
  value

crocks/List.spec.js
```diff
@@ -123,7 +123,7 @@ test('List value', t => {
 test('List toArray', t => {
   const a = List([ 'some-thing', ['else', 43], 34 ]).toArray()
 
-  t.same(a, [ 'some-thing', 'else', 43, 34 ], 'folds the wrapped array')
+  t.same(a, [ 'some-thing', ['else', 43], 34 ], 'provides the wrapped array')
```
rstegg: Is this a bad test because of the nested arrays, or is it a bad test because it was failing?  I like how this shows the preserving of structure [now that `toArray` is implemented correctly].

evilsoft: It is a bad test because it was wrong. It was not testing the right thing. Also because you did not take my suggestion it is now a brittle test

crocks/List.js
```diff
@@ -52,7 +59,7 @@ function runTraverse(f) {
 }
 
 function List(xs) {
-  if(!arguments.length || !isArray(xs)) {
+  if(!isArray(xs)) {
```
evilsoft: 👍

PR #67, opened 2017-02-17
Fixes IO API

rstegg:
Moves file from `docs` to `docs/crocks` 🤕

PR #71, opened 2017-02-17
API: Cleanup & Fix typos

rstegg:
Using new branch, combines screw ups #66 #67 #68 #69 #70

iwkad22


evilsoft:
Would it be possible to add those `alt` definitions to this PR.
from gitter:
Also to Either and Maybe lets add another called alt, like this
`Maybe m => m a ~> m a -> m a`
`Either m => m c a ~> m c a -> m c a`
And a function called zero on both the constructor and instance for Maybe, like this:
`Maybe m => () -> m a`
`Maybe m => m a ~> () -> m a`

Also another thing missed in the multiple reviews (sorry) is that `type` is available on every Constructor for every type. Should probably add those as well while were doing such things.

evilsoft:
Also `map` for `Either` is wrong. It should be:

`Either m => m c a ~> (a -> b) -> m c b`

rstegg:
> Also another thing missed in the multiple reviews (sorry) is that type is available on every Constructor for every type.

I don't see the logic in `type` being a constructor, can you explain?

evilsoft:
Sure thing.

So if I need to see if something is a `Maybe`, I can use `isSameType` and pass either a constructor or an instance.

```js
const onlyEithers =
  safe(isSameType(Either))

onlyEithers(Right(23)) // Just Right 23
onlyEithers(5) // Nothing
onlyEithers(Identity(4)) // Nothing
```
Make sense?

evilsoft:
To avoid potential confusion, I think we should keep the headings in the columns the same, but change the section headings to read:

"Available on the Constructor"
"Available on the Instance"

docs/crocks/Either.md
```diff
@@ -10,80 +10,84 @@
 --
 ```
 
-`Either` exposes the following constructor and instances:
+`Either` exposes the following functions on the constructor and instance:
 
 | Constructor | Instance |
 |:---|:---|
-| [`Left`](#left), [`Right`](#right), [`of`](#of) | [`inspect`](#inspect), [`value`](#value), [`type`](#type), [`equals`](#equals), [`map`](#map), [`ap`](#ap), [`of`](#of), [`chain`](#chain), [`sequence`](#sequence), [`traverse`](#traverse) |
+| [`Left`](#left), [`of`](#of), [`Right`](#right), [`type`](#type) | [`alt`](#alt), [`ap`](#ap), [`bimap`](#bimap), [`chain`](#chain), [`coalesce`](#coalesce), [`either`](#either), [`equals`](#equals), [`inspect`](#inspect), [`map`](#map),  [`of`](#of), [`sequence`](#sequence), [`swap`](#swap), [`traverse`](#traverse), [`type`](#type) |
```
evilsoft: What do you think about maybe doing the tagged construction functions first, in a left to right order.
then `of` if it is there, and end with `type`.

So here it would be `Left`, `Right`, `of`, `type`.

And then keep the instance functions alphabetized?

docs/crocks/Maybe.md
```diff
@@ -10,67 +10,75 @@
 --
 ```
 
-`Maybe` exposes the following constructor and instances:
+`Maybe` exposes the following functions on the constructor and instance:
 
 | Constructor | Instance |
 |:---|:---|
-| [`Nothing`](#nothing), [`Just`](#just), [`of`](#of) | [`inspect`](#inspect), [`maybe`](#maybe), [`either`](#either), [`option`](#option), [`type`](#type), [`equals`](#equals), [`coalesce`](#coalesce), [`map`](#map), [`ap`](#ap), [`of`](#of), [`chain`](#chain), [`sequence`](#sequence), [`traverse`](#traverse) |
+| [`Just`](#just), [`Nothing`](#nothing), [`of`](#of), [`type`](#type), [`zero`](#zero) | [`alt`](#alt), [`ap`](#ap), [`chain`](#chain), [`coalesce`](#coalesce), [`either`](#either), [`equals`](#equals), [`inspect`](#inspect), [`map`](#map), [`maybe`](#maybe), [`of`](#of), [`option`](#option), [`sequence`](#sequence), [`traverse`](#traverse), [`type`](#type) |
```
evilsoft: So here it would be `Nothing`, `Just`, `of`, `zero`, `type`.

rstegg: I did a lot of thinking on this, and I will post on gitter my conclusion;

docs/crocks/Either.md
rstegg: After playing around and over-thinking this, I think we need to keep it all alphabetized for :corn:; we can have a `API: Phase 2` to fill in gaps like these

rstegg: The only guy that is hurt by this is `Either`, but even so check it out:

```
### Left 

[ sig a ]

### of

[ sig b ]

### Right

[ sig b ]

### type

[ sig c ]
```

So we actually don't really don't lose much by re-ordering by abc, as face-value suggests

rstegg: and like you said, people are coming to the API to look for stuff, let's maybe hold off on anything that can lead to confusion

evilsoft: I would say, lets go full on Alphabetical. Capital letters in ASCII come first, so it would be something like:
* `Maybe`: `Just`, `Nothing`, `of`, `type`, `zero`
* `Either`: `Left`, `Right`, `of` `type`

And I think that should do it.

docs/crocks/Maybe.md
rstegg: ASCII J = 74, N = 78, they are in the correct order

rstegg: `Just`, `Nothing` is correct here

evilsoft:
Looks good...have a squirrel!!

PR #72, opened 2017-02-18
Adds Const API

rstegg:


docs/crocks/Const.md
```diff
@@ -0,0 +1,55 @@
...
+`Const` exposes the following constructor and instance functions:
+
+| Constructor | Instance |
+|:---|:---|
+| -- | [`inspect`](#inspect), [`value`](#value), [`type`](#type), [`equals`](#equals), [`concat`](#concat), [`map`](#map), [`ap`](#ap), [`chain`](#chain) |
+
+## Constructors
+
+--
+
```
evilsoft: Need to add a `type` function to the constructor

rstegg: Is this new?

evilsoft: Nope. Just missed it on the past reviews. They have always been there

evilsoft:
To avoid potential confusion, I think we should keep the headings in the columns the same, but change the section headings to read:

"Available on the Constructor"
"Available on the Instance"

PR #76, opened 2017-02-19
Adds Pred API

rstegg:


docs/crocks/Pred.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Instance
+
+### concat
+
+`Pred m => m a ~> a -> m a`
+
+### contramap
+
+`Pred m => m a ~> a -> m ()`
```
evilsoft: Now this one I KNEW was going to trip you up, because the signature is strange a heck.

It looks like this:
`Pred m => m a ~> (b -> a) -> m b`

So contravarience is use for things like: I have a function `(Number -> Boolean)`, but I do not have `Numbers`, say I have an `Object` that has a `Number` on it. I need to adapt my Pred somehow: 
```js

// pred : Pred Number
const pred = Pred(x => x > 10)

// objPred : Pred { 'num' }
const objPred =
  pred
    .contramap(compose(option(0), prop('num'))

console.log(pred.runWith(3)) // false
console.log(pred.runWith(30)) // true

console.log(objPred.runWith({ num: 8 })) // false
console.log(objPred.runWith({})) // false
console.log(objPred.runWith({ num: 11 })) // true
```

docs/crocks/Pred.md
```diff
@@ -0,0 +1,57 @@
...
+
+### empty
+
+`Pred m => () ~> m ()`
```
evilsoft: Same comment as above please

docs/crocks/Pred.md
```diff
@@ -0,0 +1,57 @@
...
+
+### inspect
+
+`() => String`
+
+### runWith
+
+`Pred m => m a ~> () -> a`
```
evilsoft: So `runWith` literally means *run* this *with* this value. That is the difference between `run` and `runWith`.

This should be:
`Pred m => m a ~> a -> Boolean`

docs/crocks/Pred.md
```diff
@@ -0,0 +1,57 @@
...
+
+### type
+
+`() -> String`
+
+### value
+
+`Pred m => m b ~> () -> b`
```
evilsoft: So this is a strange one, and it is only there to work with Monoidial reduction.
This is going to extract the inner function. So really this returns the function `(a -> Boolean)`.
So this should look like:

`Pred m => m a ~> () -> (a -> Boolean)`

we do not auto-curry `value` that is why it is not `() -> a -> Boolean`

docs/crocks/Pred.md
```diff
@@ -0,0 +1,57 @@
...
+
+### type
+
+`() -> String`
+
+## Instance
+
+### concat
+
+`Pred m => m a ~> a -> m a`
```
rstegg: dang, i had that but i thought, "It can't be the same as Const"

docs/crocks/Pred.md
rstegg: Hah yeah this one was strange, i tried to make sense of it.

rstegg:
Thanks! Your review has cleared up a lot of lingering confusions.

PR #79, opened 2017-02-20
Adds List API

rstegg:
homers-bucket-list

### DIFF
--- a/docs/crocks/List.md
+++ b/docs/crocks/List.md
@@ -1,0 +1,111 @@
+# List
+
+`List a`
+
+--
+
+--
+
+```js
+--
+```
+
+`List` exposes the following functions on the constructor and instance:
+
+| Constructor | Instance |
+|:---|:---|
+| [`empty`](#empty), [`fromArray`](#fromArray), [`of`](#of) | [`ap`](#ap), [`chain`](#chain), [`concat`](#concat), [`cons`](#cons), [`empty`](#empty), [`equals`](#equals), [`filter`](#filter), [`head`](#head), [`map`](#map), [`of`](#of), [`reduce`](#reduce), [`reject`](#reject), [`sequence`](#sequence), [`tail`](#tail), [`toArray`](#toArray), [`traverse`](#traverse), [`value`](#value) |
+
+## Constructors
+
+### empty
+
+`List m => () -> m a`
+
+### fromArray
+
+`List m => [ a ] -> m a`
+
+### of
+
+`List m => a -> m a`
+
+## Instances
+
+## Instance
+
+### ap
+
+`List m => m (a -> b) ~> m a -> m b`
+
+### chain
+
+`List m => m a ~> (a -> m b) -> m b`
+
+### concat
+
+`List m => m a ~> m a -> m a`
+
+### cons
+
+`List m => m a ~> a -> m a`
+
+### empty
+
+`List m => () -> m a`
+
+### equals
+
+`a -> Boolean`
+
+### filter
+
+`List m => m a ~> ((a -> Boolean) | Pred a) -> m a`
+
+### head
+
+`List m => m a ~> () -> Maybe a`
+
+### inspect
+
+`() -> String`
+
+### map
+
+`List m => m a ~> (a -> b) -> m b`
+
+### of
+
+`List m => a -> m a`
+
+### reduce
+
+`List m => m a ~> (((b, a) -> b), b) -> b`
+
+### reject
+
+`List m => m a ~> ((a -> Boolean) | Pred a) -> m a`
+
+### sequence
+
+`List m, Applicative f => m (f a) ~> (b -> f b) -> f (m a)`
+
+### tail
+
+`List m => m a ~> () -> Maybe (m a)`
+
+### toArray
+
+`List m => m a ~> () -> [ a ]`
+
+### traverse
+
+`List m, Applicative f => m a ~> ((c -> f c), a -> f b) -> f (m b)`
+
+### type
+
+`() -> String`
+
+### value
+
+`List m => m a ~> () -> [ a ]`


evilsoft:
This needs to be against the `docs` branch.

rstegg:
derp

docs/crocks/List.md
```diff
@@ -0,0 +1,111 @@
...
+
+| Constructor | Instance |
+|:---|:---|
+| [`empty`](#empty), [`fromArray`](#fromArray), [`of`](#of) | [`ap`](#ap), [`chain`](#chain), [`concat`](#concat), [`cons`](#cons), [`empty`](#empty), [`equals`](#equals), [`filter`](#filter), [`head`](#head), [`map`](#map), [`of`](#of), [`reduce`](#reduce), [`reject`](#reject), [`sequence`](#sequence), [`tail`](#tail), [`toArray`](#toArray), [`traverse`](#traverse), [`value`](#value) |
+
+## Constructors
+
+### empty
+
+`List m => () -> m ()`
```
evilsoft: Even though the List is empty, it still needs to align with a type. This should be:
`List m => () -> m a`

docs/crocks/List.md
```diff
@@ -0,0 +1,111 @@
...
+
+### fromArray
+
+`List m => a -> m a`
```
evilsoft: `a` would be the contained type, so we would need to do something like:
`List m => [ a ] -> m a`

To move an `Array` of `a` to a `List` of `a`

docs/crocks/List.md
```diff
@@ -0,0 +1,111 @@
...
+
+`List m => m a ~> (a -> m b) -> m b`
+
+### concat
+
+`List m => m a ~> m a -> m a`
+
+### cons
+
+`List m => m a ~> (b -> m b) -> m a`
```
evilsoft: `cons` adds an element to the front of the `List` so this should be:

`List m => m a ~> a -> m a`

docs/crocks/List.md
```diff
@@ -0,0 +1,111 @@
...
+
+### empty
+
+`List m => () -> m ()`
```
evilsoft: Same comment as above.

docs/crocks/List.md
```diff
@@ -0,0 +1,111 @@
...
+
+### equals
+
+`a -> Boolean`
+
+### filter
+
+`List m => m a ~> (a -> Boolean) -> m a`
```
evilsoft: Filter can take a `Pred` or a predicate function. This should be:

`List m => m a ~> ((a -> Boolean) | Pred a) -> m a`

docs/crocks/List.md
```diff
@@ -0,0 +1,111 @@
...
+
+### head
+
+`List m => m a ~> m a -> a`
```
evilsoft: `head` returns a `Maybe a`, to account for an empty `List` and takes no arguments. This should be:

`List m => m a ~> () -> Maybe a`

rstegg:
I was clueless on:

- cons
- head / tail
- reduce

docs/crocks/List.md
```diff
@@ -0,0 +1,111 @@
...
+
+`List m => m a ~> (a -> b) -> m b`
+
+### of
+
+`List m => a -> m a`
+
+### reduce
+
+`List m => m a ~> (a -> b) -> b -> m a`
```
evilsoft: `reduce` takes a binary function as its first argument and is NOT curried. Also it fold out a `b` so this should be something like:

`List m => m a ~> (((b, a) -> b), b) -> b`

docs/crocks/List.md
```diff
@@ -0,0 +1,111 @@
...
+
+### reject
+
+`List m => m a ~> m (a -> Boolean) -> m a`
```
evilsoft: Like `filter`, `reject` takes wither a predicate function or a `Pred` so this should be:

`List m => m a ~> ((a -> Boolean) | Pred a) -> m a`

docs/crocks/List.md
```diff
@@ -0,0 +1,111 @@
...
+
+### sequence
+
+`List m, Applicative f => m (f a) ~> (b -> f b) -> f (m a)`
+
+### tail
+
+`List m => m a ~> m a -> a -> m a`
```
evilsoft: `tail` returns a `Maybe` in the case that there is only 1 or 0 elements in the `List`. Also it take no arguments, so this should be:

`List m => m a ~> () -> Maybe (m a)`

docs/crocks/List.md
```diff
@@ -0,0 +1,111 @@
...
+
+### toArray
+
+`List m => m a ~> () -> a`
```
evilsoft: This appears to only be returning an element, we need to return an `Array` of `a`. So this should be:

`List m  => m a ~> () ->  [ a ]`

docs/crocks/List.md
```diff
@@ -0,0 +1,111 @@
...
+
+`List m, Applicative f => m a ~> (c -> f c) -> (a -> f b) -> f (m b)`
+
+### type
+
+`() -> String`
+
+### value
+
+`List m => m a ~> () -> a`
```
evilsoft: This is here only for `Monoid` bits like `mreduce` so eventhough it looks like `toArray` it is signalling something, so this should match `toArray` and be:

`List m  => m a ~> () -> [ a ]`

docs/crocks/List.md
rstegg: > `(((b, a) -> b), b)` 

I was hesitant on this comma notation

docs/crocks/List.md
rstegg: Hesitant on adding `Maybe`

docs/crocks/List.md
rstegg: Hesitant on adding `Pred`

docs/crocks/List.md
rstegg: Hesitant on `[ ]`

docs/crocks/List.md
evilsoft: That is how one shows uncurried function calls.

docs/crocks/List.md
evilsoft: That is what it returns. So that is what we show. Just try to match what the source code says please!!

docs/crocks/List.md
evilsoft: That is what the signature is. So......

docs/crocks/List.md
evilsoft: That is what the signature is. So......

evilsoft:
🍳 cellent!

PR #84, opened 2017-02-24
Adds Reader API

rstegg:


docs/crocks/Reader.md
```diff
@@ -0,0 +1,61 @@
+# Reader
+
+`Reader a`
```
evilsoft: `Reader` provides an environment on its left. Usually it is denoted with an `e`

docs/crocks/Reader.md
```diff
@@ -0,0 +1,61 @@
...
+
+| Constructor | Instance |
+|:---|:---|
+| [`ask`](#ask), [`of`](#of), [`type`](#type) |  [`ap`](#ap), [`chain`](#chain), [`inspect`](#inspect), [`map`](#map), [`of`](#of), [`run`](#runWith), [`type`](#type) |
+
+## Constructor
+
+### ask
+
+`Reader m => a -> m a`
```
evilsoft: Ask takes a function as its input. this is used a lot for chaining.
It asks for what the environment is and passes it to function you provide.
So looking below, that x is the value that is passed into `f`. `f` lifts it into a `Reader`. This is the `a` in the siggy.
So anytimne that `ask` is used it will always pass the `e` to it. In this case `5`.

But instead of numbers think of `e` being an object that holds a database connection, or some global configuration that is used by a family of functions in a flow.

```js
const { add, multiply } = require('ramda')
const { Reader, compose, chain } = require('../crocks')

const { ask } = Reader

const f = compose(
  chain(x => ask(multiply(x))),
  chain(x => ask(add(x))),
  Reader.of
)

log(
  f(10).runWith(5)
)

// logs 75
```

docs/crocks/Reader.md
```diff
@@ -0,0 +1,61 @@
...
+
+### of
+
+`Reader m => a -> m a`
```
evilsoft: Need to account for `e` in here.

docs/crocks/Reader.md
```diff
@@ -0,0 +1,61 @@
...
+
+### type
+
+`() -> String`
+
+## Instance
+
+### ap
+
+`Reader m => m (a -> b) ~> m a -> m b`
```
evilsoft: Need to account for `e` in here

docs/crocks/Reader.md
```diff
@@ -0,0 +1,61 @@
...
+
+### chain
+
+`Reader m => m a ~> m a -> (a -> m b) -> m b`
```
evilsoft: Need to account for `e` in here

docs/crocks/Reader.md
```diff
@@ -0,0 +1,61 @@
...
+
+### inspect
+
+`() -> String`
+
+### map
+
+`Reader m => m a ~> (a -> b) -> m b`
```
evilsoft: Need to account for `e` in here

docs/crocks/Reader.md
```diff
@@ -0,0 +1,61 @@
...
+
+### of
+
+`Reader m => a -> m a`
```
evilsoft: Need to account for `e` in here

docs/crocks/Reader.md
```diff
@@ -0,0 +1,61 @@
...
+
+### runWith
+
+`Reader m => m a ~> () -> a`
```
evilsoft: `runWith` is different then `run`. `runWith` will always take a value. where `run` is always `()`

This is how `e` gets into the reader, need to show that.

docs/crocks/Reader.md
rstegg: If something applies to all the sigs, if you prefer you can rather add it to the review note instead of to each line

docs/crocks/Reader.md
```diff
@@ -0,0 +1,61 @@
...
+
+| Constructor | Instance |
+|:---|:---|
+| [`ask`](#ask), [`of`](#of), [`type`](#type) |  [`ap`](#ap), [`chain`](#chain), [`inspect`](#inspect), [`map`](#map), [`of`](#of), [`run`](#runWith), [`type`](#type) |
+
+## Constructor
+
+### ask
+
+`Reader m => m e ~> a -> m e a`
```
evilsoft: Don't need the `m e ~>`

docs/crocks/Reader.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Instance
+
+### ap
+
+`Reader m => m e (a -> b) ~> m e a -> m e b`
+
+### chain
+
+`Reader m => m e a ~> m e a -> (a -> m b) -> m e b`
```
evilsoft: chain takes a function that returns a `Reader`

docs/crocks/Reader.md
evilsoft: and `ask` takes a function that will receive the environment
(check the source code to see how `fn` is used)

docs/crocks/Reader.md
```diff
@@ -0,0 +1,61 @@
...
+
+`Reader m => m e a ~> (a -> b) -> m e b`
+
+### of
+
+`Reader m => a -> m e a`
+
+### runWith
+
+`Reader m => m a ~> e -> m e a`
```
evilsoft: `runWith` does not return another `Reader`

docs/crocks/Reader.md
```diff
@@ -0,0 +1,61 @@
...
+
+`Reader m => m e a ~> (a -> b) -> m e b`
+
+### of
+
+`Reader m => a -> m e a`
+
+### runWith
+
+`Reader m => m a ~> (e -> a) -> a`
```
evilsoft: `runWith` does not take a function as it's input

docs/crocks/Reader.md
```diff
@@ -0,0 +1,61 @@
...
+
+`Reader m => m e a ~> (a -> b) -> m e b`
+
+### of
+
+`Reader m => a -> m e a`
+
+### runWith
+
+`Reader m => m a ~> e -> a`
```
evilsoft: need to represent the `e` in as in `m e a`

PR #85, opened 2017-02-25
Add `concat` to `Maybe`, `Either` and `Identity`

evilsoft:
## Lets just start with these

This PR was requested by an early tester of things. It adds `concat` to the simple value types and will concat inner values if they contain `Semigroups` of the same type.

Also there is a BIG ol breaking change: I updated `isSameType` to be able to compare native JS types, not just the `crocks` types. So things like this can be done:

* `isSameType(String, 'hello') // true`
* `isSameType('goodbye', 'hello') // true`
* `isSameType('goodbye', Array) // false`
* `isSameType([ 'a' ], 'a') // false`

### DIFF
--- a/internal/innerConcat.js
+++ b/internal/innerConcat.js
@@ -1,0 +1,26 @@
+/** @license ISC License (c) copyright 2017 original and current authors */
+/** @author Ian Hofmann-Hicks (evil) */
+
+const curry = require('../helpers/curry')
+const isSameType = require('../predicates/isSameType')
+const isSemigroup = require('../predicates/isSemigroup')
+
+function innerConcat(type, m) {
+  const t = type.type()
+
+  return function(left) {
+    if(!isSemigroup(left)) {
+      throw new TypeError(`${t}.concat: Both containers must contain Semigroups of the same type`)
+    }
+
+    return m.map(right => {
+      if(!isSameType(left, right)) {
+        throw new TypeError(`${t}.concat: Both containers must contain Semigroups of the same type`)
+      }
+
+      return left.concat(right)
+    })
+  }
+}
+
+module.exports = curry(innerConcat)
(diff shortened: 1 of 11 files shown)


internal/innerConcat.js
```diff
@@ -0,0 +1,26 @@
...
+
+  return function(left) {
+    if(!isSemigroup(left)) {
+      throw new TypeError(`${t}.concat: Both containers must contain Semigroups of the same type`)
+    }
+
+    return m.map(right => {
+      if(!isSameType(left, right)) {
+        throw new TypeError(`${t}.concat: Both containers must contain Semigroups of the same type`)
+      }
```
rstegg: `M(String).concat(M(number))` or `M(Array).concat(M(number))` should work, so should be:
```
const rightT = left.concat(right)
    return m.map(right => {
       if(!isSameType(left, rightT)) {
         throw new TypeError(`${t}.concat: Both containers must contain Semigroups of the same type`)
       }
```
What is your thought?

rstegg: and if rightT (right testable) is undefined, `isSameType` will still throw;

evilsoft: If I think I understand what you are saying, I think you are saying that this would not throw an Error:
```js
const m = Maybe.of('hello')
const n = Maybe.of(3)

// will fail as Number is NOT a semigroup
m.concat(n)
```

if you look we are mapping on `m`, which is the container passed into us. So `right` will be a `Number` and it is not the same type as `String`, so this fails.

Now if you are saying that they are passing in the actual constructors `Array` and `String`
That would fail because `Array` is not a semigroup.

Is that what you mean?

evilsoft: What would this possible fix actually fix.
Do you have a working example that shows what you found that is broken?
It is hard to pin down what you found to be broken with that description.

evilsoft: Or are you saying that it should not error?

evilsoft: From gitter chat, it seems that you meant that I should be able to [ 'a' ].concat('b'). While that is true in JS, for these types of interaction I am going to go with the FL spec for semigroup.

So I do not see an error here.

evilsoft:
Seems legit. Merging!

PR #91, opened 2017-02-27
`Pred` should only return `Boolean`s

evilsoft:
## Through the cracks

Looks like we were not returning `Boolean` values from runWith in `Pred`, we should.
This PR fixes that.

### DIFF
--- a/crocks/Pred.js
+++ b/crocks/Pred.js
@@ -15,10 +15,13 @@ const _type =
 const _empty =
   () => Pred(constant(true))
 
-function Pred(runWith) {
-  if(!isFunction(runWith)) {
+function Pred(pred) {
+  if(!isFunction(pred)) {
     throw new TypeError('Pred: Predicate function required')
   }
+
+  const runWith =
+    x => !!pred(x)
 
   const type =
     _type
@@ -37,7 +40,7 @@ function Pred(runWith) {
       throw new TypeError('Pred.concat: Pred required')
     }
 
-    return Pred(x => runWith(x) && m.runWith(x))
+    return Pred(x => !!runWith(x) && !!m.runWith(x))
   }
 
   function contramap(fn) {
--- a/crocks/Pred.spec.js
+++ b/crocks/Pred.spec.js
@@ -56,11 +56,11 @@ test('Pred type', t => {
 })
 
 test('Pred value', t => {
-  const f = constant('some Predicate')
+  const f = constant('')
   const p = Pred(f)
 
   t.ok(isFunction(p.value), 'is a function')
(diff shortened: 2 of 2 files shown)


crocks/Pred.js
```diff
@@ -37,7 +40,7 @@ function Pred(runWith) {
       throw new TypeError('Pred.concat: Pred required')
     }
 
-    return Pred(x => runWith(x) && m.runWith(x))
+    return Pred(x => !!runWith(x) && !!m.runWith(x))
```
rstegg: what is this locking down?  lack of errors will make this part confusing to debug [e.g. if a curried function returned is going to always return true]

evilsoft: Good question.
We lock the result of `Pred` to a `Boolean`.
We need to return a `Boolean`.
Now as far as the Error, it is perfectly acceptable to have functions that return non-boolean values and be interpreted as Booleans. So if we were to throw, then that means they would have to ensure that all functions they wrap will need to return a boolean. That means if they have existing functions they want to use as Pred, then they will need to create wrappers.

Does not seem very JS to me to do that.
Thoughts?

rstegg: so having a `toPredicate` transformer? or, 

```
!isSameType(m.runWith(x), runWith(x)) { throw new TypeError("Pred.concat: return values must be of same semigroup") }
```

evilsoft: I would rather not TBH. Just seems like fluff and an extra step we would put on the consumer.
No reason to add all this. One thing we could do is mention in the docs that it will coerce  your return value into a Boolean. 
The point of this framework is not really for correctness, it is to make things easier on the dev. We say that we are going to return a Boolean, so a Boolean we must return.

rstegg: ok gotcha, looks good then!

evilsoft: What do you thin about the docs mention?

And I just realized, I need to push a fix to the spec. we do not have coverage over the function returned from `value`. I deleted it and forgot to replace it with an appropriate test.

How embarrassing for me 😲

evilsoft: Actually, no need to do the `!!` here. That is just extra, as we already do it in `runWith`

evilsoft:
This PR addresses [issue#92](https://github.com/evilsoft/crocks/issues/92)

evilsoft:
Updated to provide better coverage around the coercion, so this will not slip past again.

crocks/Pred.js
evilsoft: This is ready for another go around, please let me know if you feel good about the coverage around this change. Do not want this getting introduced again

PR #94, opened 2017-02-27
Arrow api, fixes Pred.concat

rstegg:
remove-tooth-bow-arrow-genius-13614095010

### DIFF
--- a/docs/crocks/Arrow.md
+++ b/docs/crocks/Arrow.md
@@ -1,0 +1,77 @@
+# Arrow
+
+`Arrow a b`
+
+--
+
+--
+
+```js
+--
+```
+
+`Arrow` exposes the following functions on the constructor and instance:
+
+| Constructor | Instance |
+|:---|:---|
+| [`empty`](#empty), [`type`](#type) | [`both`](#both), [`concat`](#concat), [`contramap`](#contramap), [`empty`](#empty), [`first`](#first), [`inspect`](#inspect), [`map`](#map), [`promap`](#promap), [`runWith`](#runWith), [`second`](#second), [`type`](#type), [`value`](#value) |
+
+## Constructor
+
+### empty
+
+`Arrow m => () -> m a a`
+
+### type
+
+`() -> String`
+
+## Instance
+
+### both
+
+`Arrow m, Pair p => m a b ~> () -> m (p a a) (p b b)`
+
+### concat
+
+`Arrow m => m a b ~> m a b -> m a b`
+
+### contramap
+
+`Arrow m => m a b ~> (c -> a) -> m c b`
+
+### empty
+
+`Arrow m => () -> m a a`
+
+### first
+
+`Arrow m, Pair p => m a b ~> () -> m (p a c) (p b c)`
+
+### inspect
+
+`() => String`
+
+### map
+
+`Arrow m => m a b ~> (b -> c) -> m a c`
+
+### promap
+
+`Arrow m => m a b ~> ((c -> a), (b -> d)) -> m c d`
+
+### runWith
+
+`Arrow m => m a b ~> a -> b`
+
+### second
+
+`Arrow m, Pair p => m a b ~> () -> m (p c a) (p c b)`
+
+### type
+
+`() -> String`
+
+### value
+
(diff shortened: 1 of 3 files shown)


docs/crocks/Arrow.md
```diff
@@ -0,0 +1,77 @@
...
+
+| Constructor | Instance |
+|:---|:---|
+| [`empty`](#empty), [`type`](#type) | [`both`](#both), [`concat`](#concat), [`contramap`](#contramap), [`empty`](#empty), [`first`](#first), [`inspect`](#inspect), [`map`](#map), [`promap`](#promap), [`runWith`](#runWith), [`second`](#second), [`type`](#type), [`value`](#value) |
+
+## Constructor
+
+### empty
+
+`Arrow m => () -> m () b`
```
evilsoft: The identity or `empty` for `Arrow` is an identity function, so this should be:

`Arrow m => () -> m a a`

docs/crocks/Arrow.md
```diff
@@ -0,0 +1,77 @@
...
+
+### type
+
+`() -> String`
+
+## Instance
+
+### both
+
+`Arrow m, Pair p => m a b ~> () -> m (p a b) (p c d)`
```
evilsoft: `both` applies the wrapped computation to both parts of the `Pair`. so it should look like:

`Arrow m, Pair p => m a b ~> () -> m (p a a) (p b b)`

docs/crocks/Arrow.md
```diff
@@ -0,0 +1,77 @@
...
+
+### concat
+
+`Arrow m => m a b ~> m a b -> m a b`
```
evilsoft: This would be something like this as we are not constrained to an mono-morphisim here and it is composition under the hood:
`Arrow m => m a b ~> m b c -> m a c`

We will keep it for now, but now that I see this, this should not be `concat`. Will need to do something else.

docs/crocks/Arrow.md
```diff
@@ -0,0 +1,77 @@
...
+
+### contramap
+
+`Arrow m => m a b ~> (b -> a) -> m c b`
```
evilsoft: So `contramap` works on the input of the `Arrow`, should be:

`Arrow m => m a b ~> (c -> a) -> m c b`

docs/crocks/Arrow.md
```diff
@@ -0,0 +1,77 @@
...
+
+### empty
+
+`Arrow m => () -> m () b`
```
evilsoft: Same as above.

docs/crocks/Arrow.md
```diff
@@ -0,0 +1,77 @@
...
+
+`() => String`
+
+### map
+
+`Arrow m => m a b ~> (b -> c) -> m a c`
+
+### promap
+
+`Arrow m => m a b ~> (a -> c, b -> d) -> m c d`
```
evilsoft: The `contramap` function is reversed, would be nice to wrap functions in parentheses.

`Arrow m => m a b ~> ((c -> a), (b -> d)) -> m c d`

docs/crocks/Arrow.md
```diff
@@ -0,0 +1,77 @@
...
+
+### runWith
+
+`Arrow m => m a b ~> (b -> c) -> c`
```
evilsoft: `runWith` executes the computation and well, `runs` it `with` the expected input, returning the final result. Should be something like:

`Arrow m => m a b ~> a -> b`

docs/crocks/Arrow.md
```diff
@@ -0,0 +1,77 @@
...
+
+`Arrow m, Pair p => m a b ~> () -> m (p c a) (p c b)`
+
+### type
+
+`() -> String`
+
+### value
+
+`Arrow m => m a b ~> () -> a`
```
evilsoft: `value` (which will go away once I figure out how to replace `concat`) provides the wrapped function. So this should look like:

`Arrow m => m a b ~> (a -> b)`

docs/crocks/Arrow.md
rstegg: That is the strangest thing.  Good to learn!

Issue #42, opened 2017-01-28
Documentation


evilsoft:
## All the Things

Would like to have these bits get documented with the following general organization in mind:
* Each `crock`  gets its own API file
  * Signature
  * Sales Pitch for using the specific `crock`
  * Table of functions implemented on the `crock` (By Constructor and Instance).
  * Section for Constructor Functions
  * Section for Instance Functions
* One API file for all `Monoids`,
  * High level description of what a Monoid is and how to use them.
  * Examples showing General use cases (pick a couple)
  * Table of common functions shared between them
  * Section for Each Monoid and its API.
* One API file for each of the function classifications.

- [ ] crocks
  - [ ] `Arrow`
  - [ ] `Async`
  - [ ] `Const`
  - [ ] `Either`
  - [ ] `Free` _(still to come)_
  - [ ] `Identity`
  - [ ] `IO`
  - [ ] `List`
  - [ ] `Maybe`
  - [ ] `Pred`
  - [ ] `Reader`
  - [ ] `Star`
  - [ ] `State`
  - [ ] `Unit`
  - [ ] `Validation` _(still to come)_
  - [ ] `Writer`

- [ ] Monoids
 
- [ ] Functions
  - [ ] Combinator functions
  - [ ] Helper functions
  - [ ] Logic functions _(still to come)_
  - [ ] Predicate functions
  - [ ] Pointfree functions

rstegg:
> Sales Pitch for using the specific crock

should be 

> TL;DR of what the specific crock does

dalefrancis88:
@evilsoft are there anyof these I can take off your hands?

StevenLangbroek:
Hey all! I'm starting to work crocks into some of our infrastructure, and think you could make a good case for having some guides or tutorials, particularly about how to combine crocks in common use-cases (e.g., I had to combine state and async and had some trouble figuring out a good / clean approach). Would you be interested in contributions in that area?

dalefrancis88:
Hey man, welcome to the world of `crocks` glad you're here with us 👍 

We're always happy when people want to contribute and help, especially around documentation and guides. We actually have an issue for it here, https://github.com/evilsoft/crocks/issues/196 

If you need any help getting things set up or what to put where just reach out, happy to help where we can.

StevenLangbroek:
Thanks @dalefrancis88, working with crocks has been a pleasure so far, and if the welcome is any indicator, I look forward to the community just as much :).

[...]

PR #98, opened 2017-03-02
Add Star skeleton

evilsoft:
## A Super Star is what you are

This PR adds another skeleton to the docs for this [issue#42](https://github.com/evilsoft/crocks/issues/42). This one for the `Star`. Also in anticipation of the removal of the monoidial implementation on `Arrow` (broken in sooo many ways), this PR also removes the `concat`, `empty` and `value` bits off of `Arrow`

### DIFF
--- a/docs/crocks/Star.md
+++ b/docs/crocks/Star.md
@@ -1,0 +1,61 @@
+# Star
+
+`Star a b` / `Monad m => Star a (m b)`
+
+--
+
+--
+
+```js
+--
+```
+
+`Star` exposes the following functions on the constructor and instance:
+
+| Constructor | Instance |
+|:---|:---|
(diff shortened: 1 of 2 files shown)


evilsoft:
May want to consolidate the types a bit, make it easier to follow.
I mean come on, they are all Monad, we get that!!

docs/crocks/Star.md
```diff
@@ -0,0 +1,61 @@
+# Star
+
+`Monad m => Star a (m b)`
```
rstegg: Let's keep :corn: and only try to say `Star a b`, but here are some ideas I came up with:

- `Star a (Monad b)`

- `Star a b`
   ` . Monad m => b ~> m b`

- `Star a (m b)`
    ` . Monad m`

The superstar inside me says "keep the laws implicit until skeleton is drawn!"

evilsoft: I got the docs. thanks.

rstegg: I thought the new line could better say "under the implication of", the `.` was just a poopie way of indenting.  Great jorb never less!!

PR #113, opened 2017-03-20
Better Linting and Some :lipstick:

evilsoft:
## Spring has sprung, time to :nail_care: 
image

Thought there were default options on and was wondering why some of my code was passing lint when it was horrible. Turns out, things ***MUST*** be specified. This PR update the lint options and adjusts the existing code base accordingly,


evilsoft:
image

PR #114, opened 2017-03-21
`tryCatch` should return a `Result` and not an `Either`

evilsoft:
## Getting results
image

Now that we have the `Result` type, it makes a lot more sense to have the `tryCatch` helper return this new type and not an `Either`.


evilsoft:
seems legit
image

PR #115, opened 2017-03-25
Remove parent type on inspect string for sum types

evilsoft:
## What does that even mean?!?
image

Found that prefixing the type was so gross in practice. After seeing something that @joneshf did with his `toString` function on his Trees read sooo much better. So we gonna drop dat type from the inspect string. So instead of `Maybe.Just 23`, only going to show `Just 23`


evilsoft:
image

Issue #117, opened 2017-04-07
`reject` and `filter` should work with Objects as well


evilsoft:
## So useful
I really enjoy ramda version of `reject` and `filter` and how they work on Objects as well. Should be easy to add that. So let us do that.

evilsoft:
Submitted [PR#118](https://github.com/evilsoft/crocks/pull/118) to address this issue

evilsoft:
Merged [PR#118](https://github.com/evilsoft/crocks/pull/118).
Closing this out.

PR #118, opened 2017-04-09
Update `reject` and `filter` to accept Objects

evilsoft:
## All inclusive!!
image

This PR address [issue#117](https://github.com/evilsoft/crocks/issues/117).

* Adds a type rep for `Object` in the internal folder to implement a common `filter`.
* Update point-free `filter` to work with `Object` values.
* Update point-free `reject` to work with `Object` values.
* fix up some existing specs for both `filter` and `reject`


evilsoft:
image

PR #119, opened 2017-04-17
Add a bunch of Object related functions.

evilsoft:
## Sending some :heart: to all the `Object`s
image

After a few people were using `crocks` on a project, one of the complaints was a lack of :heart: in the Object department. This PR tries to address some of their concerns. I will be adding an `evolve` function similar to [`Ramda's evolve`](http://ramdajs.com/docs/#evolve) function. But for now this PR adds the following:

* `assign`: Like `Object.assign`, but returns a shallow copy and reverses arguments for pointfree application.
* `assoc`: Assosiate a key-value pair to a given object
* `defaultProps`: Default properties in a given object, by specifying an object of defaults.
* `defaultTo`: Not an `Object` function, but handy none the less. Give it a default value and it will use that value when passed a `null`, `undefined` or `NaN`.
* `fromPairs`: Build an `Object` for either a `List` or `Array` of key-values `Pair`s
* `objOf`: Allows you to create an `Object` with a given key and value.
* `omit`: Reject key-value pairs in a given `Object` using an `Array` of keys.
* `pick`: Filter key-values pairs in a given `Object` using an `Array` of keys.
* `toPairs`: Transform an `Object` into a `List` of key-value `Pairs` 

Also updated `map` to map over an Object and `isNil` predicate now returns `true` for `NaN`

All `Object` related functions will omit any keys in which the value is explicitly `undefined`. If for some reason you need to add keys as placeholders, use `null` instead.

As part of the build process, needed to add another babel plugin for computed properties.


README.md
```diff
@@ -153,6 +153,18 @@ substitution : (a -> b -> c) -> (a -> b) -> a -> c
...
+
+#### assoc
+```haskell
+assoc : String -> a -> Object -> Object
+```
+There may come a time when you want to add a key-value pair to an `Object` and want control over how the key and value are applied. That is where `assoc` can come to your aid. Just provide a `String` key and a value of any type to be associated to the key. Finally pass it any `Object` and you will get back a shallow copy with your key-value pair merged in. This will overwrite any exiting keys with new value specified. Used with [`flip`](#flip), you can do some interesting things with this function, give it a play! If you just want to create an `Object` and not concatenate it to another `Object`, [`objOf`](#objOf) may be the function for you.
```
evilsoft: link for `objOf`, needs to be "objof"

README.md
```diff
@@ -153,6 +153,18 @@ substitution : (a -> b -> c) -> (a -> b) -> a -> c
...
 
 ### Helper Functions
+#### assign
+```haskell
+assign : Object -> Object -> Object
+```
+When working with `Object`s, a common operation is to combine (2) of them. This can be accomplished in `crocks` by reaching for `assign`. Unlike the `Object.assign` that ships with JavaScript, this `assign` will combine your `Object`s into a new shallow copy of their merger. `assign` only takes two arguments and will overwrite keys present in the second argument with values from the first. As with most of the `crocks` `Object` based functions, `assign` will omit any key-value pairs that are `undefined`. Check out a related function named [`defaultProps`](#defaultProps) that will only assign values that are `undefined` in the second argument.
```
evilsoft: link for `defaultProps` needs to be "defaultprops"

README.md
```diff
@@ -209,6 +233,12 @@ fanout : Monad m => Star a (m b) -> Star a (m c) -> Star a (m (Pair b c))
...
 
+#### fromPairs
+```haskell
+fromPairs : [ (Pair String a) ] | List (Pair String a) -> Object
+```
+As an inverse to [`toPairs`](#toPairs), `fromPairs` takes either an `Array` or `List` of key-value `Pair`s and constructs an `Object` from it. The `Pair` must contain a `String` in the `fst` and any type of value in the `snd`. The `fst` will become the key for the value in the `snd`. All primitive values are copied into the new `Object`, while non-primitives are references to the original. If you provide an `undefined` values for the second, that `Pair` will not be represented in the resulting `Object`. Also, when if multiple keys share the same name, that last value will be moved over.
```
evilsoft: link for `toPairs` needs to be "topairs"

README.md
```diff
@@ -233,12 +263,30 @@ mreduceMap : Monoid m => m -> (b -> a) -> ([ b ] | List b) -> a
...
+```haskell
+objOf : String -> a -> Object
+```
+If you ever find yourself in a situation where you have a key and a value and just want to combine the two into an `Object`, then it sounds like `objOf` is the function for you. Just pass it a `String` for the key and any type of value, and you'll get back an `Object` that is composed of those two. If you find yourself constantly concatenating the result of this function into another `Object`, you may want to use [`assoc`](#assoc) instead.
+
+#### omit
+```haskell
+omit : ([ String ] | List String) -> Object -> Object
+```
+Sometimes you just want to strip `Object`s of unwanted properties by key. Using `omit` will help you get that done. Just pass it a `Foldable` structure with a series of `String`s as keys and then pass it an `Object` and you will get back not only a shallow copy, but also an `Object` free of any of those pesky `undefined` values. You can think of `omit` as a way to black-list or reject `Object` properties based on key names. This function ignores inherited properties and should only be used with POJOs. If you want to filter or white-list properties rather than reject them, take a look at `[pick](#pick)`.
```
evilsoft: only surround `pick` in backticks.

README.md
```diff
@@ -233,12 +263,30 @@ mreduceMap : Monoid m => m -> (b -> a) -> ([ b ] | List b) -> a
...
 ```haskell
 once : ((*) -> a) -> ((*) -> a)
 ```
 There are times in Javascript development where you only want to call a function once and memo-ize the first result for every subsequent call to that function. Just pass the function you want guarded to `once` and you will get back a function with the expected guarantees.
 
+#### pick
+```haskell
+pick : ([ String ] | List String) -> Object -> Object
+```
+When dealing with `Object`s, sometimes it is necessary to only let some of the key-value pairs on an object through. Think of `pick` as a sort of white-list or filter for `Object` properties. Pass it a `Foldable` structure of `String`s that are the keys you would like to pick off of your `Object`. This will give you back a shallow copy of the key-value pairs you specified. This function will ignore inherited properties and should only be used with POJOs. Any `undefined` values will not be copied over, although `null` values are allowed. For black-listing properties, have a look at `[omit](#omit)`
```
evilsoft: only surround `omit` in backticks

README.md
```diff
@@ -293,6 +341,12 @@ tap : (a -> b) -> a -> a
 ```
 It is hard knowing what is going on inside of some of these ADTs or your wonderful function compositions. Debugging can get messy when you need to insert a side-effect into your flow for introspection purposes. With `tap`, you can intervene in your otherwise pristine flow and make sure that the original value is passed along to the next step of your flow. This function does not guarantee immutability for reference types (`Objects`, `Arrays`, etc), you will need to exercise some discipline here to not mutate.
 
+#### toPairs
+```haskell
+toPairs : Object -> List (Pair String a)
+```
+When dealing with `Object`s, sometimes it makes more sense to work in a `Foldable` structure like a `List` of key-value `Pair`s. `toPairs` provides a means to take an object and give you back a `List` of `Pairs` that have a `String` that represents the key in the `fst` and the value for that key in the `snd`. The primitive values are copied, while non-primitive values are references. Like most of the `Object` functions in `crocks`, any keys with `undefined` values will be omitted from the result. `crocks` provides an inverse to this function named [`fromPairs`](#fromPairs).
```
evilsoft: link for `fromPairs` needs to be "frompairs"

helpers/assign.js
```diff
@@ -0,0 +1,18 @@
...
+/** @author Ian Hofmann-Hicks (evil) */
+
+const curry = require('./curry')
+const isObject = require('../predicates/isObject')
+
+const object = require('../internal/object')
+
+// assign : Object -> Object -> Object
+function assign(x, m) {
+  if(!isObject(x) || !isObject(m)) {
```
evilsoft: That darn De Morgan, should probably be:
```js
if(!(isObject(x) && isObject(m)) {
  ...
}
```

helpers/assoc.js
```diff
@@ -0,0 +1,19 @@
...
+
+// assoc : String -> a -> Object -> Object
+function assoc(key, val, obj) {
+  if(!isString(key)) {
+    throw new TypeError('assoc: String required for first argument')
+  }
+  else if(!isObject(obj)) {
+    throw new TypeError('assoc: Object required for third argument')
+  }
+  return Object.assign({}, obj, { [key]: val })
```
evilsoft: Should use the `assign` in the new object TypeRef.

helpers/fromPairs.js
```diff
@@ -0,0 +1,36 @@
...
+
+  const key = pair.fst()
+  const value = pair.snd()
+
+  if(!isString(key)) {
+    throw new TypeError('fromPairs: String required for fst of every Pair')
+  }
+
+  return value !== undefined
+    ? Object.assign({}, acc, { [key]: value })
```
evilsoft: May be better to directly assign, no need to use `Object.assign` here. 
```js
acc[key] = value
```

helpers/omit.js
```diff
@@ -0,0 +1,28 @@
...
+/** @author Ian Hofmann-Hicks (evil) */
+
+const curry = require('./curry')
+const isFoldable = require('../predicates/isFoldable')
+const isObject  = require('../predicates/isObject')
+
+function omitKeys(keys, obj) {
+  return function(acc, key) {
+    return keys.indexOf(key) === -1 && obj[key] !== undefined
+      ? Object.assign({}, acc, { [key]: obj[key] })
```
evilsoft: May be better to directly assign, no need to use `Object.assign` here. 
```js
acc[key] = obj[key]
```

helpers/pick.js
```diff
@@ -0,0 +1,32 @@
...
+const isObject  = require('../predicates/isObject')
+const isString = require('../predicates/isString')
+
+function pickKeys(obj) {
+  return function(acc, key) {
+    if(!isString(key)) {
+      throw new TypeError('pick: Foldable of Strings is required for first argument')
+    }
+    return key && obj[key] !== undefined
+      ? Object.assign({}, acc, { [key]: obj[key] })
```
evilsoft: May be better to directly assign, no need to use `Object.assign` here. 
```js
acc[key] = obj[key]
```

evilsoft: No need to assign into a new `Object`. should just reuse accumulator and keep creation cost down.

amsross:
Just curious: Is there a benefit to adding these to the library rather than just suggesting/expecting the user to use the ramda counterparts? Are they notably different than said counterparts?

evilsoft:
@amsross that is a great question.
They are pretty much the same with some subtle differences:
* These functions use the `crocks` curry, which differs from ramda in that it will continue to apply until a non-function is returned.
* `assign` is *like* `merge` in ramda, except I have found myself constantly flipping `merge` for it to be pointfree.
* And finally, I have found that when using functions like `pick` and `omit` with other 3rd party libs (like AWS api, some cookie libs, etc), I am forced to `reject(isNil, someObj)` to remove things that have been explicitly set to `undefined`. So for all `Object` related functions (that iterate over some `Object`) strip all `undefined`s. Of course this does ***not*** apply to `map` so it conforms to Functor identity laws.

evilsoft:
image

PR #121, opened 2017-05-07
Add some Function based Helper functions

evilsoft:
## Function, Function what is your Function
image

While we are adding all the `Object` bits, there was a large need for some `Function` based helpers. This PR addressed that need. The following functions where added on this PR:

* `binary` - limit a function of any arity to a curried, binary function
* `nAry` - limit a function of any arity to a curried function of any arity
* `partial` - take any function and any number of arguments and you will get back a function with those arguments applied.
* `unary` - limit a function of any arity to a unary function
* `unit` - no matter what you pass this bad-boy, you will always get `undefined` back
* `empty` - provide the empty value for all `Monoids` as well as builtin `Array`, `Object` and `String`. Can accept either a value or constructor.

With the addition of `nAry` there is no need for `curryN` as `nAry` is a more powerful version. This PR also removes `curryN` from the public API. Although it is still available in internal, as it is used in a few places.

Also with the addition of `unit`, there is no need for the `noop` helper in the test suite. Most of the nasty noise in this PR is due to that change. Sorry about that.


evilsoft:
image

PR #122, opened 2017-05-09
Pointfree fork

amsross:
Add a pointfree `fork` function that looks like so:
```
Async m => (a -> _) -> (b -> _) -> m a b -> _
```

Because `Async.fork` does not currently return a value, this function explicitly also does not return a value.


amsross:
I'm not sure why the code coverage is returning different results on travis than locally. Open to suggestions.

amsross:
I need to update this to accept the cancel from #124

README.md
```diff
@@ -606,6 +606,7 @@ These functions provide a very clean way to build out very simple functions and
 | `filter` | <code>((a -> Boolean) &#124; Pred a) -> m a -> m a</code> |
 | `first` | `m (a -> b) -> m (Pair a c -> Pair b c)` |
 | `fold` | `Semigroup s => m s -> s` |
+| `fork` | `Async m => (a -> _) -> (b -> _) -> m a b -> _` |
```
amsross: I'm unsure how the cancel should be represented here. Any thoughts, @evilsoft?

README.md
```diff
@@ -606,6 +606,7 @@ These functions provide a very clean way to build out very simple functions and
 | `filter` | <code>((a -> Boolean) &#124; Pred a) -> m a -> m a</code> |
 | `first` | `m (a -> b) -> m (Pair a c -> Pair b c)` |
 | `fold` | `Semigroup s => m s -> s` |
+| `fork` | `Async m => (a -> _) -> (b -> _) -> () -> m a b -> _` |
```
amsross: I'm unsure how the cancel should be represented here. Any thoughts, @evilsoft?

evilsoft: hmmm. thinking it is a function that is basically a `unit` with a wonderful little side-effect. But since we cannot represent "hey this is a side-effect" in a type siggy, what do you think of something like:
```haskell
Async m => (a -> ()) -> (b -> ()) -> m a b -> (() -> ())
```
We will have to do some explaining when we do the documentation. (sooo much documentation needed)

What do you think about that?

pointfree/fork.js
```diff
@@ -0,0 +1,23 @@
+/** @license ISC License (c) copyright 2017 original and current authors */
+/** @author Matt Ross (amsross) */
```
evilsoft: 👍

pointfree/fork.spec.js
```diff
@@ -0,0 +1,88 @@
...
+  t.ok(m.fork.calledWith(rej, res), 'calls fork on forkable passing the rej, res, and cancel functions')
+  rej.reset()
+  res.reset()
+  cancel.reset()
+  m.fork.reset()
+
+  fork(rej, res, null, m)
+
+  t.ok(m.fork.calledWith(rej, res), 'calls fork on forkable passing the rej and res functions')
+  rej.reset()
```
evilsoft: curious the reason to reset all of these?
Just good form?

README.md
evilsoft: Unless you were talking about the function that gets passed into the `fork` function?
For that it gets SUPER tricky as that cancel function is optional. If we make it required here things could get nasty really quick.

In that case, we *could* do something like, if the third argument is a function, go ahead and take one more that is the `Async`, otherwise call it done.

Now the alternative to this is to provide (2) functions `fork` and `forkCancel`. The `fork` function has your original definition, and the `forkCancel` takes your new definition. That way we are explicit about it and do not have to deal with the nasty.

What is your thoughts on that?

evilsoft: Now that I am putting a lot of thought into this. As this is a side-effect, what if `fork` acted like `tap`? So instead of returning the cancel function, what if it just returns the Async back.

```js
function fork(rej, res, m) {
  m.fork(rej, res)
  return m
}
```

evilsoft:
I am going to cut a release sans this PR. But as soon as this is merged I will cut a `0.6.1` to add this in. I have a couple peeps waiting on the `dissoc` and `partial` functions. No rush those, I just wanted to let you know that I was doing that.

evilsoft:
closing for now.
May revisit later when we can figure out this darn return value.

PR #124, opened 2017-05-13
add cancellation to Async

evilsoft:
## This is your cancellation notice
image

With the submission of [PR#122](https://github.com/evilsoft/crocks/pull/122) which provides a pointfree `fork` function, it makes sense for `fork` to now return something. Even though the typical use case of the function would be to wrap the `fork` in a `tap` function, there may be times (like at the end of a composition) where the user would want is returned from `fork`.

At the suggestion of @gabejohnson it would make sense for a cancel function to be returned. This function can be called to cancel any pending chained `Async`s and not run them. Also when called, the `reject` and `resolve` functions passed to `fork` will never be called.

In addition to this cancel function being added, the ability to provide `onCancel` functions at the `Async` and `fork` levels. This modifies the signatures of the `Async` constructor and the `fork` function. `Async` now accepts an optional `onCancel` function that will be called when the `Async` it is associated to is cancelled. `fork` also accepts an `onCancel` function that will be called last in the chain of `onCancel` functions.

Looks like this:
```js
const { Async } = require('../crocks')

const fn = x => Async((_, res) => {
  setTimeout(() => {
    console.log(x + 1)
    res(x + 1)
  }, 1000)
})

const withCleanup = x => Async((_, res) => {
  console.log('with Cleanup', x + 1)
  res(x + 1)
}, () => { console.log('Async cleanUp') })

const m = Async.of(0)
  .chain(fn)
  .map(x => x * 2)
  .chain(withCleanup)
  .chain(fn)
  .chain(fn)

const cancel = m.fork(
  console.log.bind(console, 'rejected: '),
  console.log.bind(console, 'resolved: '),
  () => { console.log('fork cleanUp') }
)

// without a call to cancel:
// => 1
// => with Cleanup 3
// => 4
// => 5
// => resolved:  5

setTimeout(cancel, 2000)

// cancel around 2nd chain:
// => 1
// => with Cleanup 3
// => Async cleanUp
// => fork cleanUp
// => 4
```

### DIFF
--- a/crocks/Async.js
+++ b/crocks/Async.js
@@ -16,6 +16,7 @@ const sequence = require('../pointfree/sequence')
 
 const once = require('../helpers/once')
 const mreduceMap = require('../helpers/mreduceMap')
+const unit = require('../helpers/unit')
 
 const All = require('../monoids/All')
 
@@ -79,10 +80,17 @@ function fromPromise(fn) {
   }
 }
 
-function Async(fn) {
+function Async(fn, parentCancel) {
   if(!isFunction(fn)) {
     throw new TypeError('Async: Function required')
   }
+
+  var cancelled
+
+  const cancel = composeB(
+    () => { cancelled = true },
+    isFunction(parentCancel) ? parentCancel : unit
+  )
 
   const type =
     _type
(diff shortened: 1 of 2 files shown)


evilsoft:
@flintinatux
Any opinions or words of wisdom here?

evilsoft:
need to connect up `ap` and `alt` a bit better.

crocks/Async.js
```diff
@@ -93,12 +101,21 @@ function Async(fn) {
...
     constant(`Async${_inspect(fn)}`)
 
-  function fork(reject, resolve) {
+  function fork(reject, resolve, cleanup) {
     if(!isFunction(reject) || !isFunction(resolve)) {
       throw new TypeError('Async.fork: Reject and resolve functions required')
     }
 
-    fn(reject, resolve)
+    const forkCancel = isFunction(cleanup)
```
evilsoft: Should be more like:
```js
const forkCancel =
  isFunction(cleanup) ? cleanup : unit
```

evilsoft:
image

PR #126, opened 2017-05-18
Better compose and pipe functions

evilsoft:
## Making the world a little better...One commit at a time
image

The initial implementation of one of the most commonly used functions made me...well...a little :sob:. We were traversing the list of functions twice, once for validation and another time to build the composition. On top of that, if people use `compose` (the more popular of these functions) we actually went through it (3) times. 

While there is a little duplication, that may turn out to be okay as we trudge onto a mono-repo. So I am okay with it here. (I guess). In addition to these flow changes, I also updated the specs to look for specific errors instead of the `TypeError` blanket for `pipe` and `compose`

### DIFF
--- a/helpers/compose.js
+++ b/helpers/compose.js
@@ -1,25 +1,41 @@
 /** @license ISC License (c) copyright 2016 original and current authors */
 /** @author Ian Hofmann-Hicks (evil) */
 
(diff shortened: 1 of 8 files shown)


evilsoft:
image

PR #128, opened 2017-06-08
Add `extend` to `Pair`

evilsoft:
## Full extension
image

Was playing with `Pair` and was like, "gosh I wish I could take the entire `Pair` into consideration before updating this value" and remembered this blog post:
[Comonads, Monoids and Trees](https://joneshf.github.io/programming/2015/12/31/Comonads-Monoids-and-Trees.html)

and then it clicked, did some diagrams and was like...ohhhhh THAT is what `extend` is for!! So now there is this.

This PR does the following:
* adds `extend` to the `Pair` type
* add the `extend` point-free function
* adds the `isExtend` predicate function
* :wrench: the wording for an error on the point-free `map` function


evilsoft:
image

PR #129, opened 2017-06-11
Add `@@implements`to allow Constructors to signal what they do

evilsoft:
## The Start of a Long Journey
image

As types like `Star` and `Writer` and some soon to be Monad Transformers would like to be fixed with the Constructor of the type, we need a way to communicate the algebras that the instances implement. By adding an `@@implements` function to the type constructors, this is a cheap and simple way to make that so.

This PR:
* Adds a couple more missing predicates:
  * `isBifunctor`
  * `isContravariant`
  * `isPlus`
  * `isProfunctor`
* Implements `@@implements` on all crocks
* Implements `@@implements` on all monoids
* Updates all algebra predicate functions to look for `@@implements` in addition to normal duck typing for instances.
* Add some internal helpers for working with `@@implements`:
  * `hasAlg`
  * `implements`


evilsoft:
image

PR #130, opened 2017-06-26
Add the Last Monoid

evilsoft:
image

Would be nice to have both the `First` and `Last` monoid in crocks. They are handy for all kinds of things.
This PR implements the `Last` monoid, which has Transformations to other Sum types (as it is a Monoid that wraps a Sum Type).

This PR does the following:
* Provide the `Last` monoid
* Provide all Transformation permutations for `Last` to the other valid Sum Types.


evilsoft:
image

PR #131, opened 2017-07-01
Add the First Monoid

evilsoft:
## First Things First
image

The last of the `Maybe` based Monoids going into `crocks` is `First`.
This PR adds the Monoid itself as well as the transformation functions associated with it.
Also noticed a couple documentation issues that made it in with the `Last` monoid, so fixed those up a little bit.


evilsoft:
image

Issue #123, opened 2017-05-10
Need to compile ALL THE THINGS and be a better citizen


evilsoft:
## So Much SMH
![](https://media.tenor.co/images/b79ed09bcac0d4a3e7fe60d5fca3f989/tenor.gif)
So there are a couple issues that can arise from the current build and publish system:

* Code style is not properly executed in an `eval` within a webpack bundle. (well that is where the issue was found, there are other places this breaks down. So if ANY of the eval based devtools are used with Webpack, then users are not going to have a good time especially in Safari
* Also we are being a bit deceptive and not providing our due diligence with how the code is provided. Most people do not babelify their `node_modules`, so if they are expecting this to work in dear ol Safari or an older browser, then they will be super sad.

There are a few other reasons to do this, but these are big enough to call out.
So this will be an API change to the max:
* Going to go single dir in a folder called `src` that will compile to a folder called `lib`.
* `src` will have a flat structure, no `crocks/Async.js` nonsense. just `Async.js`.
* All specs will also be in the same folder, like now.
* Everything in `src` will be compiled into `lib`, specs so testing can be done in the CI environment. This will be tied to `npm test`.
* `npm run spec:dev` will have to doing something different, no idea what that is now, but I do not want to do a full build on every file save. Will probably have to do something silly to get it to run with `tape`, but we will see.
* We will ONLY publish the lib folder(sans specs), `package.json`, `README.md` and the License.
* Update the README to show `import Async from 'crocks/lib/Async'` instead of `'crocks/crocks/Async'`
* Remove the references to the folders like `combinators`, `helpers` and what have you.

evilsoft:
Address this issue with [this PR](https://github.com/evilsoft/crocks/pull/132).

evilsoft:
Released version `0.7.0` that addresses this issue.
Closing.

Issue #116, opened 2017-03-30
Have you considered a monorepo?


bebraw:
Hi,

Just came by your library through a chance. The functionality looks great. I started to wonder, though, if it would make sense to port it to monorepo style. You would end up with `@crocks/combinators` and so on. Note that npm made organizations free for open source projects a while ago and you can find tools like [Lerna](https://lernajs.io/) for managing the scheme.

The added benefit would be that this would allow you to split up the current README into something more focused per package. The main README of the project would give an overview of what the library is about and give your documentation some abstraction. You can also aggregate all this as a neat documentation site quite easily.

I wrote [a little blog post](https://medium.com/@bebraw/the-case-for-monorepos-907c1361708a) if you want to learn more about the approach.

Keep up good work. 👍 

PS. Let me know if you want to do a little interview for the SurviveJS blog. I've been highlighting gems like this there every once in a while. 😄

evilsoft:
Thanks so much for your random stumble.

I am doing more than considering, it is on the roadmap. I plan to get it out after a few major versions if all goes well after 1.x.x.

I expect much churn on the initial API and inner-dependencies (?) and want to get that solid in one shot instead of managing those change across multiple projects. So while it will be a bit bloated until around 4.x (I hope) it will allow us to get there quicker IMO.

I do not mind doing an interview about `crocks` or whateves, but I am a pretty boring bloke, so I do not know if your readers would end up hating you for wasting your time on me. But I am game for just about anything.

bebraw:
> I expect much churn on the initial API and inner-dependencies (?) and want to get that solid in one shot instead of managing those change across multiple projects. So while it will be a bit bloated until around 4.x (I hope) it will allow us to get there quicker IMO.

Yeah, I went from monolith to monorepo in Reactabular. Even though there was churn, I think the move was worth it as it forced me to re-evaluate the design. As a result I ended up with something decoupled. The interesting thing to notice was that the move allowed me to push secondary functionality that wasn't related to the core outside of the monorepo so I ended up with a mixed model.

> I do not mind doing an interview about crocks or whateves, but I am a pretty boring bloke, so I do not know if your readers would end up hating you for wasting your time on me. But I am game for just about anything.

Ok, cool. I sent you mail about this.

Feel free to close the issue unless you want to keep this open till monorepo is done. 👍

evilsoft:
Just stated the first steps into mono repoing this silly project.
[...]

PR #132, opened 2017-07-05
Compile all code for a better experience and flatten API 

evilsoft:
## It all starts NOW
image

This is a **BIG BREAKING CHANGE**!

This mainly addresses [this issue](https://github.com/evilsoft/crocks/issues/123) and compiles all code into ES5 mainly so it can be used in Safari and IE/Edge without having the end user compile our stuff from their project (total bad news bears).

Two other things were started with the change, as it was going to be a breaking change anyway, decided to make the first step to a monorepo to address [this issue](https://github.com/evilsoft/crocks/issues/116). By flattening the API, it makes it easier to determine inner-dependency within the library and isolate it appropriately.

All code that is referenced by other functions/types in the library is now inside of a `core` folder, which will be the start of `@crocksjs/core` which will hold all the base code needed by the other functions. Another benefit of flattening the API is a lightened cognitive load of rememebring if this is a `pointfree` or `helper` function. All functions, regardless of their type can be reached from `crocks/lib/[function_name]`, so no need to keep classification for each function in your head.

Another benefit of the separation of core, is that I removed all type checking (except for the `crocks` and `monoids`) and currying for the core versions of the functions. This greatly improves performance by only doing type checks once, before calling the functions in core. Also no extra currying, it leaves all currying in the public available API. This is the first stage of optimizing the code and improving for other libs that would like to use some of these bits.


evilsoft:
image

PR #133, opened 2017-07-12
Remove name conflict with new structure

evilsoft:
## When OSX makes you sad and disappointed.
image

Well that first attempt at squashing the folder structure created all kinds of issues on OSX due to naming conflicts from the case-insensitve nature of the filesystem used on a standard OSX system. While I am okay with turning a blind eye to windows dev, removing development on crocks on OSX is not an option.

Need to move back to a folder based structure, but this time going to make things DataType based, bringing us one step closer to better mono-repo options. This is the first part to a new series of PRs. * Move all helpers that construct data types into their respective folders.
* Shed some weight in `src/core` but only keeping Pair, Maybe and Unit inside of the folder. This will entail creating type proxies that can be used for type checking in the various Sum Type transformation functions.
* Before the next publish, will need to alter the publish flow to have the generated `build` folder be the root of the npm package. This means as part of the publish flow: `package.json`, `README.md` and `LICENCE` need to be copied into the `build` folder.


evilsoft:
image

PR #134, opened 2017-07-15
Move type based functions into type folders

evilsoft:
## Cozy up. Going to get Comfy!
image

This PR is the second in a series focused on moving onto a more mono-repo friendly structure. This PR takes care of moving the Type-specific functions into their respective Type folders. Shared or non-specific functions stay where they are.


evilsoft:
image

PR #135, opened 2017-07-22
Add publish bits for new api

evilsoft:
## All coming together nicely!
image

This is the penultimate PR for the new API change. This will get us yet another step to a well needed mono-repo. All this does is add a simple script that copies over the non-built assets needed for publish and ONLY publishes the compiled code directory allowing for references like:
```js
const Maybe = require('crocks/Maybe')
const safe = require('crocks/Maybe/safe')

// also can just:
const { Maybe, safe } = require('crocks')
```

The FINAL step before `0.7.0` is to allow for Type Proxies to remove dependencies for simple type checks. This will allow, say, functions that depend on a `Pair` but do not construct a `Pair` exists without needing to bring in `Pair`. Removing unneeded inner-dependencies.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,23 +1,6 @@
-node_modules
-test
-src
 *.spec.js
-
 .npmignore
-.babelrc
-webpack.config.js
 
-CHANGELOG.md
-
-.travis.yml
-.eslintignore
-.eslintrc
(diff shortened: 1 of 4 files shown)


evilsoft:
image

PR #136, opened 2017-07-24
Use TypeProxies to remove simple type check dependencies

evilsoft:
## That should do it!!
image

This is the final PR for the new API. It allows us to move dependencies from core when a function or other type depends on a type reference for type checking. This is done by creating a shared Type name lookup in `core` and using it to create a simple proxy that can be used for type comparisons.
So many type were able to be moved out of `core` leaving just the bare essentials (for now anyway).

The following types were removed from `core` and into their respective folders:
* `Arrow`
* `Async`
* `Either`
* `First`
* `Last`
* `Pred`
* `Result`
* `Star`

In addition to this new Type Proxy, there is another additional breaking change, that removes `gets` from `State` and makes `State.get` behave like `Reader.ask`, taking a function and applying it to the Functor portion on the structure. (this will soon turn into something that defaults to `identity` when nothing is passed, but will map the function if present).


evilsoft:
image

PR #137, opened 2017-07-24
Make `Reader.ask` and `State.get` use identity by default

evilsoft:
## Make things more flexible
image

A common pattern when using `Reader.get` is to do something like `Reader.get(identity)` when one wants to use the environment as is, with no additional mapping. Because of this, it makes sense to allow for a version of the function that will automatically map identity when no arguments are passed.

As `State` provides a similar function with `State.get`, it makes sense to have that function follow suit. So now, with this PR both `Reader.ask` and `State.get` will map identity when nothing is passed to the function. If a function is in fact passed to those functions, then they will behave as normal, mapping the function over the environment or state.

### DIFF
--- a/src/Reader/Reader.spec.js
+++ b/src/Reader/Reader.spec.js
@@ -88,23 +88,25 @@ test('Reader runWith', t => {
 test('Reader ask errors', t => {
   const ask = bindFunc(Reader.ask)
 
-  t.throws(ask(undefined), TypeError, 'throws with undefined')
-  t.throws(ask(null), TypeError, 'throws with null')
-  t.throws(ask(0), TypeError, 'throws with falsey number')
-  t.throws(ask(1), TypeError, 'throws with truthy number')
-  t.throws(ask(''), TypeError, 'throws with falsey string')
-  t.throws(ask('string'), TypeError, 'throws with truthy string')
-  t.throws(ask(false), TypeError, 'throws with false')
-  t.throws(ask(true), TypeError, 'throws with true')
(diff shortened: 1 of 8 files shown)


src/State/State.spec.js
```diff
@@ -137,7 +137,7 @@ test('State evalWith', t => {
   t.end()
 })
 
-test('State get', t => {
+test('State get errors', t => {
   const get = bindFunc(State.get)
 
   const err = /State.get: Function Required/
```
evilsoft: Need to change the wording on this error to include language about no arguments.

src/Reader/Reader.spec.js
```diff
@@ -88,23 +88,25 @@ test('Reader runWith', t => {
...
-  t.throws(ask(null), TypeError, 'throws with null')
-  t.throws(ask(0), TypeError, 'throws with falsey number')
-  t.throws(ask(1), TypeError, 'throws with truthy number')
-  t.throws(ask(''), TypeError, 'throws with falsey string')
-  t.throws(ask('string'), TypeError, 'throws with truthy string')
-  t.throws(ask(false), TypeError, 'throws with false')
-  t.throws(ask(true), TypeError, 'throws with true')
-  t.throws(ask([]), TypeError, 'throws with an array')
-  t.throws(ask({}), TypeError, 'throws with an object')
+  const err = /Reader.ask: Function required/
```
evilsoft: Need to change the wording on this error to include language about no arguments.

evilsoft:
image

PR #138, opened 2017-07-26
Make `Star` a `Category` by fixing it to a given `Monad`

evilsoft:
## A new Category of Star
image

This is a breaking change on `Star` that now requires a `Monad` constructor to get back the actual constructor to use.
```js
const MaybeStar = Star(Maybe)

// isValid : a -> Maybe a
const isValid = 
  safe(isNumber)

const m = 
  MaybeStar(isValid)
    .map(add(10))
    .map(multiply(2))

m.runWith(5)   // Just 30
m.runWith('a') // Nothing

// now with `id`
const left =
  MaybeStar.id().compose(m)

const right =
  m.compose(MaybeStar.id())

left.runWith(10)  // Just 40
right.runWith(10) // Just 40

```

Doing this allows for implementation of `id` making this `Star` a `Category`.

### DIFF
--- a/src/Star/Star.spec.js
+++ b/src/Star/Star.spec.js
@@ -10,32 +10,65 @@ const constant = require('../core/constant')
 const identity = require('../core/identity')
 const isFunction = require('../core/isFunction')
 const isObject = require('../core/isObject')
+const isSameType = require('../core/isSameType')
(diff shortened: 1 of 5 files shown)


evilsoft:
image

PR #139, opened 2017-07-27
Update README and some :lipstick:

evilsoft:
## 💄 :allthethings:
image

This is the documentation PR to go along with all the organization changes from the last stretch of PRs. In addition to some simple documentation, also needed to move `fromPairs` out of the `Pair` namespace and into `helpers`. It does not construct a `Pair` and as such does not need to 🚢 with it.


evilsoft:
image

PR #140, opened 2017-07-28
Accumulate `Err` on `Alt` for `Result` type

evilsoft:
## Best Laid Plans
image

After messing about with `Alt` and trying to come up with helpers for `Alternative` and `Alt`, I was really sad that `Result` did not accumulate `Err` like it does with `ap`. It would be very useful and does not ***seem*** to break any laws.

So that is what this is. Now if you use a `Semigroup` for your `Err` and use `alt` it will concat your Errs. If you do not use a `Semigroup`, then `alt` behaves in a manner that works like `Either.alt`, in which it return the first `Ok`, but the last `Err`. Like so:

```js
const withSemi =
  Err([ 3 ])
    .alt(Err([ 13 ]))
    .alt(Err([ 42 ]))

const noSemi =
  Err(3)
    .alt(Err(13))
    .alt(Err(42))

withSemi
// > Err [ 3, 13, 42 ]

withSemi
  .alt(Ok('yep'))
// > Ok 'yep'

noSemi
// > Err 42

noSemi
  .alt(Ok('yep'))
// > Ok 'yep'
```

### DIFF
--- a/src/Result/Result.spec.js
+++ b/src/Result/Result.spec.js
@@ -522,17 +522,32 @@ test('Result alt errors', t => {
   t.end()
 })
 
-test('Result alt functionality', t => {
+test('Result alt functionality with Semigroup Err', t => {
   const right = Result.of('Ok')
   const anotherOk = Result.of('Another Ok')
 
-  const left = Result.Err('Err')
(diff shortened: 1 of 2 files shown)


evilsoft:
image

PR #141, opened 2017-08-01
Clean up  `All`, `Any`, `Arrow` and `Pair`

evilsoft:
## They can't all be zingers
image

As the various READMEs and Documentation is being add, we will be taking the opertunity to do some things that needed to be done to each of the Types in `crocks`:

1. Make sure all specs around errors are looking for the exact error message instead of simply looking for `TypeError`s.
1. Remove all dependency on internal functions like `compose`, `constant`, `identity`, etc. It is cool for things like `implements` and `inspect`. Mostly want to remove dependencies on the functions that are available in the public API. (This helps keep the noise down on the stacks when we throw, as well as makes it easier to ship with a small core when we get all monorepo)
1. Remove any derived functions (like `map` and `ap` derived from `chain`).
1. Consolidate type checks in the algebras when possible.

This is the first wave, will be doing these for every ADT in the lib. These PRs will be driven by the documentation of the individual namespaces.


evilsoft:
image

PR #142, opened 2017-08-27
Add Some Pair Transforms

evilsoft:
## Some Flair for the Pair
image

This PR adds a couple ways to transform around the `Pair` type:
* adds `writerToPair` as a standard transformation function
* adds a `toArray` method to the `Pair` instance, that will...well...return an `Array` from a given `Pair`.

No big whoop!


evilsoft:
image

PR #143, opened 2017-08-27
Add `mapReduce` helper function

evilsoft:
## Straight off the press (well the gitter anyhow)

image

A conversation on the gitter with @amsross showed how useful a function like `mapReduce` can be. This PR just implements it.


evilsoft:
image

amsross:
I wouldn't guess this warrants its own release, but is there a general release timeline in which merged PRs can be expected to become available on npm?

evilsoft:
I was going to try to get the new `ReaderT`, but I can totally cut `0.7.1` tonight.
Not a problem!
Release Early, Release Often.

amsross:
Awesome stuff, btw. I was able to use `mapReduce` immediately. I thought there would be a closer-to-the-monoid solution, but once I pulled it in and used it, it became apparent that this was about as simple as one could hope for!

Issue #144, opened 2017-09-02
Maybe/prop throws when passed `undefined`


evilsoft:
This should not happen, a Nothing should be returned if a non-object is passed to `prop`.
:smh:

evilsoft:
This was fixed and merged in. Will go out in the next release cycle.
Closing.

PR #145, opened 2017-09-04
Allow `prop` and `propPath` to accept `null` and `undefined` as data

evilsoft:
## All the bases covered
image

As reported in [this issue](https://github.com/evilsoft/crocks/issues/144) there was a problem when `null` or `undefined` is passed as the data to `prop` and `propPath`. As both of these functions take any `a` as data, they should not throw when passed terminal types. Instead they should map to `Nothing`.

This PR does just that. Also because safe was not really used anywhere except for those (and `safeLift`, but that is a story for another time), moved it out of core. Will deal with `safeLift` with another idea I have surrounding that function.


evilsoft:
image

PR #147, opened 2017-10-01
Move to ES6 Rest operator instead of slicing arguments.

evilsoft:
## We can do it, we have the tech

Now that we are doing the whole `buble` thing, makes sense to use the built in Rest operator instead of casting to an array. In order to accomplish this, we had to move some of the test mocks into the src folder to make sure things pass on versions 4 and 5 of node. Due to the mocks being moved, the build, publish, linting and coverage flows had to be updated.

Also removed the `constant` and `identity` functions from the core folder and substituted with their implementations when needed. Mostly these occurred in the specs, so it did not make sense to keep such simple functions in `core`.


evilsoft:
image

PR #148, opened 2017-11-04
Add READMEs for All, Any and Arrow types

evilsoft:
## Seems like an OKAY start
image

These are the first in a series of Readmes to be added. This is another step in the [Quest for monorepo](https://github.com/evilsoft/crocks/issues/116).

Both `All` and `Any` did not require any API changes, while there was no need to keep `value` for `Arrow` as it no longer is used as a "traditional" `Monoid` as it was before `Category` was a thing.

This is a BREAKING API change. No depreciation, just dropping it.

### DIFF
--- a/src/All/README.md
+++ b/src/All/README.md
@@ -1,0 +1,100 @@
+ # All
+```haskell
+All Boolean
+```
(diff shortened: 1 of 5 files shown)


evilsoft:
image

Actually I am okay with these examples. They are not so bad.
Merging.

Issue #8, opened 2016-11-20
Equality should be by value and not by reference


evilsoft:
Implement equality by reference by:
- [x] Adding a `pointfree/equals` that will compare values inside of `Object`, `Arrays`, etc.
- [x] Update all `Setioids` to use the value equality when comparing their values
- [x] Add a `isSame` predicate function, for a cheap way to check equality by reference,

Note: Will implement equality for all Javascript Types, excluding `Map`, `WeakMap`, `Set` and `WeakSet`. Those type are commonly used for referential indexing and can be expensive to generally compare. Will leave it up to the user of these type to define their own equality that is optimized for their usage of those types.

evilsoft:
Merged [#149](https://github.com/evilsoft/crocks/pull/149).
Closing.

PR #149, opened 2017-11-06
Equality is by value

evilsoft:
## ... and Justice for All
image

This PR addresses [this issue](https://github.com/evilsoft/crocks/issues/8) and finally adds equality by value throughout the library. It does the following:

* Adds `predicates/isSame` for referetial based equality (basically `Object.is`)
* Adds `pointfree/equals` for value based equality and `Setoid` compares
* Updates all `Setoid`s to use value based equality by default.


evilsoft:
image

PR #150, opened 2017-11-06
Return `Pair` for `Writer`'s `read` method

evilsoft:
## A bit more flexible
image

I really hate the lack of usability with the `Writer`'s `read` function. I also not like how it removed the Monoid from the `log` value. This PR now returns a `Pair` for `w.read()` with the Monoid wrapped `log` in the `fst` and the `value` in the `snd`.

With this change we now get the power of a `Pair` when we are done with the writing. Also it makes more sense as `Pair` is the canonical Product type.

This is a BREAKING change, will not depreciate, just changing straight away.


evilsoft:
image

PR #151, opened 2017-11-06
Use `valueOf` instead of `value`

evilsoft:
## A bit more Javascript-y
image

We can get a better pointfree function that can work on most types if we go with `valueOf` over `value`. That is what this PR addresses. It moves all things that provided a `value` function to now provide a `valueOf` function.


evilsoft:
image

PR #152, opened 2017-11-10
Better `runWith` errors for State

evilsoft:
## Just a spoonful of Sugar
image

On `State` the `runWith` method did not throw if one did not return a `Pair`. This PR now does that and uses this new error to clean up some more of the error handling for all the methods.
Thinking this is more a Bug Fix then a Breaking change...But we will see hows I Feelz when the release actually goes out.


evilsoft:
image

PR #153, opened 2017-11-11
Add a new `mapProps` helper function

evilsoft:
## Mad Props you say?
image

Was really in need of a function that worked similarly to `Ramda`'s `evolve` function. So I added one, but with a twist. I have found that were where more times than not that I wanted the ability to nest `evolve` instead of evolving inside of other `evolve`s.

Other than nesting, another difference is this version with throw a TypeError, if you do not provide a function in the mapping `Object`. Boy oh boy the hours spent trying to find where a function was not available.

**Bonus**
Cleaned up the README, to make diffs a let easier to follow. Also fixed up the type siggys to be a bit more Haskell-y.


evilsoft:
image

Issue #157, opened 2017-12-16
Add propOr on Maybe


HenriqueLimas:
IMHO should be nice to have a `propOr` function to use in addition to `prop` on the `Maybe` module, where we can pass a default value when the prop is not there. Maybe composing 'default' and 'prop' together can be an idea. I can create a PR if you want.

evilsoft:
`propOr` and `propPathOr` are not specifically in there because I assumed that people would want to keep things in the `Maybe` for further expansion. For example say I want a prop that needs to be a number, I can then do things like:
```js
const chain = require('crocks/pointfree/chain')
const compose = require('crocks/helpers/compose')
const isNumber = require('crocks/predicates/isNumber')
const prop = require('crocks/Maybe/prop')
const safe = require('crocks/Maybe/safe')

const validNumber = key =>
  compose(chain(safe(isNumber)), prop(key))
```

So it made sense that if a user would want to "stay" in the vanilla JS they would either bring it in from a lib like ramda or tiny-func or something that stays in JS. If you think there is value in adding this to `crocks` then, heck yeah, submit a PR.

But I do not think Maybe is the place for it as it does not return a Maybe all the time. It will have a signature like `propOr :: a -> String -> b -> c` 
```js
// propOr :: a -> String -> b -> c
function propOr(def, key, data) {
  ...
}
```

So I am think this should live inside of `helpers` as it really has nothing to do with `Maybe`. When implementing, try to use the bare minimum of `crocks` functions as this will bloat stack traces on errors, and only use function inside of the `core` folder. Will keep some :eyes: out for a PR.

And if you want, might as well make `propPathOr`.

evilsoft:
PR is merged.
Looks like this feller should be closed.

PR #158, opened 2017-12-17
Add `propOr` and `propPathOr` and Documentation for both

HenriqueLimas:
PR for #157 

Looking at the code, and noticed that `Maybe` has the `option` method, so the `propOr` should not need, but yeah, here is the PR :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -930,6 +930,32 @@ some data, and it will attempt to resolve your provided path. If the path is
 valid, it will return the value residing there (`null` included!) in a `Just`.
 But if at any point that path "breaks" it will give you back a `Nothing`.
 
+#### propOr
+`crocks/helpers/propOr`
+```haskell
+propOr :: a -> (String | Integer) -> b -> c
+```
+If you want some safety around pulling a value out of an Object or Array with a
+single key or index, you can always reach for `propOr`. Well, as long as you are
+working with non-nested data that is. Just tell `propOr` either the key or index
+you are interested in, and you will get back a function that will take anything
+and return the wrapped value if the key/index exists. If the
+key/index does not exist however, you will get back a default value.
+
+#### propPathOr
+`crocks/helpers/propPathOr`
+```haskell
+propPathOr :: a -> [ String | Integer ] -> b -> c
+```
+While [`propOr`](#propor) is good for simple, single-level structures, there may
+come a time when you have to work with nested POJOs or Arrays. When you run into
+this situation, just pull in `propPathOr` and pass it a left-to-right traversal
+path of keys, indices or a combination of both (gross...but possible). This will
+kick you back a function that behaves just like [`propOr`](#propor). You pass it
+some data, and it will attempt to resolve your provided path. If the path is
+valid, it will return the value.
+But if at any point that path "breaks" it will give you back the default value.
+
 #### safe
 `crocks/Maybe/safe`
 ```haskell
--- a/src/helpers/propOr.js
+++ b/src/helpers/propOr.js
@@ -1,0 +1,25 @@
+/** @license ISC License (c) copyright 2017 original and current authors */
+/** @author Henrique Limas */
+
+const curry = require('../core/curry')
+const isNil= require('../core/isNil')
+const isInteger = require('../core/isInteger')
+const isString = require('../core/isString')
+
+const lift = (x, def) =>
+  isNil(x) ? def : x
+
+// propOr : a -> String | Integer -> b -> c
+function propOr(def, key, target) {
+  if(!(isString(key) || isInteger(key))) {
+    throw new TypeError('propOr: String or integer required for second argument')
+  }
+
+  if(isNil(target)) {
+    return def
+  }
+
+  return lift(target[key], def)
+}
+
+module.exports = curry(propOr)
--- a/src/helpers/propOr.spec.js
+++ b/src/helpers/propOr.spec.js
@@ -1,0 +1,52 @@
+const test = require('tape')
+const helpers = require('../test/helpers')
+
+const bindFunc = helpers.bindFunc
+
+const isFunction = require('../core/isFunction')
+
+const propOr = require('./propOr')
+
+test('propOr function', t => {
+  const p = bindFunc(propOr)
+  const def = 'default value'
+
+  t.ok(isFunction(propOr), 'is a function')
+
+  const err = /propOr: String or integer required for second argument/
+  t.throws(p(def, undefined, {}), err, 'throws with undefined in second argument')
+  t.throws(p(def, null, {}), err, 'throws with null in second argument')
+  t.throws(p(def, false, {}), err, 'throws with false in second argument')
+  t.throws(p(def, true, {}), err, 'throws with true in second argument')
+  t.throws(p(def, 1.5, {}), err, 'throws with float in second argument')
+  t.throws(p(def, [], {}), err, 'throws with an array in second argument')
+  t.throws(p(def, {}, {}), err, 'throws with an object in second argument')
+
+  const value = 'Bobby Joe'
+  const obj = { a: value, bad: null }
+
+  const objGood = propOr(def)('a')(obj)
+  const objBad = propOr(def)('b', obj)
+  const objNull = propOr(def)('bad')
+
+  t.equals(objGood, value, 'returns the value when key is found')
+  t.equals(objBad, def, 'returns default value when key is not found')
+  t.equals(objNull(obj), def, 'returns default value when key is found and value is null')
+
+  const arr = [ value, null ]
+
+  const arrGood = propOr(def, 0, arr)
+  const arrBad = propOr(def, 5)(arr)
+  const arrNull = propOr(def, 1)
+
+  t.equals(arrGood, value, 'returns the value when index is found')
+  t.equals(arrBad, def, 'returns default value when index is not found')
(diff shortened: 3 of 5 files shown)


src/helpers/propOr.js
```diff
@@ -0,0 +1,25 @@
+/** @license ISC License (c) copyright 2017 original and current authors */
+/** @author Ian Hofmann-Hicks (evil) */
```
evilsoft: You own this :sire: you should put your name here.

src/helpers/propPathOr.js
```diff
@@ -0,0 +1,32 @@
+/** @license ISC License (c) copyright 2017 original and current authors */
+/** @author Ian Hofmann-Hicks (evil) */
```
evilsoft: You are the owner of this. You can replace my name with yours.

evilsoft:
:eyes:

README.md
```diff
@@ -930,6 +930,32 @@ some data, and it will attempt to resolve your provided path. If the path is
...
+`crocks/helpers/propOr`
+```haskell
+propOr :: a -> (String | Number) -> b -> c
+```
+If you want some safety around pulling a value out of an Object or Array with a
+single key or index, you can always reach for `propOr`. Well, as long as you are
+working with non-nested data that is. Just tell `propOr` either the key or index
+you are interested in, and you will get back a function that will take anything
+and return the wrapped value if the key/index exists. If the
+key/index does not exist however, you will get back a default value.
```
evilsoft: Not needed, but if you want you can add a brief example on these.
Only if you want to, this is not a requirement.

evilsoft: I REALLY :heart: how you match the overall voice of the documents!
Good show m8!!

README.md
```diff
@@ -930,6 +930,32 @@ some data, and it will attempt to resolve your provided path. If the path is
...
+#### propPathOr
+`crocks/helpers/propPathOr`
+```haskell
+propPathOr :: a -> [ String | Number ] -> b -> c
+```
+While [`propOr`](#propOr) is good for simple, single-level structures, there may
+come a time when you have to work with nested POJOs or Arrays. When you run into
+this situation, just pull in `propPathOr` and pass it a left-to-right traversal
+path of keys, indices or a combination of both (gross...but possible). This will
+kick you back a function that behaves just like [`propOr`](#propOr). You pass it
```
evilsoft: I believe that the hash ref needs to be all lowercase.
One way I check my refs is by clicking view and then checking if the clickys go where they are supposed to.

src/helpers/propOr.spec.js
```diff
@@ -0,0 +1,53 @@
...
+  const def = 'default value'
+
+  t.ok(isFunction(propOr), 'is a function')
+
+  const err = /propOr: String or integer required for second argument/
+  t.throws(p(def, undefined, {}), err, 'throws with undefined in second argument')
+  t.throws(p(def, null, {}), err, 'throws with null in second argument')
+  t.throws(p(def, false, {}), err, 'throws with false in second argument')
+  t.throws(p(def, true, {}), err, 'throws with true in second argument')
+  t.throws(p(def, 1.5, {}), err, 'throws with float in second argument')
```
evilsoft: may want to throw in a falsey zero as well just to be safe.

src/helpers/propOr.js
```diff
@@ -0,0 +1,25 @@
...
+function propOr(def, key, target) {
+  if(!(isString(key) || isInteger(key))) {
+    throw new TypeError('propOr: String or integer required for second argument')
+  }
+
+  if(isNil(target)) {
+    return def
+  }
+
+  return lift(target[key], def)
```
evilsoft: I am thinking that we should let null through if the value is set to null. I am okay with guarding for `unit` (`undefined`), but null could be a valid value. What are your thoughts?

src/helpers/propOr.spec.js
```diff
@@ -0,0 +1,53 @@
...
+  const arrGood = propOr(def, 0, arr)
+  const arrBad = propOr(def, 5)(arr)
+  const arrNull = propOr(def, 1)
+
+  t.equals(arrGood, value, 'returns the value when index is found')
+  t.equals(arrBad, def, 'returns default value when index is not found')
+  t.equals(arrNull(arr), def, 'returns default value when index is found and value is null')
+
+  const fn =
+    x => propOr(def, 'key', x)
```
evilsoft: this *could* just be:
```js
const fn =
  propOr(def, key)
```

src/helpers/propPathOr.js
```diff
@@ -0,0 +1,32 @@
...
+  if(isNil(target)) {
+    return def
+  }
+
+  return keys.reduce((target, key) => {
+    if(!(isString(key) || isInteger(key))) {
+      throw new TypeError('propPathOr: Array of strings or integers required for second argument')
+    }
+
+    return lift(target[key], def)
```
evilsoft: Same comment here about the letting `null` through.

src/helpers/propOr.spec.js
evilsoft: NOPE!! I am wrong. I should take care to look at signatures!
Sorry about that, this is perfect as is.

README.md
```diff
@@ -930,6 +930,32 @@ some data, and it will attempt to resolve your provided path. If the path is
 valid, it will return the value residing there (`null` included!) in a `Just`.
 But if at any point that path "breaks" it will give you back a `Nothing`.
 
+#### propOr
+`crocks/helpers/propOr`
+```haskell
+propOr :: a -> (String | Number) -> b -> c
```
evilsoft: this should be `propOr :: a -> (String | Integer) -> b -> c`

README.md
```diff
@@ -930,6 +930,32 @@ some data, and it will attempt to resolve your provided path. If the path is
...
+single key or index, you can always reach for `propOr`. Well, as long as you are
+working with non-nested data that is. Just tell `propOr` either the key or index
+you are interested in, and you will get back a function that will take anything
+and return the wrapped value if the key/index exists. If the
+key/index does not exist however, you will get back a default value.
+
+#### propPathOr
+`crocks/helpers/propPathOr`
+```haskell
+propPathOr :: a -> [ String | Number ] -> b -> c
```
evilsoft: This should be `propPathOr :: a -> [ String | Integer ] -> b -> c

src/helpers/propOr.js
```diff
@@ -0,0 +1,25 @@
...
+
+const curry = require('../core/curry')
+const isNil= require('../core/isNil')
+const isInteger = require('../core/isInteger')
+const isString = require('../core/isString')
+
+const lift = (x, def) =>
+  isNil(x) ? def : x
+
+// propOr : a -> String | Number -> b -> c
```
evilsoft: should be `// propOr : a -> String | Integer -> b -> c`

src/helpers/propPathOr.js
```diff
@@ -0,0 +1,32 @@
...
+const curry = require('../core/curry')
+const isArray = require('../core/isArray')
+const isInteger = require('../core/isInteger')
+const isNil= require('../core/isNil')
+const isString = require('../core/isString')
+
+const lift = (x, def) =>
+  isNil(x) ? def : x
+
+// propPathOr : a -> [ String | Number ] -> b -> c
```
evilsoft: this should be `// propPathOr : a -> [ String | Integer ] -> b -> c`

evilsoft:
So sorry to waffle on this, but I think you are :100: correct to default on `null` values.
mostly because of `propPathOr`, I am going to update `prop` and `propPath` to also go to a `Nothing` on `null` and `NaN`. (this curiosity led me to find a :bug: with `propPath`.

src/helpers/propOr.js
HenriqueLimas: Yeah, I was thinking about that. I thought to consider `null` as not a valid one to not get possible errors. Maybe this should be a problem on `propPathOr` if one of the father props is null. But on `propOr` I will consider null as a value, make more sense. But on the `pathPropOr` what you think is better? JS Throwing the error or checking if the target is null and return it?

README.md
evilsoft: Oh you just copied the original copy for those Maybe methods.
But either way awesome!!
:smile_cat:

HenriqueLimas: Yeah, was to say that. I am not so good on documentation 😅. But I can try change something if you think it is not a good one for `propOr` and `propPathOr`

src/helpers/propOr.js
HenriqueLimas: Yep, I thought was the same. It needs to change on `prop` and `propPath` too, I take from there 😅

src/helpers/propOr.js
evilsoft: I think your implementation is correct. And the other implementation are not.
I Updated prop and propPath to match what you are doing here.

README.md
evilsoft: I will fix the original typo on the other functions.

HenriqueLimas:
Any problem if I add `*.swp` (vim files) files on `.gitignore`? I can do it in another PR if you want

evilsoft:
Looks amazing!!
Will be in the next release. Until then, please have a squirrel on me:
image

evilsoft:
I will add the index bits sorry about that!!

HenriqueLimas:
@evilsoft, Ouch, sorry about that, I added right now and the examples too. Thank you to accept the PR 😄

PR #165, opened 2017-12-20
Improve documentation layout using electric.js

HenriqueLimas:
Live demo currently on https://henriquelimas.github.io/crocks


evilsoft:
We can totally work with this!
❤️ the heck out of this. Thanks again for your valued time and effort.
And...

Have a squirrel:
image

Issue #162, opened 2017-12-19
Catch up with Fantasy-Land


evilsoft:
## Well Mostly

Been getting a little lazy with matching the spec, everything should match except for `ap` and `traverse`. We need to do the following:

- [x] Add `constructor` to instances pointing to the `TypeRep`
- [x] Add FL namespaces for everything but `ap`, `traverse` and eventually `ChainRec`
- [x] Use the Sanctuary JS `@@type` format for type reporting.

evilsoft:
Got all caught up on Fantasy Land.
Will go out in version `0.8.5`
Closing.

PR #168, opened 2017-12-23
add constructor to instances to comply

evilsoft:
## Kompliance is Key
image

This is the first non-breaking step in [this issue](https://github.com/evilsoft/crocks/issues/162) bringing us started on the path to fantasy-land compliance. As these are very large, sweeping changes, each step will be implemented in its own PR.

This adds a reference to the `TypeRep` on the `constructor` property of the ADT instances.


evilsoft:
image

PR #155, opened 2017-11-18
Add Assign, Prod, Sum docs

rstegg:

PR #170, opened 2017-12-24
Add linting for the READMEs

evilsoft:
## Lint Madness
image

It was brought up on [this PR](https://github.com/evilsoft/crocks/pull/155) that linting in the READMEs would be beneficial. So here it is, with some bit of caveat:

The eslint plugin we are using has not released the `skip` feature to npm yet. We will need that for the main README. For now, until [this issue](https://github.com/eslint/eslint-plugin-markdown/issues/86) is addressed and `beta.7` is release into the wild.


evilsoft:
@rstegg This should get the linting in the READMEs

evilsoft:
Seems legit.
image

PR #174, opened 2018-01-02
Add Github style header links

evilsoft:
## Nobody should watch the sausage being made
image

This PR adds slugging and link/anchor generation for link headers in markdown.
Also updated and started the link process that we could only dream about on normal github READMEs.


evilsoft:
image

PR #175, opened 2018-01-03
Index and Cross Reference Function Documentation

evilsoft:
## Give'n ❤️ to dem Funcs
image

The functions were a bit difficult to navigate in the docs. This PR lays the ground work for how functions and other bits should be xref. It is very (e)manual (lewis), but it is a start. Once we get the pattern down, we can see about using something like `markdown-toc` or use some build state to track links. But it all starts with the horrible, tedious manual process.


evilsoft:
image

PR #176, opened 2018-01-03
Correct casing for Point-free links in Docs

evilsoft:
## OSX does it again
image

There were some casing issues that were missed in the original implementation of the pointfree links. Thinking about moving the docs system into Docker if this becomes a problem for those times we do docs work in a case-preserving, case-insensitive filesystem like HFS.


evilsoft:
image

PR #182, opened 2018-01-13
Better Error Specs for some of the ADTs

evilsoft:
## Things Can be Better
image

When implementing some other bits, noticed that we were looking for blanket `TypeError`s for the specs for the following types:

* `Either`
* `Identity`
* `Maybe`


evilsoft:
image

PR #183, opened 2018-01-14
Writing changes to docs index

rstegg:
Short PR

- Changes to intent.
- Fixes Grammar & Spelling issues
- (tries to) set 🌽 sistency with playful tone through rest of docs

### DIFF
--- a/docs/src/pages/index.soy
+++ b/docs/src/pages/index.soy
@@ -48,7 +48,7 @@ description: "A collection of well known Algebraic Data Types for your utter enj
       <div class="row">
         <div class="col-md-12 col-md-offset-2">
           <h3 class="about-title">Why?</h3>
-          <p class="about-description"> Functional Programming matters and Crocks makes that world simplier by both curating and providing not only a common interface between each type (where possible of course), but also all of the helper functions needed to hit the ground running.</p>
+          <p class="about-description">Because Functional Programming! Crocks curates and provides a common interface between each data type (where possible of course), not forgetting all of the helper functions necessary to hit the ground running.</p>
         </div>
       </div>
     </div>
@@ -108,7 +108,7 @@ description: "A collection of well known Algebraic Data Types for your utter enj
       <section class="highlight row">
         <div class="col-md-6 col-md-push-9">
           <h4 class="highlight-title">Use a collection of popular Algebraic Data Types (ADTs)</h4>
-          <p class="highlight-description">You have heard of things like Maybe and Either and heck maybe even IO, that is what Crocks provide.</p>
+          <p class="highlight-description">Have you ever heard of Maybe, Either, or, heck, even IO? These are just a few of the ADTs which Crocks brings to the table.</p>
         </div>
         <div class="col-md-5 col-md-offset-2 col-md-pull-6">
           <div class="highlight-graphic">


docs/src/pages/index.soy
```diff
@@ -1,6 +1,6 @@
 ---
 title: "Home"
-description: "A collection of well known Algebraic Data Types for your utter enjoyment."
```
evilsoft: Why this change? This is the tag line for the lib.

docs/src/pages/index.soy
```diff
@@ -48,7 +48,7 @@ description: "A collection of well known Algebraic Data Types for your utter enj
       <div class="row">
         <div class="col-md-12 col-md-offset-2">
           <h3 class="about-title">Why?</h3>
-          <p class="about-description"> Functional Programming matters and Crocks makes that world simplier by both curating and providing not only a common interface between each type (where possible of course), but also all of the helper functions needed to hit the ground running.</p>
+          <p class="about-description">Because Functional Programming! Crocks adds more tools to your belt by curating a common interface between each data type (where possible of course), and including all of the helper functions needed to hit the ground running.</p>
```
evilsoft: Do not know about this change either, crocks curates the types not the interface, that is defined by the Fantasy Land Spec. The purpose of crocks is so a user of the lib does not have to maintain their own implementation of common types.

So it does not add tools to their belt, it keeps them from implementing those "tools"

docs/src/pages/index.soy
```diff
@@ -108,7 +108,7 @@ description: "A collection of well known Algebraic Data Types for your utter enj
       <section class="highlight row">
         <div class="col-md-6 col-md-push-9">
           <h4 class="highlight-title">Use a collection of popular Algebraic Data Types (ADTs)</h4>
-          <p class="highlight-description">You have heard of things like Maybe and Either and heck maybe even IO, that is what Crocks provide.</p>
+          <p class="highlight-description">Have you ever heard of a Maybe? What about an Either? Or even an IO? These are a few ADTs which Crocks brings to the table.</p>
```
evilsoft: Maybe, Either and all of them are types, when an article is added to them, it feels like it refers to an instance of that type. Think of types as sets and instances as elements in that set.

I like the "bring to the table" tho.

docs/src/pages/index.soy
```diff
@@ -118,8 +118,8 @@ description: "A collection of well known Algebraic Data Types for your utter enj
       </section>
       <section class="highlight row">
         <div class="col-md-6 col-md-offset-2">
-          <h4 class="highlight-title">Combine a list of values down to one value</h4>
-          <p class="highlight-description">Crocks provides a large variety of Monoids, which are ADTs that represent a binary operation and are usually locked down it to a specific type. These are great when you need to combine two things into one under a specific operation.</p>
+          <h4 class="highlight-title">Combine values like never before</h4>
```
evilsoft: These bullet points are going to change to highlight why to use crocks over other solutions, like:
* No external dependencies 
* Common interface between all the types
* Everything in one place.
* Pointfree and Fluent style friendly.

So no need to change these as they will change to provide reasons like above in the near fututre.

docs/src/pages/index.soy
rstegg: I'll revert this change.  It looked nice on paper

docs/src/pages/index.soy
rstegg: > crocks curates the types not the interface, that is defined by the Fantasy Land Spec

You wrote this, I edited to make it more readable

> The purpose of crocks is so a user of the lib does not have to maintain their own implementation of common types

That's a given for any lib... what are we doing here?

docs/src/pages/index.soy
```diff
@@ -108,7 +108,7 @@ description: "A collection of well known Algebraic Data Types for your utter enj
       <section class="highlight row">
         <div class="col-md-6 col-md-push-9">
           <h4 class="highlight-title">Use a collection of popular Algebraic Data Types (ADTs)</h4>
-          <p class="highlight-description">You have heard of things like Maybe and Either and heck maybe even IO, that is what Crocks provide.</p>
+          <p class="highlight-description">Have you ever heard of a Maybe, an Either, or, heck, even an IO? These are a few ADTs which Crocks brings to the table.</p>
```
evilsoft: Still think the articles need to be removed.
We should talk about the Types and not instances of.

rstegg: We could be talking about the operation?

docs/src/pages/index.soy
```diff
@@ -108,7 +108,7 @@ description: "A collection of well known Algebraic Data Types for your utter enj
       <section class="highlight row">
         <div class="col-md-6 col-md-push-9">
           <h4 class="highlight-title">Use a collection of popular Algebraic Data Types (ADTs)</h4>
-          <p class="highlight-description">You have heard of things like Maybe and Either and heck maybe even IO, that is what Crocks provide.</p>
+          <p class="highlight-description">Have you ever heard of Maybe, Either, or, heck, even IO? These are a few ADTs which Crocks brings to the table.</p>
```
evilsoft: I like this direction and qualms about it reading like:

"These are just a few of the ADTs which Crocks brings to the table."

rstegg: Sorry I didn't get notified 8 hours ago!  I wish there was a in-browser editor for times like these.

evilsoft:
looks great!!
Have a squirrel.
image

Issue #181, opened 2018-01-12
Allow for `Array` as the `Applicative` in all `Traversable` types


evilsoft:
## Some people's Applicatives

So we allow for the use of `Array` in the pointfree `traverse` functions, now we just need the Traversable types also respect the Applicative aspect of `Array`. The following types need to be updated to work with `Array`:

- [x] Array
- [x] Either
- [x] Identity
- [x] List
- [x] Maybe
- [x] Result

evilsoft:
Submitted [this PR](https://github.com/evilsoft/crocks/pull/184) to address this issue

evilsoft:
Merged [this PR](https://github.com/evilsoft/crocks/pull/184), closing

PR #184, opened 2018-01-14
Accept Array as the Applicative for Traversable Types

evilsoft:
## Traverse the path less traveled
image

Noticed that we were not accepting `Array` for traversable types while it was a valid Applicative. This PR addresses [this issue](https://github.com/evilsoft/crocks/issues/181) and allows for `Array`s as the Applicative for the following types:
* `Array`
* `Either`
* `Identity`
* `List`
* `Maybe`
* `Result`

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 
-rm -rf node_modules && npm i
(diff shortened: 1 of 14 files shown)


evilsoft:
image

PR #185, opened 2018-01-15
Add `Maybe` Documentation

evilsoft:
## This is good...Maybe?
image

This PR adds the README documentation for the `Maybe` type and all of it's support functions. Also updated some issues with some of the siggys.

Once this is merged, there will follow another PR to move this into the documents.

This goes against [this issue](https://github.com/evilsoft/crocks/issues/42)


src/Maybe/README.md
```diff
@@ -0,0 +1,1321 @@
...
+//=> Just 32
+
+Just('hello')
+  .map(joinWith(' -- '))
+  .ap(Just('friend'))
+//=> Just "hello -- friend"
+
+Maybe.of(add)
+  .ap(Just(29))
+  .ap(Nothing())
```
foxdonut: `//=> Nothing`

src/Maybe/README.md
```diff
@@ -0,0 +1,1321 @@
...
+Maybe a = Nothing | Just a
+```
+
+Defined as a Sum Type with it's Left side fixed to `()` (`Nothing`), `Maybe` is
+well suited for capturing disjunction when the cause of the "error" case does
+not need to be communicated. For example, providing default values on specific
+conditions.
+
+A `Maybe` represents disjunction by using (2) constructors, `Nothing` or `Just`.
+A `Just` instance represents the truth case while, `Nothing` is considered
```
rpearce: methinks you don't need a comma here

src/Maybe/README.md
```diff
@@ -0,0 +1,1321 @@
...
+
+#### Just
+
+```haskell
+Maybe.Just :: a -> Maybe a
+```
+
+Used to construct a `Just` instance that represents the "true" portion of a
+disjunction or a valid value. `Just` will wrap any given value in a `Just`,
+which signals the validity of the wrapped value.
```
rpearce: Another comma thing... how do either of these work for ya:
```
`Just` will wrap any given value in a `Just`, and this signals the validity of the wrapped value.
```
or
```
`Just` will wrap any given value in a `Just`, signalling the validity of the wrapped value.
```

src/Maybe/README.md
```diff
@@ -0,0 +1,1321 @@
...
+```
+
+#### of
+
+```haskell
+Maybe.of :: a -> Maybe a
+```
+
+Used to wrap any value into a `Maybe` as a `Just`, `of` is used mostly by
+helper functions that work "generically" with `Applicative` or `Monad` types
```
rpearce: "types" should probs be "type" as "instances" covers the plurality?

src/Maybe/README.md
evilsoft: I :heart: the latter.

src/Maybe/README.md
evilsoft: Actually think we can :hocho: the word `type` altogether.

src/Maybe/README.md
evilsoft: Addressed.

evilsoft:
image

PR #187, opened 2018-01-16
Add Documentation for `Endo`

evilsoft:
## Not just a bike trick anymore
image

This PR addresses adding the `Endo` documentation from [this issue](https://github.com/evilsoft/crocks/issues/42). Once this PR is merged, will do a follow up PR to move this over to the git pages documentation.

### DIFF
--- a/src/Endo/README.md
+++ b/src/Endo/README.md
@@ -1,0 +1,247 @@
+# Endo
+
+```haskell
+Endo a
+```
+
+`Endo` is a `Monoid` that will combine (2) functions that have matching domains
+and codomains (endofunctions) under function composition. Due to the nature
+of `Endo` wrapping a function, the underlying value can either be extracted
+using [`valueOf`](#valueof) like any other `Monoid` or can be executed using
+directly using [`runWith`](#runWith), supplying the input.
+
+```javascript
+const Endo = require('crocks/Endo')
+
+const curry = require('crocks/helpers/curry')
+const mconcat = require('crocks/helpers/mconcat')
+const valueOf = require('crocks/pointfree/valueOf')
+
+// add :: Number -> Number -> Number
+const add =
+  x => y => x + y
+
+// multiply :: Number -> Number -> Number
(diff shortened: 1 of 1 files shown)


src/Endo/README.md
```diff
@@ -0,0 +1,247 @@
+# Endo
+
+```haskell
+Endo a
+```
+
+`Endo` is a `Monoid` that will combine (2) functions that have matching domains
+and codomains (endofunctions) under function composition. Due to the nature
+of `Endo` wrapping a function, the underlying value can either be extracted
+using [`valueOf`](#valueof) like any other `Monoid` or can be executed using
```
foxdonut: `can be executed using directly using` should be `can be executed directly using`

src/Endo/README.md
```diff
@@ -0,0 +1,247 @@
...
+#### empty
+
+```haskell
+Endo.empty :: () -> Endo a
+```
+
+`empty` provides the identity for the `Monoid` in that when the value it
+provides is `concat`ed to any other value, it will return the other value. In
+the case of `Endo` the result of `empty` is the identity function, which echos
+it's input. `empty` is available on both the Constructor and the Instance for
```
foxdonut: should be `its input`

src/Endo/README.md
```diff
@@ -0,0 +1,247 @@
...
+//=> []
+```
+
+#### runWith
+
+```haskell
+Endo a ~> a -> a
+```
+
+`Endo` wraps a function and as such, it's underlying endofunction can be run
```
foxdonut: should be `its underlying endofunction`

src/Endo/README.md
```diff
@@ -0,0 +1,247 @@
...
+while inside of an `Endo` by calling `runWith`. Providing a valid value  of the
```
foxdonut: there are two spaces between `value  of the`

evilsoft:
image

PR #190, opened 2018-01-20
Add the `liftN` helper function

evilsoft:
## Do you even lift?
image

When doing more with Applicative IRL, found the need for one of those `liftN` functions. So this PR addresses that need. The following is done in this:

* Add `liftN` helper and specs
* Document the new function with examples
* Expose it to the public API
* Fix up some signatures that take an arbitrary amount of arguments, for 🌽sistancy


evilsoft:
image

PR #192, opened 2018-01-25
Add `toString` to all ADTs for those without `inspect`

evilsoft:
## Better than Cheese
image

Added a `toString` method each of the ADTs and had to 🔧 up how we deal with `isObject` as that would have created many problems if this PR never became a thing.


evilsoft:
image

PR #193, opened 2018-01-26
Remove mention of the type method from Documentation

evilsoft:
## Transitions are hard
image

With the `isSameType` function being added a while ago, there is not much use for `type` to really be used. So there is going to be the first in a three step transition to move to the more well know `@@type` method and format (lib/type/version).

To keep newcomers from even know about this, the first step to not talk about. Followed by the implementation and hooks for `@@type`. Then the completle and breaking removal of `type` from the ADTs.


evilsoft:
image

PR #194, opened 2018-01-28
Add `@@type` to all ADTs

evilsoft:
## The path to enlightenment
image

This PR is another step in addressing [this issue](https://github.com/evilsoft/crocks/issues/162) to add the `@@type` method/static function to all ADTs. This PR only adds `@@type` and ADT versioning to the ADTs and does not actually utilize it yet.

There will be (2) follow up PRs to this one to complete the actual item on the issue:
1. Switch all type checking to use the core `isSameType` function and have `isSameType` check `@@type`. (non-breaking)
1. Remove `type` from every ADT. (breaking)


evilsoft:
image

PR #195, opened 2018-01-28
Add `First` documentation

evilsoft:
## First things is First
image

This PR addresses another item on [this issue]() and adds the README for the `First` Monoid. Before the next release (2) other READMEs will be added, one for `Async` and another for `Last`. Once all (3) are merged, we will move them over to the overall docs.

### DIFF
--- a/src/First/README.md
+++ b/src/First/README.md
@@ -7,7 +7,7 @@ First a = First (Maybe a)
 `First` is a `Monoid` that will always return the first, non-empty value when
 (2) `First` instances are combined. `First` is able to be a `Monoid` because
 it implements a `Maybe` under the hood. The use of the `Maybe` allows for an
-empty to be represented with a `Nothing`.
+[`empty`][#empty] `First` to be represented with a `Nothing`.
 
 `First` can be constructed with either a value or a `Maybe` instance. Any value
 passed to the constructor will be wrapped in a `Just` to represent a non-empty
@@ -127,11 +127,11 @@ First a ~> a -> a
 ```
 
 `First` wraps an underlying `Maybe` which provides the ability to option out
-a value in the case of an empty instance. Just like `option` on a `Maybe`
-instance, it takes a value as its argument. When run on an empty instance,
-the provided default will be returned. If `option` is run on a non-empty
-instance however, the wrapped value will be extracted not only from the
-`Last` but also from the underlying `Just`.
+a value in the case of an [`empty`](#empty) instance. Just like `option` on
+a `Maybe` instance, it takes a value as its argument. When run on
+an [`empty`](#empty) instance, the provided default will be returned.
+If `option` is run on a non-empty instance however, the wrapped value will be
+extracted not only from the `Last` but also from the underlying `Just`.
 
 If the underlying `Maybe` is desired, the [`valueOf`](#valueof) method can be
 used and will return the `Maybe` instead.
@@ -222,7 +222,7 @@ eitherToFirst :: (a -> Either c b) -> a -> First b
 Used to transform a given `Either` instance to a `First`
 instance, `eitherToFirst` will turn a `Right` instance into a non-empty `First`,
 wrapping the original value contained in the `Right`. All `Left` instances will
-map to an empty `First`, mapping the originally contained value to a `Unit`.
+map to an [`empty`](#empty) `First`, mapping the originally contained value to a `Unit`.
 Values on the `Left` will be lost and as such this transformation is considered
 lossy in that regard.
 
@@ -258,6 +258,7 @@ const firstNumber = mapReduce(
   First.empty()
 )
 
+// "Bad Times" is lost, mapped to Nothing
 eitherToFirst(Left('Bad Times'))
 //=> First( Nothing )
 
@@ -282,8 +283,8 @@ lastToFirst :: (a -> Last b) -> a -> First b
 
 Used to transform a given `Last` instance to a `First` instance, `lastToFirst`
 will turn a non-empty instance into a non-empty `First` wrapping the original
-value contained within the `Last`. All empty instances will map to an empty
-`First`.
+value contained within the `Last`. All [`empty`](#empty) instances will map
+to an [`empty`](#empty) `First`.
 
 Like all `crocks` transformation functions, `lastToFirst` has (2) possible
 signatures and will behave differently when passed either a `Last` instance
@@ -335,7 +336,7 @@ maybeToFirst :: (a -> Maybe b) -> a -> First b
 Used to transform a given `Maybe` instance to a `First`
 instance, `maybeToFirst` will turn a `Just` into a non-empty `First` instance,
 wrapping the original value contained within the `First`.
-All `Nothing` instances will map to an empty `First` instance.
+All `Nothing` instances will map to an [`empty`](#empty) `First` instance.
 
 This function is available mostly for completion sake, as `First` can always
 take a `Maybe` as its argument during construction. So while there is not a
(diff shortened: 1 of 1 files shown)


src/First/README.md
```diff
@@ -0,0 +1,451 @@
+# First
+
+```haskell
+First a = First (Maybe a)
+```
+
+`First` is a `Monoid` that will always return the first, non-empty value when
+(2) `First` instances are combined. `First` is able to be a `Monoid` because
+it implements a `Maybe` under the hood. The use of the `Maybe` allows for an
+empty to be represented with a `Nothing`.
```
rpearce: What is "an empty"? Should this be "an empty _________ " (thing) or is "empty" supposed to be the thing we're talking about?

src/First/README.md
```diff
@@ -0,0 +1,451 @@
...
+
+`First` can be constructed with either a value or a `Maybe` instance. Any value
+passed to the constructor will be wrapped in a `Just` to represent a non-empty
+instance of `First`. Any `Maybe` passed to the constructor will be lifted as
+is, allowing the ability to "choose" an value based on some disjunction.
```
rpearce: `an value` or `a value`?

src/First/README.md
```diff
@@ -0,0 +1,451 @@
...
+const isValid =
+  and(isNumber, isEven)
+
+// chooseFirst :: [ * ] -> First Number
+const chooseFirst =
+  mconcatMap(First, safe(isValid))
+
+chooseFirst([ 21, 45, 2, 22, 19 ])
+  .valueOf(0)
+//=> Just 2
```
rpearce: nice example

src/First/README.md
```diff
@@ -0,0 +1,451 @@
...
+  .concat(First(3))
+//=> First( Just 3 )
+```
+
+## Instance Methods
+
+#### concat
+
+```haskell
+First a ~> First a -> First a
```
rpearce: I might be brainfreezing – what does the `~>` (tilde-arrow) signify?

src/First/README.md
```diff
@@ -0,0 +1,451 @@
...
+```
+
+`concat` is used to combine (2) `Semigroup`s of the same type under an operation
+specified by the `Semigroup`. In the case of `First`, it will always provide the
+first non-empty value. Any subsequent non-empty values will be thrown away, and
```
rpearce: don't need a comma here as no new subject?

src/First/README.md
```diff
@@ -0,0 +1,451 @@
...
+```
+
+#### option
+
+```haskell
+First a ~> a -> a
+```
+
+`First` wraps an underlying `Maybe` which provides the ability to option out
+a value in the case as of an empty instance. Just like `option` on a `Maybe`
```
rpearce: is the `as` needed here? think it works fine as `...to option out a value in the case of an empty instance`

src/First/README.md
```diff
@@ -0,0 +1,451 @@
...
+instance, it takes a value as its argument. When run on an empty instance,
+the provided default will be returned. If `option` is run on a non-empty
+instance however, the wrapped value will be extracted not only from the
+`Last`, but also from the underlying `Just`.
```
rpearce: don't think comma needed

src/First/README.md
```diff
@@ -0,0 +1,451 @@
...
+
+`crocks/First/eitherToFirst`
+
+```haskell
+eitherToFirst :: Either b a -> First a
+eitherToFirst :: (a -> Either c b) -> a -> First b
+```
+
+Used to transform a given `Either` instance to a `First`
+instance, `eitherToFirst` will turn a `Right` instance into a non-empty `First`
```
rpearce: comma after `First` I think?

src/First/README.md
```diff
@@ -0,0 +1,451 @@
...
+wrapping the original value contained in the `Right`. All `Left` instances will
+map to an empty `First`, mapping the originally contained value to a `Unit`.
+Values on the `Left` will be lost and as such this transformation is considered
+lossy in that regard.
```
rpearce: Is `lossy` an understood term? Should you be linking to an article or definition or something? Never heard this term in this context, myself, but I'm a noob...

src/First/README.md
evilsoft: an empty `First`. Say nothing meets the "criteria" like say an empty array when using `mconcat`, or if we are using `Maybe` to check a condition an neither of the values meet the criteria then there is no `First` to return, resulting in an empty `First` (`First( Nothing )`).

Basically for something to be a `Monoid` is must have a value the represents idenity for the operation in that anything `concat`-ed to the identity (or empty) will result in the same value. So `First(3).concat(First.empty())` results in `First( Just 3 )` as does `First.empty().concat(First(3))`

Any suggestions on how I can communicate that in this paragraph without too much theory?

src/First/README.md
evilsoft: `a value`

src/First/README.md
evilsoft: I use the term from Information Theory for source coding, on some encodings, for things like say a bitmap to a jpeg. I do not have an isomorphism from bitmap -> jpeg and jpeg -> bitmap and have them point back to the same exact bitmap on the round trip.

Same thing happens here, because `Either` has both `Left` and `Right` and the target `First` maps all `Lefts` to `Unit`, I can never get back to the same `Left` as I started with, without providing some value to map the `Unit` back into the original `Left`.

So I do not know if it is an equivalent term, but could not come up with one that is quite so close.

Do you have any suggestions on how I could make this more clear to the curious?

src/First/README.md
evilsoft: Except it is misleading, `valueOf` takes no arguments

src/First/README.md
evilsoft: That signifies an instance in which the given method is executed on, so this reads:

When called on an Instance of `First a`, give me a `First a` and I will give back a `First a`.

src/First/README.md
rpearce: I think doing two things could really clear this up for beginners (like me!) and one thing applied everywhere at a later date could make the docs more accessible to beginners:

1. `empty` in the sentence sounds like a descriptor, so maybe using `empty 'First'` would help to clarify it
1. it doesn't have to include theory -- you can show values in a small example. I'm not sure if I'm doing these right, but here we go:
  ```js
  First.empty()                  // => First( Nothing )
  First(3)                       // => First( Just 3 )
  First(3).concat(First.empty()) // => First( Just 3 )
  First.empty().concat(First(3)) // => First( Just 3 )
  ```

* separately, all the category theory and ADTs are awesome for people that are familiar, but it'd also be potentially useful to point to the fantasy land spec or wikipedia or articles or something the first time a term is used in any doc.

src/First/README.md
rpearce: I WAS WONDERING ABOUT THAT. I mean the example as a whole 🤕

src/First/README.md
rpearce: A link to something on the subject is always a nice thing to have.

evilsoft:
@rpearce your questions are making me think that a brief "primer" on some basic category theory, type signatures and typeclasses may be in order.
Maybe I should extend the getting started section to address this.

What do you think about something like that?

src/First/README.md
evilsoft: I like everything about this and think I can address these like this:
1. Add just like you suggest with empty `First` instance in the cases I speak about empty instance.
1. As the examples as described are shown in the `empty` static function, so what do you think about me adding a link to `#empty` where I have the word empty?
1. So I think that the docs could benefit from a section on basic category theory, typeclasses and signatures may address the reference issue.

Do you think that will do it?

src/First/README.md
evilsoft: Since this occurs any time we have a transformation with most sum types to `Maybe`, it will end up being a common theme. 
So I think that primer I mentioned a couple times could have a section on natural transformations with Sum Types.

But in the meantime, how would you feel about emphasizing this point in the code example with a comment over [this line](https://github.com/evilsoft/crocks/pull/195/files#diff-dfe56abb2d91236bdea5f40bdcb447d7R260) to the effect of:
```js
// 'Bad Times' is lost
eitherToFirst(Left('Bad Times'))
//=> First( Nothing )
```

Will also do the same with `resultToFirst`.
Do you think that may help?

src/First/README.md
rpearce: A link to the `empty` function would be real nice.

With regards to `3.`: if there are solid resources out there that can be linked to or copied+pasted+cited to handle this, then I would advise not recreating the wheel. However, if this could increase the accessibility and understanding of the project more than external links could and can provide more of a community benefit, then that's a noble thing _if you have the time_. Heck, maybe the docs get turned into a book.

rpearce: Makes sense

evilsoft:
Looks like I was able to address (for the most part), the provided feedback.
Will follow up with an issue to add Guides for the following:
* typeclasses (Monoids, Alts, Functors, etc)
* type signatures
* Natural Transformations

evilsoft:
image

PR #197, opened 2018-01-31
Add `Last` Documentation

evilsoft:
## Last but not Least
image

This is another PR that addresses the `Last` item on [this issue](https://github.com/evilsoft/crocks/issues/42) by adding the README for the `Last` monoid. It is pretty much the same thing as `First`, but kinda the opposite.

Also made some corrections to the `First` README in this PR as well.

### DIFF
--- a/src/Last/README.md
+++ b/src/Last/README.md
@@ -31,7 +31,7 @@ const safe = require('crocks/Maybe/safe')
 
 // isEven :: Number -> Boolean
 const isEven =
-  x => !(x % 2)
+  x => x % 2 === 0
 
 // isValid :: a -> Boolean
 const isValid =
--- a/src/Last/README.md
+++ b/src/Last/README.md
@@ -7,7 +7,7 @@ Last a = Last (Maybe a)
 `Last` is a `Monoid` that will always return the last, non-empty value when
 (2) `Last` instances are combined. `Last` is able to be a `Monoid` because
 it implements a `Maybe` under the hood. The use of the `Maybe` allows for an
-[`empty`][#empty] `Last` to be represented with a `Nothing`.
+[`empty`](#empty) `Last` to be represented with a `Nothing`.
 
 `Last` can be constructed with either a value or a `Maybe` instance. Any value
 passed to the constructor will be wrapped in a `Just` to represent a non-empty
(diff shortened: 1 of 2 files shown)


src/Last/README.md
```diff
@@ -0,0 +1,453 @@
+# Last
+
+```haskell
+Last a = Last (Maybe a)
+```
+
+`Last` is a `Monoid` that will always return the last, non-empty value when
+(2) `Last` instances are combined. `Last` is able to be a `Monoid` because
+it implements a `Maybe` under the hood. The use of the `Maybe` allows for an
+[`empty`][#empty] `Last` to be represented with a `Nothing`.
```
foxdonut:
```
[`empty`][#empty] should be [`empty`](#empty)
```
Note that this is also in `First.md`.

src/Last/README.md
```diff
@@ -0,0 +1,453 @@
...
+
+Last(Just(99))
+//=> Last( Just 99 )
+
+Last(Nothing())
+//=> Last( Nothing )
+
+lastNumVal({ val: 97 })
+  .concat(Last(80))
+//=> Last( Just 97 )
```
foxdonut: Not sure what this example is supposed to be, but the result is incorrect. The following example is exactly the same but has the correct result.

evilsoft: That should have been switched to:
```js
Last(Just(80))
  .concat(lastNumVal({ val: 97 }))
//=> Last( Just 97 )
```

src/Last/README.md
```diff
@@ -0,0 +1,456 @@
...
+const Last = require('crocks/Last')
+
+const and = require('crocks/logic/and')
+const isNumber = require('crocks/predicates/isNumber')
+const mconcatMap = require('crocks/helpers/mconcatMap')
+const safe = require('crocks/Maybe/safe')
+
+// isEven :: Number -> Boolean
+const isEven =
+  x => !(x % 2)
```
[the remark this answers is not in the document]
evilsoft: Good call.
🔧-ing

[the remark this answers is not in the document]
rpearce: ty ❤️

evilsoft:
image

PR #198, opened 2018-02-04
Apply traversals in the proper order, add `reduceRight` to `List`

evilsoft:
## Well that is just silly

image

Turns out `Traversals` were not being applied in the proper order for `Array` and `List`. This PR fixes that so that traversing over an `Array` or `List` is built and executed as expected. In order to get this to fix right in, I needed to add `reduceRight` to `List`.

So all in all, this PR does the following:
* Apply traversals on `Array` in the proper order.
* Apply traversals on `List` in the proper order.
* Add `reduceRight` to `List`
* Add `reduceRight` to pointfree functions
* Expose new pointfree `reduceRight` on public API
* Add documentation for `reduceRight`
* 🔧 some specs.


evilsoft:
image

PR #200, opened 2018-02-08
Add Fantasy Land Prefixed Method Names

evilsoft:
## FINALLY (-ish)
image

This PR addresses the last item on [this issue](https://github.com/evilsoft/crocks/issues/162) and adds the remaining bits for what `crocks` will implement for the [fantasy-land spec](https://github.com/fantasyland/fantasy-land).

This adds the prefixed method names, which is the 🍒 on the 🍰. For usability reasons we will not be implementing `fantasy-land/ap` and `fantasy-land/traverse`, those will stay with the pre-1.x version of the API.

### DIFF
--- a/src/All/All.spec.js
+++ b/src/All/All.spec.js
@@ -41,6 +41,17 @@ test('All', t => {
   t.end()
 })
 
+test('All fantasy-land api', t => {
+  const m = All(true)
+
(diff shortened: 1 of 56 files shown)


evilsoft:
image

PR #204, opened 2018-02-13
Add `Async` Documentation

evilsoft:
## At the SAME time
image

This PR is mostly centered around adding the `Async` docs to address [this issue](https://github.com/evilsoft/crocks/issues/42).

Like many of the other docs added, I took the time to address some legacy error tests that have been providing false positives. Will need to circle back on some of the spec updates for some of the `Monoid`s that have already been documented. TBH I should have done a separate PR to address the specs, and WILL do so in the future to keep review time down. Also corrected a spelling error in the `Maybe` docs which really should have ALSO been its own PR.


src/Async/README.md
```diff
@@ -0,0 +1,1369 @@
...
+//=> rej: { error: "null is not valid", hasError: true }
+```
+
+#### ap
+
+```haskell
+Async e (a -> b) ~> Async e a -> Async e b
+```
+
+Short for apply, `ap` is used to apply an `Async` instance containing a value
```
HenriqueLimas: Any reason to not use `apply` instead of `ap`? IMHO `ap` can be confused. (Probably this is not related to that PR)

src/Async/README.md
```diff
@@ -0,0 +1,1369 @@
...
+The first and more common signature takes (2) functions that will have their
+return values ignored. The first function will be run in the event of the
+`Async` instance settling on `Rejected` and will receive as its single argument
+the value or "cause" of rejection. The second function provided will be executed
+in the case of the instance settling on `Resolved` and will receive as its
+single argument the value the `Async` was resolved with.
+
+The second signature is used when any cleanup needs to be performed after a
+given `Async` is cancelled by having the function returned from `fork` called.
+The first to arguments to the signature are the same as the more common
```
HenriqueLimas: Some little typos here (in bold):
The first t**w**o arguments to the signature are the same as the more common
signature described above but takes an addition**al** function that can be used
for "**clean up**" after cancellation.

src/Async/README.md
```diff
@@ -0,0 +1,1369 @@
...
+a `Resolved` instance wrapping the original value contained in the
+original non-empty.
+
+The `Last` datatype is based on a `Maybe` and as such its left or empty value
+is fixed to a `()` type. As a means to allow for convenient
+transformation, `lastToAsync` takes a default `Rejected` value as the first
+argument. This value will be wrapped in a resulting `Rejected` instance, in the
+case of empty.
+
+Like all `crocks` transformation functions, `firstToAsync` has (2) possible
```
HenriqueLimas: `lastToAsync` instead of `firstToAsync`

src/Async/Async.spec.js
```diff
@@ -118,36 +119,37 @@ test('Async Resolved', t => {
 })
 
 test('Async fromPromise', t => {
-  const resProm = x => new Promise((res) => res(x))
+  const resProm = x => new Promise(res => res(x))
 
   t.ok(isFunction(Async.fromPromise), 'is a function')
   t.ok(isFunction(Async.fromPromise(resProm)), 'returns a function')
 
   const fn = bindFunc(Async.fromPromise)
```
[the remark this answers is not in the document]
evilsoft: Typically I use `fn` for a function, you will see that mostly with error testing functions.
Now `a` and `m` should probably, eventually be replaced with just `m`...I use `m` to represent a given ADT.

src/Async/README.md
evilsoft: `ap` is defined by Fantasy-Land. Say we want to have "static" functions on our constructor (we do not do this in crocks, but some libs do), `apply` would override the `Function.prototype.apply`. Even though we do not implement the `ap` spec as defined by Fantasy-Land, it makes sense to still keep the naming 🌽sistant

src/Async/README.md
```diff
@@ -0,0 +1,1369 @@
...
+Defined as a "lazy" Sum Type that implements an asynchronous control structure,
+`Async` represents either the success or failure of a given asynchronous
+operation. The "laziness" of `Async` allows the ability to build complex
+asynchronous operations by defining each portion of that flow as smaller "steps"
+that can be composed together.
+
+Depending on your needs, an `Async` can be constructed in a variety of ways. The
+typical closely resembles how a `Promise` is constructed with one major
+difference, the arguments used in the function that is passed to the `Promise`
+constructor are reversed in an `Async`.
```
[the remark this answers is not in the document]
evilsoft: It may not be the right place to explain it in the docs. But I can do it here: So it is typical and easier to follow when the far right parameter of a type is the covariant functor (the portion that works with `map`). So  because of this, the `Resolved` is the parameter that works with `map`. So to keep everything lined up and easier to follow, the arguments to the functions are in the same order.

src/Async/README.md
HenriqueLimas: Yeah, was thinking something about the `apply` method, but since was an object I asked it. But yeah, now make sense. 👍

src/Async/README.md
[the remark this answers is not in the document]
evilsoft: That said I think I can throw something in like:
> ... reversed in an `Async` *to match the order in which `Async` is parameterized*.

Do you think that could provide some hint to why?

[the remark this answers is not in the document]
HenriqueLimas: Yes, it can help IMHO, but it's your call here.

HenriqueLimas:
@evilsoft do you think is better to put everything in one single commit?

evilsoft:
@HenriqueLimas Not really, what I should have done, to make it easier to review, is put the Specs in its own commit (well really its own PR), 
But until a PR is made, I usually keep it to one. This makes rebasing and other branch management tasks easier.

Once I PR it publicly it though, I never amend. That way the:
* Reviewer never has to kill their local before pulling down a force push 
* Easier to review a commit to see what was changed after feedback was addressed.
* it gives a nice breadcrumb trail on changes from PR feedback.

But I still rebase if their are merge conflicts during the review cycle.

evilsoft:
Okay going to call this done. Merging.
image

PR #210, opened 2018-02-14
Add Async crock to Documentation

evilsoft:
## Yet Another 
image

This PR adds the `Async` README to the documentation. That is all.

### DIFF
--- a/docs/src/pages/docs/crocks/Async.md
+++ b/docs/src/pages/docs/crocks/Async.md
@@ -1,0 +1,1357 @@
+---
+title: "Async"
+description: "Async Crock"
+layout: "guide"
+weight: 20
+---
+
+```haskell
+Async e a = Rejected e | Resolved a
+```
+
+Defined as a "lazy" Sum Type that implements an asynchronous control structure,
+`Async` represents either the success or failure of a given asynchronous
+operation. The "laziness" of `Async` allows the ability to build complex
+asynchronous operations by defining each portion of that flow as smaller "steps"
+that can be composed together.
+
+Depending on your needs, an `Async` can be constructed in a variety of ways. The
+typical closely resembles how a `Promise` is constructed with one major
+difference, the arguments used in the function that is passed to the `Promise`
+constructor are reversed in an `Async` to match the order in which Async is
+parameterized.
+
+There are many ways to represent asynchronous operations in JavaScript, and as
+such, the libraries available to us in our ecosystem provide different means
+to take advantage of these operations. The two most common use
+either `Promise` returning functions or allow for the Continuation Passing Style
+prevalent in the asynchronous function that ship with NodeJS. `Async` provides
+(2) construction helpers that wrap functions using these styles of asynchronous
+processing and will give you back a function that takes the same arguments as
+the original and will return an `Async` instead. These functions are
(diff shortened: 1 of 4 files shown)


docs/src/pages/docs/crocks/Async.md
```diff
@@ -0,0 +1,1357 @@
...
+
+// length :: String -> Integer
+const length =
+  x => x.length
+
+// repeat :: String -> Number -> String
+const repeat =
+  char => n => char.repeat(n)
+
+// values :: Async String Number -> Async String Number
```
karthikiyengar: I'm still pretty new to reading signatures, but this doesn't feel right? Did you mean `values :: Async String Number -> Async Number String`?

docs/src/pages/docs/crocks/Async.md
```diff
@@ -0,0 +1,1357 @@
...
+  regex.test(str)
+
+// lookup :: String -> Async String String
+const lookup = compose(
+  maybeToAsync('Not Found'),
+  flip(prop, { 'file-a': 'file-b', 'file-b': 'file-c' })
+)
+
+// fake :: String -> Async String Object
+const fake = composeK(
```
karthikiyengar: This example assumes that the reader is already familiar with what `composeK` does. Do you think being more explicit would help?

evilsoft: I put function like that in these examples as teasers, do people will be like what is this `composeK` and then go look it up. Like reading a book with a Dictionary, do you think it would be more helpful to the reader to do this instead:
```js
const fake = compose(
  chain(lookup),
  chain(safe(test(/^file-(a|b|c)/)),
  chain(safe(isString))
)
```

docs/src/pages/docs/crocks/Async.md
evilsoft: Nope, that siggy is correct as is. In this case we are map values to their respective types:

```js
// repeat :: String -> Number -> String
// length :: String -> Integer
```

so the `Rejected` which runs through `length`, will take the `String` and deposit a `Number` in the `Resolved` and the `repeat('a')` will take a `Resolved` `Number` and map it to a `String` before dumping it into a `Rejected`.

Using the functions to allow mapping keep from the `.swap().bimap(repeat('a'), length)` which for me at least occurs more often then just a straight `.swap()`. So the `crocks` `swaps` allow for mapping values.

If that makes sense?

karthikiyengar: Sure, thanks for the clarification :-)

docs/src/pages/docs/crocks/Async.md
karthikiyengar: Sure, we can retain the `composeK`. The only reason I felt that way because the section header was `chain` but there was no reference to it in the code snippet.

evilsoft: That is an excellent point! Good call! I will make that change for this one, and leave the other one as `composeK`, which will give a hint as to what the heck `composeK` could mean. What do you think?

karthikiyengar: Sure, I think that makes sense.

evilsoft:
image

PR #211, opened 2018-02-15
Add `sequence` and `traverse` to `Pair`

evilsoft:
## Super Power that Pair
image

Thought I had already added `traverse` and `sequence` to both the `Pair` and `Writer` types...until I went to write an example for someone. :smh:

This PR adds those methods to `Pair` and each will traverse over the far right (second) parameter. So you can turn your `Pair Array (Maybe a)` into a `Maybe (Pair Array a)`...all the good things.


evilsoft:
image

PR #212, opened 2018-02-15
Update `traverse` and `sequence` Errors to signal for `Apply` and not `Applicative`

evilsoft:
## The Times are a Changin'

image

Because we are not implementing the fantasy-land methods for the public `traverse` and `sequence` methods, we only need to constrain to `Apply` and not as strict as `Applicative` as `Apply` returning functions are all that are needed.

So we should report that 🌽striant properly in our Errors. That, my friends, is all this PR does.


evilsoft:
image

PR #213, opened 2018-02-15
Update Error specs for the remaining Monoids

evilsoft:
## Some things just need to be done
image

As this next release will see the completion of all Monoid Documentation, figured now is the best time to bring the Error Reporting specs up to par with a majority of the lib. That is all this PR does is get the proper error checks in to check against the message.

This will be the first part before the final `Max` and `Min` docs are submitted.


evilsoft:
image

Issue #202, opened 2018-02-12
Add fantasy-land integration for valid `pointfree` functions


evilsoft:
Now that we have the types working with fantasy-land functions from other libs, makes sense to make the integration both ways. All of the pointfree functions that have corresponding fantasy-land definitions should check for the namespace. These functions should check if the fantasy-land methods exist first and then fall back to the "public" if it does not exist.

evilsoft:
Got the last bit for this done (except for `traverse` and `ap`).
closing.

PR #214, opened 2018-02-15
Move Fantasy-Land method name into one place

evilsoft:
## The Start of Something Abstract
image

In response to [this issue](https://github.com/evilsoft/crocks/issues/202) we will need to have these names peppered in even more places. Makes sense to start abstracting this sooner rather than later. Eventually it will be used to invoke the function and also will be needed for comparison.

Also should clear up some of the errors when Travis runs node 6 due to the way buble was handling `@@type` before running the spec suite.


evilsoft:
image

PR #160, opened 2017-12-18
`propPath` will no longer throw with `NaN` and `null` values.

evilsoft:
## A Bugs Path
image

When reviewing [this PR](https://github.com/evilsoft/crocks/pull/158) found out that `propPath` was not being very nice with `null`s and `NaN`s in the path. That spawned [this issue](https://github.com/evilsoft/crocks/issues/159) which in turn spawned this PR.

This PR adds further :corn:straint on acceptable values. Now `Nothing` will return whenever `NaN`, `null` or `undefined` is a value. If this becomes a problem we can totes revisit, but this seems reasonable for now.

PR #217, opened 2018-02-15
Fix issue in propPathOr if default value is an object

jonwhelan:
In the previous implementation, if a property in the target is not found,
then the default value would be returned as the accumulator to the next
iteration of reduce. If the default value is an object, then it is
errantly traversed with the next property in the specified path.

### DIFF
--- a/src/helpers/propPathOr.js
+++ b/src/helpers/propPathOr.js
@@ -1,13 +1,11 @@
 /** @license ISC License (c) copyright 2017 original and current authors */
 /** @author Henrique Limas */
 
-const constant = require('../combinators/constant')
 const curry = require('../core/curry')
-const either = require('../pointfree/either')
-const identity = require('../combinators/identity')
-const isArray = require('../predicates/isArray')
-const isNil = require('../predicates/isNil')
-const propPath = require('../Maybe/propPath')
+const isArray = require('../core/isArray')
+const isInteger = require('../core/isInteger')
+const isNil= require('../core/isNil')
+const isString = require('../core/isString')
 
 // propPathOr : a -> [ String | Integer ] -> b -> c
 function propPathOr(def, keys, target) {
@@ -19,8 +17,17 @@ function propPathOr(def, keys, target) {
     return def
   }
 
-  return either(constant(def), identity, propPath(keys, target))
+  return keys.reduce((target, key) => {
+    if (isNil(target)) {
+      return target
+    }
+
+    if(!(isString(key) || isInteger(key))) {
+      throw new TypeError('propPathOr: Array of strings or integers required for second argument')
+    }
+
+    return target[key]
+  }, target) || def
 }
 
 module.exports = curry(propPathOr)
-
--- a/src/helpers/propPathOr.js
+++ b/src/helpers/propPathOr.js
@@ -13,10 +13,6 @@ function propPathOr(def, keys, target) {
     throw new TypeError('propPathOr: Array of strings or integers required for second argument')
   }
 
-  if(isNil(target)) {
-    return def
-  }
-
   const value = keys.reduce((target, key) => {
     if (isNil(target)) {
       return target
--- a/src/helpers/propPathOr.js
+++ b/src/helpers/propPathOr.js
@@ -1,14 +1,13 @@
 /** @license ISC License (c) copyright 2017 original and current authors */
 /** @author Henrique Limas */
 
+const constant = require('../combinators/constant')
 const curry = require('../core/curry')
-const isArray = require('../core/isArray')
-const isInteger = require('../core/isInteger')
-const isNil= require('../core/isNil')
-const isString = require('../core/isString')
-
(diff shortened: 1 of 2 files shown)


src/helpers/propPathOr.js
```diff
@@ -1,14 +1,13 @@
...
-const isNil= require('../core/isNil')
-const isString = require('../core/isString')
-
-const lift = (x, def) =>
-  isNil(x) ? def : x
+const either = require('../pointfree/either')
+const identity = require('../combinators/identity')
+const isArray = require('../predicates/isArray')
+const isNil = require('../predicates/isNil')
+const propPath = require('../Maybe/propPath')
```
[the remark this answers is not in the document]
jonwhelan: Hey, thanks for the detailed explanation of the project conventions. All these reasons make perfect sense. I think my latest push is more in line with your suggested fix. Let me know if that's what you were thinking.

evilsoft:
Looks good, I am out and about right now, but this will be merged in and go out with `0.9.0` which will be release tonight when I get off of work.
Also as a side note: If you want you can add your information in an `@author` tag up top and if you want on this pr do a `yarn add-contributor jonwhelan bug,code` and that will add you to the collabortors section at the bottom of the main `README`. just commit these changes in on your branch. (if you want to that is).

jonwhelan:
Oops, introduced a different bug in previous commit where we'd return the default if we found a falsy value in the object. Added additional test case and fixed in latest commit

src/helpers/propPathOr.js
```diff
@@ -20,13 +17,19 @@ function propPathOr(def, keys, target) {
...
+
     if(!(isString(key) || isInteger(key))) {
       throw new TypeError('propPathOr: Array of strings or integers required for second argument')
     }
 
-    return lift(target[key], def)
+    return target[key]
   }, target)
+
+  return isNil(value) ? def : value
```
evilsoft: 👍

evilsoft:
Actually, Just thought of something else. `null` could be a desired value, I think as this will be going out on a breaking change, we should only provide the default on `undefined`. As far as a code change, that should just be swapping out the last `isNil` to use `isDefined` that way: 

```js
const propPath = require('crocks/helpers/propPath')

const fn =
  propPath('Nope', [ 'a', 'b' ])

fn({ a: { b: null } })
//=> null

fn({ a: null })
//=> 'Nope'
```

Any thoughts around that?

src/helpers/propPathOr.js
```diff
@@ -20,13 +17,19 @@ function propPathOr(def, keys, target) {
     return def
```
HenriqueLimas: I think that `if` condition is not needed anymore since the `reduce` will check that.

jonwhelan:
@evilsoft Agree with your conclusion on the `null` case. In that same vein couldn't `undefined` also be the desired value? Using your example, if property `b` had been explicitly set to `undefined` we might want to return that.

It all comes down to what the semantics of the function should be:
a.) Only return default if path not found
b.) Return default if path not found, or value at path is not defined

I'd lean toward option `a` as it doesn't make an extra assumption.

evilsoft:
My reasoning for allowing `null` is that if I have the following:
```js
const obj = { a: 546 }

obj.b
//=> undefined
```
It is a common theme in the lib with many functions filtering out anything set to undefined (mostly to do with database integration libs out there, they 🌽sider undefined as a NULL when inserting and some other libs out there do other things). Really though, this is why I prefer `prop` and `propPath` as I can provide predicates to handle types.

All of that said, I am thinking your intuition may be correct on this. But I need to think about this and the implications. If we do go with b though, it would match the pattern (these functions and `prop` and `propPath` are an oversight on my part) established in the rest of the lib.

If we go with `a` we will need to adjust the following functions to deal with the new meaning of `undefined`:
* `assign`
* `defaultProps`
* `dissoc`
* `omit`
* `pick`
* `Assign` Monoid

jonwhelan:
To limit the scope of this PR, perhaps we should keep the established pattern on the meaning of undefined and open up a separate issue to reconsider a later breaking change to those semantics?

If you think this is the right course, then I'll go ahead and make the 2 outstanding changes to this PR:
1. Swap `isDefined` in for `isNil`
2. Remove redundant conditional as pointed out by @HenriqueLimas

evilsoft:
Sorry to keep this going but it looks like [this PR](https://github.com/evilsoft/crocks/pull/160) made the decision for us. So lets keep the rejection of `null`, `undefined` and `NaN` like you have and remove the redundant 🌽ditional as suggested. Then we can circle back if people do not like it.

evilsoft:
Okay I say this is GTG. do you want to add yourself to the contributor list, by running `yarn contributors:add jonwhelan bug,code` in your terminal?

evilsoft:
Okay, I will get this in. I will add you to the contributors page....In the mean time, have a squirrel:
image

Issue #215, opened 2018-02-15
Need to move `@@type` methods to `String` primatives


evilsoft:
## You should at LEAST get it right

So I jumped the gun on Sanctuary Type Identifiers. They need to be String primitives and NOT functions that return a `String` primitive. See when happens when we try to integrate...Patterns that contradict the already established pattern of methods everywhere :smh:. Even though it is kinda gross, we already started this Team Player route, so we should finish it!!

But I think this is as far as we are going to go, sanctuary types and as much Fantasy Land as we can muster.

evilsoft:
Totally open!!

evilsoft:
This has been resolved with 0.9.0.
Closing

PR #218, opened 2018-02-16
move over to sanctuary string types

evilsoft:
## Everywhere I say.
image

Okay, jumped the gun on Sanctuary integration, this is the first step in getting this to work with Sanctuary Defs. This PR just moves all of the new `@@type` methods over to new `@@type` `String` properties.

This is against [this issue](https://github.com/evilsoft/crocks/issues/215)

### DIFF
--- a/src/Reader/ReaderT.js
+++ b/src/Reader/ReaderT.js
@@ -1,9 +1,12 @@
 /** @license ISC License (c) copyright 2017 original and current authors */
 /** @author Ian Hofmann-Hicks (evil) */
 
+const VERSION = 1
+
 const _implements = require('../core/implements')
 const _inspect = require('../core/inspect')
 const _type = require('../core/types').type('Reader')()
+const _typeString = require('../core/types').typeFn(_type, VERSION)
 const fl = require('../core/flNames')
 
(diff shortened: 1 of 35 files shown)


src/Reader/ReaderT.js
```diff
@@ -1,9 +1,12 @@
 /** @license ISC License (c) copyright 2017 original and current authors */
 /** @author Ian Hofmann-Hicks (evil) */
 
+const VERSION = 1
```
HenriqueLimas: What is the main goal of `VERSION` here? isn't it better to put it in another file that others can use the same version?

evilsoft: That is to report the version of the specific type for the Sanctuary Type siggy. So each type may have its own VERSION, for instance if you look at `Pair` that is a version 2 now, because the API changed. I was thinking that it would make sense that that stays with the type in the file.

What are you suggesting? Like a lookup table?

HenriqueLimas: I thought was the same version for each one. So yeah, put it in the same file its ok.

evilsoft:
image

Issue #205, opened 2018-02-13
Add `propEq` and `propPathEq` to predicates


evilsoft:
These are a pair of handy predicates that should be included in the library. 

They should have the following signatures:
* `propEq :: String -> a -> Object -> Boolean`
* `propPathEq :: [ String ] -> a -> Object -> Boolean`
 
Both should return false if the key does not exist. and should throw in the case of String (or [ String ]) or Object not being the proper types. Under the hood `src/core/equals.js` should be used for equality by value.

The following needs to be accomplished:
* Add code, each function in its own file.
  - [ ] `src/predicates/propEq.js`
  - [ ] `src/predicates/propPathEq.js`
* One spec file per function (`src/predicates`)
  - [ ] `src/predicates/propEq.spec.js`
  - [ ] `src/predicates/propPathEq.spec.js`
* Documentation (`/src/pages/docs/functions/predicate-functions.md`)
  - [ ] Entry for` propEq`
  - [ ] Entry for` propPathEq`
* Expose new functions to public API (`src/index.js`, in predicates section)
  - [ ] Entry for` propEq`
  - [ ] Entry for` propPathEq`
* Add specs for new public functions (`src/index.spec.js`)
  - [ ] Entry for` propEq`
  - [ ] Entry for` propPathEq`

karthikiyengar:
For reasons of consistency, should it be `propEq :: (String | Number) -> a -> Object -> Boolean`, similar to `hasProp`

evilsoft:
Yeppers!! You got it!

karthikiyengar:
A few quick questions about `propPathEq`
1. I'm planning to use the already implemented `propPath` for this one, but it's not a part of `src/core`. Thoughts?
2. Since `propPath` returns a `Maybe`, should we also do an unboxed equality check? In other words, should `propPathEq(["some", "key"], 5, { some: { key: 5 } })` be true?

evilsoft:
This has been merged.
Closing!!

PR #219, opened 2018-02-16
Added propEq and propPathEq (#205)

karthikiyengar:


src/predicates/propPathEq.spec.js
```diff
@@ -0,0 +1,53 @@
...
+
+
+  const obj = { some: { key: 5, hah: null } }
+  const goodPath = [ 'some', 'key' ]
+  const nullPath = [ 'some', 'hah' ]
+  const badPath = [ 'this', 'is', 'really', 'bad' ]
+
+  t.equals(propPathEq(goodPath, Just(5), obj), true, 'returns true on correct path')
+  t.equals(propPathEq(badPath, Just(5), obj), false, 'returns false on incorrect path')
+  t.equals(propPathEq(nullPath, null, obj), false, 'returns false when comparing with a null value present at the path')
```
karthikiyengar: @evilsoft - Not sure if this is intended behaviour (see `pathProp` documentation)

karthikiyengar:
I'm not sure what's happening here, the tests pass on `npm run spec:dev` but they fail on `npm run spec:build`. I've tried running them 10 times :-(

evilsoft:
@karthikiyengar what happens when you run `yarn test` or `npm test`

.gitignore
```diff
@@ -8,6 +8,7 @@ coverage
 
 *.log
 
+.vscode
```
evilsoft: 👍

src/predicates/propPathEq.js
```diff
@@ -0,0 +1,35 @@
...
+const Maybe = require('../core/Maybe')
+
+const { Just, Nothing } = Maybe
+
+const lift = x => (!isNil(x) ? Just(x) : Nothing())
+
+// propPathEq :: [String | Number] -> a -> Object -> Boolean
+function propPathEq(path, key, target) {
+  if (!isArray(path)) {
+    throw new TypeError(
```
evilsoft: I ❤️ this style choice. Will be stealing that from you!!

src/predicates/propPathEq.js
```diff
@@ -0,0 +1,35 @@
...
+const isString = require('../core/isString')
+const equals = require('../core/equals')
+const Maybe = require('../core/Maybe')
+
+const { Just, Nothing } = Maybe
+
+const lift = x => (!isNil(x) ? Just(x) : Nothing())
+
+// propPathEq :: [String | Number] -> a -> Object -> Boolean
+function propPathEq(path, key, target) {
```
evilsoft: To avoid 🌽fusion, `key` should be renamed to something like `value` or something better if you have a better name.

src/predicates/propPathEq.js
```diff
@@ -0,0 +1,35 @@
+/** @license ISC License (c) copyright 2017 original and current authors */
+/** @author Karthik Iyengar (karthikiyengar) */
+
+const curry = require('../core/curry')
+const isArray = require('../core/isArray')
+const isInteger = require('../core/isInteger')
+const isNil= require('../core/isNil')
+const isString = require('../core/isString')
+const equals = require('../core/equals')
+const Maybe = require('../core/Maybe')
```
evilsoft: Do not think we need to go the `Maybe` route here. Using the types add a couple "cycles" for lifting and building the type. While this is just fine on the "user of crocks" end to capture the safety of the `Maybe` against whatever predicate they desire. I feel that as a lib, we should probably go the route of just "doing it by hand" without the `Maybe`. just iterate down.

This gets me thinking though, and this is out of scope for this PR, but maybe we should do for loops in these instances so we get an early exit on a long list and do that everywhere we path down. Do not worry about that for this, we can go back later and do all these places at once.

But what I am thinking for this PR is something like:

```js
path.reduce((acc, key) => {
  if (!(isString(key) || isInteger(key))) {
    throw new TypeError(
      'propPathEq: Array of strings or integers required for first argument'
    )
  }

  if(isNil(acc)) {
    return acc
  }

  return acc[key]
})
```

(or something like that, I did not run the code, but the gist is that if we are a value that can error from us trying to reference properties like `null` and `undefined`, we just keep whatever that value is and pass it down. `NaN` is included in `isNil`, but that is going to be a huge edge case if someone wants to access `NaN`s properties.

Thoughts?

src/predicates/propPathEq.spec.js
evilsoft: Nope I do not believe it should be. I think we would want all comps from `undefined`, `null` and `NaN` to be able to be compared to one another. getting rid of the Maybe should fix that case up.

evilsoft:
Okay I know what is going on.
It has to do with how the compiled code is being written.
A LOT of it has to do with reusing that state in the specs.
I will amend a some feedback to my initial review around the specs.

src/predicates/propEq.spec.js
```diff
@@ -0,0 +1,52 @@
...
+  t.throws(f(key, val, 'hello'), err3, 'throws with string in third argument')
+  t.throws(f(key, val, []), err3, 'throws with array in third argument')
+  t.throws(f(key, val, unit), err3, 'throws with function in third argument')
+  t.ok(f(key, val, {}), 'accepts an object in the third argument')
+
+  t.equals(propEq(key, val, { [key]: undefined }), false, 'returns false when value does not exist on object')
+  t.equals(propEq(key, '5', { [key]: 5 }), false, 'returns false for truthy string')
+  t.equals(propEq(key, 5, { [key]: '5' }), false, 'returns false for truthy number')
+
+  t.equals(propEq(key, val, obj), true, 'returns true when value exists on object')
```
evilsoft: So the reason this is failing is due to compilation issues, when `[key]: ` is used above, buble creates an `obj` to which it uses in a sequence expression so when it checks to see if the value is true here, it is not. it is actually '5'.

So to get around this should probably just loose the obj assignment and use a literal in these cases like you did above:
```js
t.equals(propEq('a', 232, { a: 232 }), true, 'returns true when value exists on object')
```

That will clear up the error. Or you can pick a different name then `obj` if you wanna keep it like this

src/predicates/propEq.spec.js
```diff
@@ -0,0 +1,52 @@
...
+const unit = require('../core/_unit')
+
+const propEq = require('./propEq')
+
+test('propEq function', t => {
+  const f = bindFunc(propEq)
+
+  t.ok(isFunction(propEq), 'is a function')
+
+  const err1 = /propEq: Number or String required for first argument/
```
evilsoft: Would be nice to have these declared above the grouping they are being checked against. Makes it easier when adjusting specs in the future.

src/index.js
```diff
@@ -146,6 +146,8 @@ const pointfree = {
 
 const predicates = {
   hasProp: require('./predicates/hasProp'),
+  propEq: require('./predicates/propEq'),
```
evilsoft: These should be in alphabetical order.

src/index.spec.js
```diff
@@ -139,6 +139,8 @@ const valueOf = require('./pointfree/valueOf')
 
 // predicates
 const hasProp = require('./predicates/hasProp')
+const propEq = require('./predicates/propEq')
```
evilsoft: Should be in alphabetical order

docs/src/pages/docs/functions/predicate-functions.md
```diff
@@ -14,7 +14,9 @@ from `crocks/predicates`.
 Below is a list of all the current predicates that are included with a
 description of their truth:
 
-* `hasProp :: (String | Number) -> a -> Boolean`: An `Array` or `Object` that contains the provided index or key
+* `hasProp :: (String | Number) -> a -> Boolean`: an `Array` or `Object` that contains the provided index or key
+* `propEq: (String | Number) -> a -> Object -> Boolean`: an `Object` that contains the provided key
```
evilsoft: These should appear in alphabetical order

src/Maybe/propPath.spec.js
```diff
@@ -25,7 +25,7 @@ test('propPath function', t => {
 
   t.throws(p([ undefined ], {}), err, 'throws with an array of undefined in first argument')
   t.throws(p([ null ], {}), err, 'throws with array of null in first argument')
-  t.throws(p(false, {}), err, 'throws with an arrau of false in first argument')
+  t.throws(p(false, {}), err, 'throws with an array of false in first argument')
```
evilsoft: 👍

src/index.spec.js
```diff
@@ -343,6 +345,8 @@ test('entry', t => {
 
   // predicates
   t.equal(crocks.hasProp, hasProp, 'provides the hasProp predicate')
+  t.equal(crocks.propEq, propEq, 'provides the propEq predicate')
```
evilsoft: Should be in alphabetical order

src/predicates/propPathEq.spec.js
```diff
@@ -0,0 +1,53 @@
...
+  t.throws(p({}, {}, val, {}), err, 'throws with an object in first argument')
+
+  t.throws(p([ undefined ], val, {}), err, 'throws with an array of undefined in first argument')
+  t.throws(p([ null ], val, {}), err, 'throws with array of null in first argument')
+  t.throws(p(false, val, {}), err, 'throws with an array of false in first argument')
+  t.throws(p(true, val, {}), err, 'throws with an array of true in first argument')
+  t.throws(p([ val, {} ], val, {}), err, 'throws with an array of objects in first argument')
+  t.throws(p([ [ 'key' ] ], val, {}), err, 'throws with a nested array in first argument')
+  t.throws(p([ 'some', 'key', null, 5 ], val, {}), err, 'throws with unexpected input beyond first index in first argument')
+  t.ok(p([ 'life', 'on', 'mars' ], val, {}), 'accepts an array of strings in the first argument')
```
evilsoft: These specs seem unnecessary as the working cases below cover that portion.

src/predicates/propEq.spec.js
```diff
@@ -0,0 +1,52 @@
...
+  const obj = { [key]: val }
+
+  t.throws(f(undefined, val, {}), err1, 'throws with undefined in first argument')
+  t.throws(f(null, val, {}), err1, 'throws with null in first argument')
+  t.throws(f(false, val, {}), err1, 'throws with false in first argument')
+  t.throws(f(true, val, {}), err1, 'throws with true number in first argument')
+  t.throws(f({}, val, {}), err1, 'throws with object in first argument')
+  t.throws(f([], val, {}), err1, 'throws with array in first argument')
+  t.throws(f(unit, val, {}), err1, 'throws with function in first argument')
+  t.ok(f(key, val, obj), 'accepts a string in the first argument')
```
evilsoft: These ok specs seem unnecessary as they are covered in the "good" cases below.

karthikiyengar:
Thanks for the review @evilsoft. I will fix them later today.

src/predicates/propEq.spec.js
karthikiyengar: Wow, that's pretty interesting. I'll keep that in mind.

src/predicates/propEq.spec.js
karthikiyengar: Sure, makes sense. Apologies, I'm still getting used to the stylistic changes. I'll be cleaner the next time around :-)

src/predicates/propPathEq.js
karthikiyengar: I'd have loved to take credit, but that's just `prettier` at work. I think it would be a good idea to have `prettier` set up as a pre-commit hook (it integrates with `eslint` too). I'd be happy to take that on if you think it makes sense.

src/predicates/propPathEq.js
karthikiyengar: With regards to introducing an early-exit loop, I am all down for that, and I'll be updating my PR with the modified logic.

With regards to `if (isNil(acc)) { return acc}`, I think it would be semantically better to return an `undefined` instead. This avoids the edge case with `NaN`.

src/predicates/propPathEq.spec.js
```diff
@@ -0,0 +1,53 @@
...
+
+  const obj = { some: { key: val, null: null, nan: NaN, undefined: undefined } }
+  const goodPath = [ 'some', 'key' ]
+  const badPath = [ 'this', 'is', 'really', 'bad' ]
+
+  t.equals(propPathEq(goodPath, val, obj), true, 'returns true on correct path')
+  t.equals(propPathEq([ 'some', 'null' ], null, obj), true, 'returns true when comparing to null values that are present')
+  t.equals(propPathEq([ 'some', 'nan' ], NaN, obj), true, 'returns true when comparing to NaN values that are present')
+  t.equals(propPathEq([ 'some', 'undefined' ], undefined, obj), true, 'returns true when comparing to undefined values that are present')
+  t.equals(propPathEq([ 'some', 'wrong', null, {}, [] ], undefined, { some: NaN }), true, 'returns true when comparing early-exited paths with an undefined')
```
karthikiyengar: Here's an ugly case with early-exits - if the array contains invalid inputs after the exit, we provide a result nevertheless. I'd argue that we get rid of the `isString` / `isInteger` validations altogether.

evilsoft: I agree. Good reasoning. We should also get rid of the check in `propEq` as well.
Make it so kind sir!!

src/predicates/propPathEq.js
```diff
@@ -0,0 +1,35 @@
...
+
+// propPathEq :: [String | Number] -> a -> Object -> Boolean
+function propPathEq(path, value, target) {
+  if(!isArray(path)) {
+    throw new TypeError('propPathEq: Array of strings or integers required for first argument')
+  }
+  let index = 0
+  let acc = target
+  const length = path.length
+  while (index < length) {
```
evilsoft: Any reason to not use a for loop here? it may read a little better.

evilsoft:
Looks GREAT. Have a Squirrel:
image

PR #222, opened 2018-02-18
Clean up Prod and Sum Docs

evilsoft:
## Things get missed at times
image

This PR just makes a couple corrections to the `Prod` and `Sum` Monoid docs. Both the READMEs and the Docs.


evilsoft:
image

PR #223, opened 2018-02-18
Add `Max` and `Min` Monoids to the docs

evilsoft:
## Done and doner!
image

These are the last (2) Docs that round out the Monoids!!
At this point all Monoids that ship with `crocks` are fully documented...


evilsoft:
image

PR #225, opened 2018-02-19
All Object Traversal Functions Behave the Same

evilsoft:
## Rein in for 🌽sistency
image

A lot of work on the last release was focused around some of the Object Traversal Functions provided by `crocks`. This forced me to look at some inconsistencies  between these functions and how `undefined` values are treated through out the library as a whole.

This PR brings them all inline and has them behave the same way across the board:

* The key/paths MUST be `(String | Integer)` or `[ String | Integer ]`. All functions will throw a `TypeError` if this condition is not met. This lets values like `null` and `undefined` to be rejected and avoid String coercion when traversing keys. A key of 'null' or 'undefined' will HAVE to be supplied to access keys of those names.
* Any traversable value/instance (except `NaN`) can be provided as the object of traversal.  Any non-traversable values (`undefined` and `null`) will behave as though the key/path is not defined. While `NaN` is traversable, due to its special nature, it is treated as non-traversable in this case one case. This allow things like:
```js
// safeLength :: a -> Maybe Number
const safeLength =
  prop('length')

safeLength('nice')
//=> Just 4

safeLength([ 1, 2, 3 ])
//=> Just 3

safeLength((x, y) => x + y)
//=> Just 2

safeLength(4)
//=> Nothing
```
* All values in a defined key/path will be treated as defined, except for `undefined`. When a valid key/path to an `undefined` value exists, it will be treated as not being defined. This matches the theme of `undefined` throughout the library as a whole. So these are valid:
```js
const data = { 
  a: {
    null: null,
    nan: NaN
  }
}

hasPropPath([ 'a', 'null' ], data)
//=> true

propPath([ 'a', 'nan' ], data)
//=>Just NaN
```
* All deep path traversals will use a `for` loop to allow for early exit on large path lists. It is more imperative, but will save those few extra cycles.

Also this adds the function: `hasPropPath` just to round out `hasProp`.

### DIFF
--- a/src/helpers/propPathOr.js
+++ b/src/helpers/propPathOr.js
@@ -1,31 +1,45 @@
 /** @license ISC License (c) copyright 2017 original and current authors */
 /** @author Henrique Limas */
+/** @author Ian Hofmann-Hicks */
 
 const curry = require('../core/curry')
 const isArray = require('../core/isArray')
+const isEmpty = require('../core/isEmpty')
 const isInteger = require('../core/isInteger')
-const isNil= require('../core/isNil')
+const isDefined = require('../core/isDefined')
+const isNil = require('../core/isNil')
 const isString = require('../core/isString')
 
 // propPathOr : a -> [ String | Integer ] -> b -> c
 function propPathOr(def, keys, target) {
   if(!isArray(keys)) {
-    throw new TypeError('propPathOr: Array of strings or integers required for second argument')
+    throw new TypeError(
+      'propPathOr: Array of Non-empty Strings or Integers required for second argument'
+    )
   }
 
-  const value = keys.reduce((target, key) => {
-    if (isNil(target)) {
-      return target
+  if(isNil(target)) {
+    return def
+  }
+
+  let value = target
+  for(let i = 0; i < keys.length; i++) {
+    const key = keys[i]
+
+    if(!((isString(key) && !isEmpty(key)) || isInteger(key))) {
+      throw new TypeError(
+        'propPathOr: Array of Non-empty Strings or Integers required for second argument'
+      )
     }
 
-    if(!(isString(key) || isInteger(key))) {
-      throw new TypeError('propPathOr: Array of strings or integers required for second argument')
+    value = value[key]
+
+    if(!isDefined(value)) {
+      return def
     }
+  }
 
-    return target[key]
-  }, target)
-
-  return isNil(value) ? def : value
+  return value
 }
 
 module.exports = curry(propPathOr)
--- a/src/helpers/propPathOr.js
+++ b/src/helpers/propPathOr.js
@@ -32,6 +32,10 @@ function propPathOr(def, keys, target) {
       )
     }
 
+    if(isNil(value)) {
+      return def
+    }
+
     value = value[key]
 
     if(!isDefined(value)) {
--- a/src/predicates/hasPropPath.js
+++ b/src/predicates/hasPropPath.js
@@ -17,11 +17,12 @@ function hasPropPath(keys, target) {
     )
   }
 
-  if (isNil(target)) {
+  if(isNil(target)) {
     return false
   }
 
   let value = target
+
   for(let i = 0; i < keys.length; i++) {
     const key = keys[i]
 
@@ -31,13 +32,16 @@ function hasPropPath(keys, target) {
       )
     }
 
+    if(isNil(value)) {
+      return false
+    }
+
     value = value[key]
 
     if(!isDefined(value)) {
       return false
     }
   }
-
 
   return true
 }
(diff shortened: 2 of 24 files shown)


src/helpers/propPathOr.js
```diff
@@ -1,31 +1,45 @@
...
+    throw new TypeError(
+      'propPathOr: Array of Non-empty Strings or Integers required for second argument'
+    )
   }
 
-  const value = keys.reduce((target, key) => {
-    if (isNil(target)) {
-      return target
-    }
+  if(isNil(target)) {
```
HenriqueLimas: I think we need to put that condition inside the `for`  loop too because we could have an object like `{ foo: null }` and calling the props like `['foo', 'bar']`. This will throw an error. I think it's nice to add a test for that too. IMHO The same needs to be done on `propPath` too.

src/predicates/hasPropPath.js
```diff
@@ -0,0 +1,45 @@
...
+  for(let i = 0; i < keys.length; i++) {
+    const key = keys[i]
+
+    if(!((isString(key) && !isEmpty(key)) || isInteger(key))) {
+      throw new TypeError(
+        'hasPropPath: Array of Non-empty Strings or Integers required for first argument'
+      )
+    }
+
+    value = value[key]
```
HenriqueLimas: The same as the `propPathOr` here. We need to check if the current target is not nil, if `isNil` return `false`.

src/predicates/propPathEq.js
```diff
@@ -1,24 +1,45 @@
...
-      return equals(undefined, value)
+  for(let i = 0; i < keys.length; i++) {
+    const key = keys[i]
+
+    if(!((isString(key) && !isEmpty(key)) || isInteger(key))) {
+      throw new TypeError(
+        'propPathEq: Array of Non-empty Strings or Integers required for first argument'
+      )
+    }
+
```
HenriqueLimas: Again here, check the `isNil(acc)` before assigning the new value.

src/helpers/propPathOr.js
evilsoft: Good find!

evilsoft:
@HenriqueLimas Great 👀!
Going to clean up the spec files as well tonight. Noticed that when I went to go add them in that those specs are just all over the place and hard to follow. Gonna do some 💅 on them.

src/helpers/propPathOr.js
```diff
@@ -1,31 +1,45 @@
...
+    return def
+  }
+
+  let value = target
+  for(let i = 0; i < keys.length; i++) {
+    const key = keys[i]
 
-    if(!(isString(key) || isInteger(key))) {
-      throw new TypeError('propPathOr: Array of strings or integers required for second argument')
+    if(!((isString(key) && !isEmpty(key)) || isInteger(key))) {
```
karthikiyengar: I'm not sure how much value this will add in early-exit cases (we discussed this in the `propPathEq` PR). But I guess some errors are better than no errors.

HenriqueLimas:
@evilsoft nice. What are you thinking to do? Maybe divide each test like `t('function x throw errors', ()=> { // test related to throwing errors })`

src/predicates/propEq.js
```diff
@@ -3,14 +3,27 @@
...
 const equals = require('../core/equals')
-const isObject = require('../core/isObject')
+const isDefined = require('../core/isDefined')
+const isEmpty = require('../core/isEmpty')
+const isInteger = require('../core/isInteger')
+const isNil = require('../core/isNil')
+const isString = require('../core/isString')
 
-// propEq: (String | Number) -> a -> Object -> Boolean
+// propEq: (String | Integer) -> a -> b -> Boolean
```
karthikiyengar: Apologies, totally slipped my mind how this could be used to compare other properties too :-)

evilsoft:
Exactly!!

src/helpers/propPathOr.js
evilsoft: Yep. And in the case of `propPathEq` it made sense. But as a whole, when all functions have been taken into account, it makes sense to add them. That is why I took a second to look at all of these functions as a whole.

src/predicates/propEq.js
evilsoft: No big deal at all.  The error was with `hasProp` and it was missed in review as well. There are a lot of moving parts to this lib so things like this happen. I obviously do it all the time as in the case with `hasProp`.

HenriqueLimas:
👌 perfect. That's will increase a little the LOC but it will organize a little bit more.

evilsoft:
@HenriqueLimas and @karthikiyengar made those updates to the specs and made sure that @HenriqueLimas mentioned is dealt with with `nils` in the key path and it throwing an error. I had to keep that `isNil` check outside of the for loop in order to handle cases with `nil` being passed into the data portion, but still allow all other path traversable types (pretty much everything else)

HenriqueLimas:
@evilsoft IMHO we can remove the `isNil` outside the for loop, since the first `value` will be the `target`.

evilsoft:
@HenriqueLimas 
We allow an empty arrays, which for the most part acts like identity. So if we get an empty array (for one reason or another) we will never make it into the loop. So that outside `isNil` catches that case.

evilsoft:
image

Issue #209, opened 2018-02-13
Add `applyTo` to all `Apply` ADTs


evilsoft:
This is the first step in 💯 fantasy-land integration of `ap`. Will define `applyTo` on all `Apply` types which will match the fantasy-land signature. This will be the method that will eventually map to `fantasy-land/ap`.
In order to handle proper error reporting, method will be need to take in the method name as context:

```js
// applyTo :: Apply m => m a ~> m a -> m (a -> b) -> m b
const applyTo = name => m => (...)
```

This will be provided for the following types:
- [ ] `Async`
- [ ] `Const`
- [ ] `Either`
- [ ] `IO`
- [ ] `Identity`
- [ ] `List`
- [ ] `Maybe`
- [ ] `Pair`
- [ ] `Reader`
- [ ] `Result`
- [ ] `State`
- [ ] `Unit`
- [ ] `Writer`

Initially we will not provide a point free function for this method as there is a combinator named `applyTo` that is used for applying a value to functions...We will need to do a breaking change and some function name juggling:

* mv `combinators/applyTo` -> `combinators/applyWith`
* mv `combinators/reverseApply` -> `pointfree/applyTo`
* Update `pointfree/applyTo` to take either (value, function) OR (m value, m function) where m is an Apply of the same type. It should error unless those pairs are exactly met.

PR #226, opened 2018-02-19
Remove original `applyTo` and replace with `thrush`

evilsoft:
## We'll miss ya buddy!
image

`reverseApply` was a silly name, and `applyTo` was the wrong name for the `A` combinator. As function application is not really needed and with the need to move in a direction to support `fantasy-land/ap`, it makes sense to do this work. This is of course a MAJOR breaking change, not even going to deprecate, just going for it as we are not 1.x yet.

Although this will not be the norm, in cases like this it is necessary.

This work is the first portion of preparation for [this issue](https://github.com/evilsoft/crocks/issues/209)


evilsoft:
image

PR #232, opened 2018-02-24
Updates to cleaner imports to es6

dalefrancis88:


.eslintrc.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
     "eqeqeq": [ "error", "always" ],
     "func-call-spacing": [ "error", "never" ],
     "indent": [ "error", 2 ],
-    "linebreak-style": [ "error", "unix" ],
+    "linebreak-style": 0,
```
evilsoft: Would it be possible to change this back to error. I would hate for those nasty linebreaks to sneak into the codebase.

src/combinators/index.spec.js
```diff
@@ -0,0 +1,19 @@
...
+
+const applyTo = require('./applyTo')
+const composeB = require('./composeB')
+const constant = require('./constant')
+const flip = require('./flip')
+const identity = require('./identity')
+const substitution = require('./substitution')
+
+test('combinators entry', t => {
+  t.equal(applyTo, applyTo, 'provides the T combinator (applyTo)')
```
evilsoft: The should test the file exports against the index, so something like:
```js
const index = require('.')

test('combinators entry', t => {
  t.equal(index.applyTo, applyTo, 'provides the T combinator (applyTo)')
  ...
})
```
That way we are testing the availability of the functions on the object returned from a `require('crocks/combinators')`.

Should match this pattern in all the new specs.

.eslintrc.js
dalefrancis88: This was an accident, wasn't meant to get checked in

evilsoft:
LGTM! Have a Squirrel:
image

PR #238, opened 2018-03-03
Fix up Error specs and check for crocks errors in helper, pointfree and logic functions

evilsoft:
## Some peoples functions
image

After @karthikiyengar noticed that `head` throws a JS error when passed a `null` or `undefined` and not a super 😎crocks error, I did some digging and looked like some `helper`, `pointfree` and `logic` functions did not have good specs around their error reporting.

This PR goes through and provides better specs around the functions in these categories. In addition cleaned up a lot of boolean checks with some DeMorgan 🎇. As it turns out `head` was not the only function that provided those nasty errors on `undefined` and `null`, these also had problems:
* `bimap`
* `coalesce`
* `cons`
* `swap`
* `tail`

And finally, to top it all off, I noticed a LOT of extra parens, so I added and applied the fixable `no-extra-parens` lint rule.

This is a large PR, but the changes are pretty much the same in the various files.


dalefrancis88:
I don't see any issues 👍

evilsoft:
image

Issue #233, opened 2018-02-26
Traverse/Sequence Should Take Applicative Type Rep or Apply returning Function


evilsoft:
In order to get `fantasy-land/traverse` to work and still keep the ability to traverse with inner `Apply`s, we are going to accept one or the other as the first argument to `traverse` and the only argument to `sequence`.

This will allow us to add the `fantasy-land/traverse` method to all of the applicable types:
- [x] `Array`
- [x] `Either`
- [x] `Identity`
- [x] `List`
- [x] `Maybe`
- [x] `Pair`
- [x] `Result`

The following functions will need to update their validation as well:
- [x] `pointfree/sequence`
- [x] `pointfree/traverse`

- [x] All of the above types need to also accept the `Array` constructor as a valid `Type Rep`.

evilsoft:
This functionallity has been merged and will go out with the `0.9.1` release.

PR #239, opened 2018-03-05
Allow `sequence` and `traverse` to accept an `Applicatve TypeRep`

evilsoft:
## As we continue down the path to Fantasy Land
image

This is another step in getting us Fantasy Land compliant for `fantasy-land/traverse` and `fantasy-land/ap`. This PR accepts a `crocks` based `Appicative TypeRep` or a `crocks` based `Apply` returning function as the single argument to `sequence` and the first argument to `traverse`. This is needed to account for what the FL spec calls for, but still retains the nimbleness of the `Apply` returning function.

This is the first part of [this issue](https://github.com/evilsoft/crocks/issues/233).

The reason we are locked to `crocks` based in this PR is for a couple reasons that will be addressed in subsequent PRs:
* `isApplicative` uses the `@@implements` implementation on the `TypeRep` which is not part of the FL spec. It also makes sure that `map` and `ap` are defined on instances of the `TypeRep` as you can **really** have just a pointed Functor that is not an Applicative, but still uses `of` for its pointing. This is NOT the case in the current FL spec, so the `isApplicative` function will need to be altered to check if `fantasy-land/of` or `constructor['fantasy-land/of']` exist on the data in question.
* We are using (in `List` and `Array)` the public `ap` method which is not guaranteed but types outside of crocks to behave in what we are calling the unbiased form. We will need to have `fantasy-land/ap` implemented using the current biased form. In order to get `fantasy-land/ap`, we will need to implement the public `applyTo` on all Apply types and map that to `fantasy-land/ap`. Then we can replace the `traverse` and `sequence` implementations in `List` and `Array` to always just use `fantasy-land/ap` and thus removing the :corn:flicts.

***NOTE:*** you will notice that this PR does not accept `Array` as a TypeRep. The next PR for this implementation around the issue will address that and allow one to `traverse(Array, x => [ x + 1 ])`

### DIFF
--- a/docs/src/pages/docs/crocks/Maybe.md
+++ b/docs/src/pages/docs/crocks/Maybe.md
@@ -586,14 +586,16 @@ fullName({ first: 'Lizzy' })
 #### sequence
 
 ```haskell
-Applicative f => Maybe (f a) ~> (a -> f a) -> f (Maybe a)
+Applicative TypeRep t, Apply f => Maybe (f a) ~> (t | (b -> f b)) -> f (Maybe a)
 ```
 
-When an instance of `Maybe` wraps a possible `Applicative` instance, `sequence`
-can be used to "swap" the "type sequence". `sequence` requires that the `of`,
-or `Applicative` function of the wrapped `Applicative` instance is provided for
-the case of when the `Maybe` instance is a `Nothing`. `sequence` can be derived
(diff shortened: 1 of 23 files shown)


evilsoft:
Had a super large play with this.
Going to merge it so we can get the last bit to do with Arrays as a TypeRep.
image

Issue #206, opened 2018-02-13
Add `find` function to `Maybe` namespace


evilsoft:
A `Maybe` returning `find` would be an excellent addition to the lib. So lets add one to `src/Maybe/find.js` that has the following signature:

`find :: Foldable f => ((a -> Boolean) | Pred) -> f a -> Maybe a`

The following needs to be accomplished:
* Add code:
  - [x] `src/Maybe/find.js`
* Add specs
  - [x] `src/Maybe/find.spec.js`
* Documentation
  - [x] Update ReadMe (`src/Maybe/README.md`)
  - [x] Add to Docs with description and example usage (`docs/src/pages/docs/crocks/Maybe.md`)
  - [x] Add link to Helpers Section (`docs/src/pages/docs/functions/index.md`)
* Expose new functions to public API (`src/index.js`)
  - [x] Entry for` find`
* Add specs for new public functions (`src/index.spec.js`)
  - [x] Entry for` find`

bennypowers:
I'm (tentatively) interested in trying out this implementation, if no one else has a stake in it.

dalefrancis88:
Go for it, I was going to give it a crack but I won't get to it till the weekend

dalefrancis88:
Lol I should say, no issues from me, not "go for it" :)

evilsoft:
This has been addressed and will go out with release: 0.9.1

PR #242, opened 2018-03-08
Add `find` function to `Maybe` namespace

dalefrancis88:
#206


evilsoft:
Initial 👀

src/Maybe/find.js
```diff
@@ -0,0 +1,30 @@
+/** @license ISC License (c) copyright 2017 original and current authors */
```
evilsoft: Should be `2018`

src/Maybe/find.js
```diff
@@ -0,0 +1,30 @@
...
+const accumulator = fn => (acc, cur) =>
+  !acc.found && predOrFunc(fn, cur) ? { found: true, value: cur } : acc
+
+// find :: Foldable f => ((a -> Boolean) | Pred) -> f a -> Maybe a
+function find(fn, foldable) {
+  if(!isFunction(fn)) {
+    throw new TypeError('find: Function required for first argument')
+  }
+
+  if(isNil(foldable) && !isFoldable(foldable)) {
```
evilsoft: As a `Foldable` is required...this should throw as the contract is not met.
Also `NaN`, `null` and `undefined` so the `isNil` check is redundant in this case.

src/Maybe/find.js
```diff
@@ -0,0 +1,30 @@
...
+const isFunction = require('../core/isFunction')
+const isFoldable = require('../core/isFoldable')
+
+const accumulator = fn => (acc, cur) =>
+  !acc.found && predOrFunc(fn, cur) ? { found: true, value: cur } : acc
+
+// find :: Foldable f => ((a -> Boolean) | Pred) -> f a -> Maybe a
+function find(fn, foldable) {
+  if(!isFunction(fn)) {
+    throw new TypeError('find: Function required for first argument')
```
evilsoft: specifically, either a `Pred` or a predicate function is required for the first argument. Should call that out explicitly.

src/Maybe/find.spec.js
```diff
@@ -0,0 +1,47 @@
...
+  t.ok(isFunction(find), 'is a function')
+
+  t.ok(isFunction(find(() => true, undefined).either), 'returns a Maybe for undefined value')
+  t.ok(isFunction(find(() => true, []).either), 'returns a Maybe for array value')
+  t.ok(isFunction(find(() => true, fromArray([])).either), 'returns a Maybe for List value')
+
+  t.equal(find(() => true, undefined).either(() => 'Nothing', x => x), 'Nothing', 'returns a Nothing for undefined value')
+  t.equal(find(x => x > 3, [ 1, 2, 3, 4, 5, 6 ]).either(() => 'Nothing', x => x), 4, 'returns the correct value')
+
+  t.equal(find(x => x > 3, [ 0, null, undefined, 4, 5, 6 ]).either(() => 'Nothing', x => x), 4, 'handles bad values in foldable')
```
evilsoft: No need for `either` on these specs.
can get the same result by doing something like:
```js
 t.equal(find(x => x > 3, [ 0, null, undefined, 4, 5, 6 ]).option('Nothing'), 4, 'handles bad values in foldable')
```

src/Maybe/find.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  if(isNil(foldable) && !isFoldable(foldable)) {
+    return Maybe.Nothing()
+  }
+
+  const result = foldable.reduce(accumulator(fn), { found: false, value: undefined })
```
evilsoft: no need to explicitly set `{ value: undefined }`, `{ found: false}` is equivalent.

src/Maybe/find.js
```diff
@@ -0,0 +1,30 @@
+/** @license ISC License (c) copyright 2017 original and current authors */
+/** @author Dale Francis (dalefrancis88) */
+
+const curry = require('../core/curry')
+const Maybe = require('./index')
```
evilsoft: This code is compiled, so you we can safely remove the need for name spacing and use desturcutoring like:
```js
const { Nothing, Just } = require('.')
```

src/Maybe/find.spec.js
```diff
@@ -0,0 +1,47 @@
...
+  t.throws(fn('string'), err, 'throws if foldable is truthy string')
+  t.throws(fn(false), err, 'throws if arg is false')
+  t.throws(fn(true), err, 'throws if arg is true')
+  t.throws(fn({}), err, 'throws if arg is empty POJO')
+  t.throws(fn({ hi: 'there' }), err, 'throws if arg is non-empty POJO')
+
+  t.end()
+})
+
+test('find works as expected', t => {
```
evilsoft: We should include a section to get some coverage around using a `Pred` datatype.

evilsoft:
Also we will need to expose this function to the main crocks index [here](https://github.com/evilsoft/crocks/blob/master/src/helpers/index.js#L14). Just a heads up also, will be putting up a PR after this is merged to remove the non-core/non-helpers folder functions from the index files.

This will become a problem when we start splitting out the mono-repo. and would require a dependency on `@crocksjs/Maybe` when the user installs `@crocks/helpers`. (in this case)

If the user wanted `Maybe` functions they will have to install `@crocksjs/Maybe`

src/Maybe/find.js
karthikiyengar: I think a loop and an early-exit wouldn't be possible here because this would also need to work on other Foldables, correct?

src/Maybe/find.spec.js
```diff
@@ -0,0 +1,64 @@
...
+  t.throws(fn({}), err, 'throws if foldable is empty POJO')
+  t.throws(fn({ hi: 'there' }), err, 'throws if foldable is non-empty POJO')
+
+  t.end()
+})
+
+test('find works as expected', t => {
+  t.ok(isFunction(find), 'is a function')
+
+  t.ok(isFunction(find(() => true, []).either), 'returns a Maybe for array value')
```
evilsoft: All Sum types in `crocks` provide an `either` method so this may not be a good test for.
you can use `isSameType(Maybe, find(() => true, []))`

src/Maybe/find.spec.js
evilsoft: Still need to add specs using the `Pred` datatype.

src/Maybe/find.spec.js
```diff
@@ -0,0 +1,64 @@
...
+  t.ok(isFunction(find(() => true, fromArray([])).either), 'returns a Maybe for List value')
+
+  t.equal(find(x => x > 3, [ 1, 2, 3, 4, 5, 6 ]).either(() => 'Nothing', x => x), 4, 'returns the correct value')
+
+  t.equal(find(x => x > 3, [ 0, null, undefined, 4, 5, 6 ]).either(() => 'Nothing', x => x), 4, 'handles bad values in foldable')
+  t.equal(find(x => x > 6, [ 1, 2, 3, 4, 5, 6 ]).either(() => 'Nothing', x => x), 'Nothing', 'returns nothing when value is not found')
```
evilsoft: We need to have some specs around an empty list/array.

src/Maybe/find.js
evilsoft: That is correct @karthikiyengar!

evilsoft:
Still need documentation on this PR as well.

docs/src/pages/docs/crocks/Maybe.md
```diff
@@ -1445,6 +1445,42 @@ Just('so good')
...
 ```
 
+#### find
+
+`crocks/Maybe/find`
+
+```haskell
+find :: Foldable f => ((a -> Boolean) | Pred) -> f a -> Maybe a
+```
+Used to take a foldable and a function or predicate and searchs through it
```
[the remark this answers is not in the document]
evilsoft: Also, we have a link to the `Pred` docs at the bottom, so that refrerence to `Pred` should link to:

<code>[&#96;Pred&#96;][pred]</code>

But only in this document not the README in Maybe

src/Maybe/find.spec.js
```diff
@@ -0,0 +1,72 @@
...
+  t.throws(fn({}), err, 'throws if foldable is empty POJO')
+  t.throws(fn({ hi: 'there' }), err, 'throws if foldable is non-empty POJO')
+
+  t.end()
+})
+
+test('find works with predicates', t => {
+  const largeNumber = Pred(isNumber).concat(Pred(x => x > 100))
+
+  t.equal(find(largeNumber, [ 10, '12', 150, 200, 2000 ]).option(0), 150, 'works with predicates')
```
evilsoft: All specs that are defined against the predicate function, should also be duplicated here

src/Maybe/find.spec.js
```diff
@@ -0,0 +1,96 @@
...
+  t.throws(fn(true), err, 'throws if foldable is true')
+  t.throws(fn({}), err, 'throws if foldable is empty POJO')
+  t.throws(fn({ hi: 'there' }), err, 'throws if foldable is non-empty POJO')
+
+  t.end()
+})
+
+test('find works with predicates', t => {
+  const largeNumber = Pred(isNumber).concat(Pred(x => x > 100))
+  const nonBoolPred =
```
evilsoft: I need to find a linting rule or write one, or at least put together a style guide for whitespace. But the general guidelines should be:
1. No *more* than 1 line of white space between expression/statement groupings.
1. block content, closed or indented will be surrounded with 1 line of whitespace.
1. single line arrow function assignment should be split (like you have here) between (2) lines at a minimum, with the assignment on the first and the body indented on the following lines.

src/Maybe/find.spec.js
```diff
@@ -0,0 +1,96 @@
...
+  t.ok(isSameType(Maybe, findWrapPred(() => true, fromArray([]))), 'returns a Maybe for List value')
+
+  t.equal(findWrapPred(x => x > 3, [ 1, 2, 3, 4, 5, 6 ]).option('Nothing'), 4, 'returns the correct value')
+
+  t.equal(findWrapPred(x => x > 3, [ 0, null, undefined, 4, 5, 6 ]).option('Nothing'), 4, 'handles bad values in foldable')
+  t.equal(findWrapPred(x => x > 6, [ 1, 2, 3, 4, 5, 6 ]).option('Nothing'), 'Nothing', 'returns nothing when value is not found')
+  t.equal(findWrapPred(() => true, []).option('Nothing'), 'Nothing', 'returns a Nothing when there are no items ([])')
+  t.equal(findWrapPred(() => true, new Array()).option('Nothing'), 'Nothing', 'returns a Nothing when there are no items (Array)')
+  t.equal(findWrapPred(() => true, fromArray([])).option('Nothing'), 'Nothing', 'returns a Nothing when there are no items (List)')
+
```
evilsoft: extra new line here.

src/Maybe/find.spec.js
```diff
@@ -0,0 +1,96 @@
...
+test('find works with predicates', t => {
+  const largeNumber = Pred(isNumber).concat(Pred(x => x > 100))
+  const nonBoolPred =
+    x => Pred(constant(x))
+  const findWrapPred =
+    (fn, foldable) => find(Pred(fn), foldable)
+
+  t.equal(find(largeNumber, [ 10, '12', 150, 200, 2000 ]).option(0), 150, 'works with predicates')
+
+  t.equals(find(nonBoolPred(undefined), []).option('Nothing'), 'Nothing', 'returns false when wrapped function returns undefined')
```
evilsoft: This is an interesting test...Do not know if we need to test this, as we would get the same results if we returned `true`, or any value on an empty Array/List. This is more a spec for an empty list.

And because of the truthy falsey bits, you could write a whole grouping of tests on JS values like

```js
t.equals(find(nonBoolPred(0), [ 3, 4 ]).option('Nothing'), 'Nothing', 'returns Nothing with falsey number')
t.equals(find(nonBoolPred(1), [ 3, 4 ]).option('Nothing'), 3, 'returns first value in a Just with truthy number')
```

But really do not think we need this test, or all of the ones I described above, as the specs below cover the context of this function, and the specs for `predOrFunc` cover the truthy/falsey values...So I think we have enough coverage to rule this test out.

Thoughts?

src/Maybe/find.spec.js
```diff
@@ -0,0 +1,96 @@
...
+
+  t.end()
+})
+
+test('find works with predicates', t => {
+  const largeNumber = Pred(isNumber).concat(Pred(x => x > 100))
+  const nonBoolPred =
+    x => Pred(constant(x))
+  const findWrapPred =
+    (fn, foldable) => find(Pred(fn), foldable)
```
evilsoft: You should probably take advantage of the currying with `find` here by defining this using something like:
```js
const wrapPred = fn =>
  find(Pred(fn))

const gt6 =
  wrapPred(x => x > 6)

const T =
  wrapPred(constant(true))
```

This way we get tests around the curried interface as well, and shorten up some of the spec defs.

src/Maybe/find.spec.js
dalefrancis88: Yeah i think it can be removed

src/Maybe/README.md
```diff
@@ -1417,3 +1417,36 @@ Just('so good')
...
 ```
+
+#### find
+
+`crocks/Maybe/find`
+
+```haskell
+find :: Foldable f => ((a -> Boolean) | Pred) -> f a -> Maybe a
+```
+Using a provided predicate function or a `Pred` datatype, `find` takes a `Foldable` instance and  executes for every value in the `Foldable`, skippy empty indexs. `find` then returns the first value it finds that passes the predicate. If found, `find` returns the value in a `Just`, otherwise a `Nothing` is returned.
```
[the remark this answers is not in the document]
evilsoft: Oh, then please keep it in. I like bits of personality sneaking in

evilsoft:
image

Super :sweet_potato:...Merging.

PR #243, opened 2018-03-10
Allow Array Constructor as TypeRep for sequence and traverse

evilsoft:
## Represent!!
image

This PR allows using the Array 🌽structor for traverse and sequence by using a new internal helper to determine if a constructor matches a target. Well really it is checking if the name is the same. As this is an internal function, it should be just fine, as we will be providing the parameters.

This is the final bit of functionality needed to complete out [this issue](https://github.com/evilsoft/crocks/issues/233).

### DIFF
--- a/src/Identity/Identity.spec.js
+++ b/src/Identity/Identity.spec.js
@@ -429,6 +429,12 @@ test('Identity sequence with Applicative TypeRep', t => {
   t.ok(isSameType(Identity, m.valueOf()), 'Provides an inner type of Identity')
   t.equal(m.valueOf().valueOf(), x, 'Identity contains original inner value')
 
+  const arM = Identity([ x ]).sequence(Array)
+
+  t.ok(isSameType(Array, arM), 'Provides an outer type of Array')
+  t.ok(isSameType(Identity, arM[0]), 'Provides an inner type of Identity')
+  t.equal(arM[0].valueOf(), x, 'Identity contains original inner value')
+
   t.end()
 })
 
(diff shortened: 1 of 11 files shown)


src/core/apOrFunc.js
```diff
@@ -2,8 +2,11 @@
 /** @author Ian Hofmann-Hicks (evil) */
 
 const isApplicative = require('./isApplicative')
+const isTypeRepOf = require('./isTypeRepOf')
 
 const apOrFunc = af => x =>
```
dalefrancis88: Is there a value in separating out the parameters and not using curry? Should we set a standard for it?

src/Identity/Identity.spec.js
```diff
@@ -501,13 +507,20 @@ test('Identity traverse with Apply function', t => {
...
 
+  const ar = x => [ x ]
+  const arM = Identity(x).traverse(Array, fn(ar))
+
+  t.ok(isSameType(Array, arM), 'Provides an outer type of Array')
+  t.ok(isSameType(Identity, arM[0]), 'Provides an inner type of Identity')
+  t.equal(arM[0].valueOf(), res, 'Identity contains transformed value')
+
   t.end()
 })
```
dalefrancis88: No final empty line?

src/Result/Result.spec.js
```diff
@@ -1030,5 +1041,17 @@ test('Result traverse with Applicative TypeRep', t => {
...
+  t.ok(isSameType(Array, arO), 'Provides an outer type of Array')
+  t.ok(isSameType(Result, arO[0]), 'Provides an inner type of Result')
+  t.equal(arO[0].either(constant(0), identity), res, 'Result contains original inner value')
+
+  t.ok(isSameType(Array, arE), 'Provides an outer type of Array')
+  t.ok(isSameType(Result, arE[0]), 'Provides an inner type of Result')
+  t.equal(arE[0].either(identity, constant(0)), 'Err', 'Result contains original Err value')
+
   t.end()
 })
```
dalefrancis88: No final empty line

src/core/List.spec.js
```diff
@@ -755,13 +761,20 @@ test('List traverse with Apply function', t => {
...
 
+  const ar = x => [ x ]
+  const arM = List.of(x).traverse(Array, fn(ar))
+
+  t.ok(isSameType(Array, arM), 'Provides an outer type of Array')
+  t.ok(isSameType(List, arM[0]), 'Provides an inner type of List')
+  t.same(arM[0].valueOf(), [ res ], 'inner List contains transformed value')
+
   t.end()
 })
```
dalefrancis88: missing again... this seems deliberate, do we want final new lines still?

src/core/apOrFunc.js
evilsoft: It is curried because of how it is used in context, if it was not curried it would look like this:
```js
return either(
  compose(x => apOrFunc(f, x), Either.Left),
  runSequence
)
```

instead of just:
```js
const af =
  apOrFunc(f)

return either(
  compose(af, Either.Left),
  runSequence
)
```
and because the SumTypes and List are the only things that really use that first argument, they will always be in some sort of composition on the Left (or empty List/Array) side. So IMO it makes sense to write to the expected interface.

dalefrancis88:
image

PR #244, opened 2018-03-10
Added new function to check for isPredOrFunc

dalefrancis88:


src/core/isPredOrFunc.spec.js
```diff
@@ -0,0 +1,30 @@
+const test = require('tape')
+
+const Pred = require('../Pred')
+const isNumber = require('./isNumber')
+const unit = require('./_unit')
+const isLargeNumber =
```
evilsoft: insert a new line above "block"ish bits of text please. Also in test 🌽structs should be places after all requires.

src/core/isPredOrFunc.js
```diff
@@ -0,0 +1,12 @@
+/** @license ISC License (c) copyright 2018 original and current authors */
+/** @author Dale Francis (dalefrancis88) */
+
+const Pred = require('./types').proxy('Pred')
+const isFunction = require('./isFunction')
+const isSameType = require('./isSameType')
+
+// isPredOrFunc :: (func | pred) -> bool
```
evilsoft: should read like:
```js
// isPredOrFunc :: ((a -> b) | Pred a) -> Boolean
```

src/logic/and.js
```diff
@@ -1,16 +1,13 @@
-/** @license ISC License (c) copyright 2017 original and current authors */
+/** @license ISC License (c) copyright 2018 original and current authors */
```
evilsoft: This was added in 2017, should keep the original year

dalefrancis88: This is reverted now

evilsoft:
image

Issue #245, opened 2018-03-11
Add support for Microsoft Windows development environments


dalefrancis88:
Currently all commands are written with the expectation of a shell/unix based environment. As much as is possible, support needs to be added for Windows based environments.

This support can ignore publishing as there is no expected need to publish from a windows based environment.

evilsoft:
This has been merged and is available to developers on the `master` branch.
Closing.

PR #246, opened 2018-03-11
Added pre/post install, updated build, fixed spec:dev, added lint watch

dalefrancis88:
#245


dalefrancis88:
I added eslint-watch to the project, it helps while i develop. I generally have two terminals open, one watching tests and one watching linting. 

I wasn't to sure about changeing the packages.json for the docs but from the research i could find on electric, its a one off generator so editing packages json seemed fine. 

Both of these can be taken out if it's not liked.

I also did some looking into tape-watch and tape-watcher, both seemed good, but neither supported tap-spec so i didn't put it in. Something interesting to note though, tape-watcher only displays failed tests and also can run the tests in ~250ms

evilsoft:
@dalefrancis88 looks like you need to use npm 5.6.0 and node 8 locally to do your install so that the package-lock.json is updated

evilsoft:
Also, this is going to be a problem on initial installs as rimraf is not going to be availible to remove the folder. So the script will fail.

dalefrancis88:
I guess the most interesting thing on this is the rm -rf node_modules, its the part that we're struggling with. The thing is, this is quite an opinionated move. We're choosing for people to delete the node_modules folder. 
The guidance is that this is what to do on a first install, but if thats the case the folders won't exist. 

what do you think of just putting in the `"postinstall": "cd docs && npm i"` and not worrying about deleting the node_modules folders? If people want to do it they can choose to delete themselves

dalefrancis88:
Added a new script that should replicate what your bash script was doing. LMK if it works with your environment

bin/setup.js
```diff
@@ -0,0 +1,32 @@
...
+var cp = require('child_process')
+var os = require('os')
+
+var root = resolve(__dirname, '..')
+var docs = resolve(__dirname, '..', 'docs')
+var isWindows = os.platform().startsWith('win')
+var npmCmd = isWindows ? 'npm.cmd' : 'npm'
+var rmrf = isWindows ? 'if exist "node_modules" (rd /s /q node_modules)' : 'rm -rf node_modules'
+
+cp.execSync('cls', { env: process.env, stdio: 'inherit' })
```
evilsoft: This is not a GNU Tool... We use `clear` but I do not see a need for clearing the console.

bin/setup
```diff
@@ -1,5 +0,0 @@
-#!/usr/bin/env bash
-
-rm -rf node_modules && npm i --no-save
```
evilsoft: In order to keep the lock files in check between different Operating Systems, we will need to include the `--no-save` option

bin/setup.js
```diff
@@ -0,0 +1,32 @@
...
+
+install(root)
+install(docs)
+
+function install(dir) {
+  console.log('working on ' + dir)
+
+  if (!fs.existsSync(join(dir, 'package.json'))) { return }
```
evilsoft: Why this return?

dalefrancis88: because i copied from StackOverflow that was originally for scanning folders. This was there originally as check for something else but rather than delete it i thought it was a nice check just incase the wrong folder or empty folder was passed in. 
Although as i write this i realise that it should console log if thats the case.

bin/setup.js
dalefrancis88: Probably don't, this is more of a personal habit. Will take it out

evilsoft:
Great Job. This will allow windows people to work on this lib without feeling like outsiders.

Have a Windows squirrel:
image

PR #248, opened 2018-03-21
Adding new md for Const

dalefrancis88:


docs/src/pages/docs/crocks/Const.md
```diff
@@ -0,0 +1,100 @@
...
+## Instance Methods
+
+#### concat
+
+```haskell
+Min ~> Min -> Min
+```
+
+`concat` is used to combine (2) `Semigroup`s of the same type under an
+operation specified by the `Semigroup`. In the case of `Min`, it will result
```
bennypowers: Here, `Min` is introduced for the first time as an aside. "In the case of `Min`"... As the reader, I didn't know `Min` was a topic until it already became a case :D Perhaps a brief introduction is in order, since the document discusses `Const`, whereas this whole section doesn't mention `Const` at all.

docs/src/pages/docs/crocks/Const.md
```diff
@@ -0,0 +1,100 @@
...
+//=> Min -Infinity
+```
+
+#### valueOf
+
+```haskell
+Min ~> () -> Number
+```
+
+`valueOf` is used on all `crocks` `Monoid`s as a means of extraction. While the
```
bennypowers: ambiguous use of "while" - do you mean "whenever" or "even though"?

docs/src/pages/docs/crocks/Const.md
```diff
@@ -0,0 +1,100 @@
...
+
+```javascript
+import Const from 'crocks/Const'
+
+Const('Hello World')
+//=> Const 'Hello World'
+
+([ 'Today', 'Tomorrow', 'Yesterday' ])
+   .map(Const)
+   .reduce ((acc, c) => acc.concat(c))
```
karthikiyengar: I think the code samples here do not follow lint rules (space after reduce)?

docs/src/pages/docs/crocks/Const.md
```diff
@@ -0,0 +1,100 @@
...
+//=> Const 'Today'
+
+Const(100)
+  .concat(Const(10))
+//=> Min 100
```
karthikiyengar: This would be a `Const`

docs/src/pages/docs/crocks/Const.md
```diff
@@ -0,0 +1,100 @@
...
+<article id="topic-implements">
+
+## Implements
+
+`ap`, `chain`, `concat`, `equals`, `map`
+</article>
+
+<article id="topic-instance">
+
+## Instance Methods
```
karthikiyengar: The instance methods seem unreleated to `Const` - maybe I am missing something?

karthikiyengar:
Apologies, I missed the discussion on the channel.

src/Const/README.md
```diff
@@ -0,0 +1,158 @@
+---
+title: "Const"
+description: "Const Monoid"
+layout: "guide"
+weight: 80
+---
+
+```haskell
+Const c _
```
evilsoft: Should still be `Const c a` even though we ignore that.

src/Const/README.md
```diff
@@ -0,0 +1,158 @@
...
+```
+`Const` a unary function which creates a simple Product type that evaluates to
+`c` for all inputs ignoring its right side. 
+`Const` is a `chainable` that will take any value `c` and regardless of the 
+function called on it will return a new `Const` with the value `c`.
+
+### Contrived Examples
```
evilsoft: Do not think we want to call out Contrived vs. Real world

src/Const/README.md
```diff
@@ -0,0 +1,158 @@
...
+```javascript
+import Const from 'crocks/Const'
+
+Const('Hello World')
+//=> Const 'Hello World'
+
+const days = [ 'Today', 'Tomorrow', 'Yesterday' ]
+
+days.map(Const).reduce((acc, c) => acc.concat(c))
```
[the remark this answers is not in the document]
evilsoft: Nope! this is all lies. Sorry about that cannot use `mconcat` here. `Const` is not a `Monoid`

src/Const/README.md
```diff
@@ -0,0 +1,158 @@
...
+```javascript
+// TBD - methods are " 'ap', 'chain', 'concat', 'equals', 'map'  "
+```
+</article>
+
+<article id="topic-instance">
+
+## Instance Methods
+
+#### ap
```
evilsoft: Also these should be in  "less constraint" order.
`equals` -> `concat` ->`map` -> `ap` -> `chain`

src/Const/README.md
```diff
@@ -0,0 +1,158 @@
...
+  compose(Const, prod(2))
+
+doubleAndFreeze(7)
+//=> Const 14
+
+Const('initial')
+  .map(x => x + x)
+  .chain(doubleAndFreeze)
+//=> Const 'initial'
+
```
evilsoft: Extra Newline

src/Const/README.md
```diff
@@ -0,0 +1,158 @@
...
+```haskell
+Const ~> Const -> Const
+```
+
+```javascript
+```
+
+#### equals
+
+```haskell
```
evilsoft: For future you, this siggy should be:
```haskell
Const c a ~> b -> Boolean
```
(it can take anything `b`)

src/Const/README.md
```diff
@@ -0,0 +1,158 @@
...
+Const ~> Const -> Boolean
+```
+
+```javascript
+```
+
+#### map
+
+```haskell
+Const ~> (a -> b) -> Const
```
evilsoft: For future you, this siggy should be:
```haskell
Const c a ~> (a -> b) -> Const c b
```

dalefrancis88:
Thanks for the sigs @evilsoft :) i am still learning

src/Const/README.md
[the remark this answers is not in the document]
dalefrancis88: hahaha yeah i started with mconcat and had the same problem :) Thats actually what triggered the whole lot of questions around monoids and what type this was

src/Const/README.md
```diff
@@ -0,0 +1,158 @@
...
+//=> Const 'Hello World'
+
+const days = [ 'Today', 'Tomorrow', 'Yesterday' ]
+
+days.map(Const).reduce((acc, c) => acc.concat(c))
+//=> Const 'Today'
+
+Const(100)
+  .concat(Const(10))
+//=> Min 100
```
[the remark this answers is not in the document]
dalefrancis88: fail :)

bennypowers: Perhaps discuss `Min` as an instance of `Const` before introducing it in `Contrived Examples`

src/Const/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+Const('Hello')
+  .map(x => x + x)
+//=> Const 'Hello'
+```
+
+### Real-world example
+```javascript
```
bennypowers: You might consider describing in english what this data flow does. i.e. `Update field allows us to store an updated value in the second slot of a Pair, and resetField allows us to reset that second slot to the constant value in the first`

src/Const/README.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+#### chain
+
+```haskell
+Const a ~> (a -> Const b) -> Const a
+```
+
+Combining a sequential series of transformations that capture disjunction can 
+beaccomplished with `chain`. `chain` expects a unary, `Const` returning
```
bennypowers: be accomplished

src/Const/README.md
```diff
@@ -0,0 +1,158 @@
...
+the function will remain the active value in the `Const`.
+
+```javascript
+import { Const } from 'crocks'
+
+// prod :: Number -> Number -> Number
+const prod = x => y => x * y
+
+// Const -> Const
+Const(prod)
```
bennypowers: a short discussion of cases where we want this behaviour might be in order

src/Const/README.md
```diff
@@ -0,0 +1,158 @@
...
+const doubleAndFreeze =
+  compose(Const, prod(2))
+
+doubleAndFreeze(7)
+//=> Const 14
+
+Const('initial')
+  .map(x => x + x)
+  .chain(doubleAndFreeze)
+//=> Const 'initial'
```
bennypowers: is `doubleAndFreeze` a red herring on line 126? really any function you pass here would return a Const of 'initial', jsut because of the special nature of Const, not because of chain, amirite?

src/Const/README.md
dalefrancis88: Sorry Min is a typo

src/Const/README.md
dalefrancis88: This was more of an example to show that multiple methods can be called but you're right it's pretty useless. I am going to change the example anyway.

src/Const/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+```haskell
+Const c a
+```
+`Const` a unary function which creates a simple Product type that evaluates to
+`c` for all inputs ignoring its right side. 
+`Const` is a `chainable` that will take any value `c` and regardless of the 
+function called on it will return a new `Const` with the value `c`.
+
+### Example
```
evilsoft: I would say that this example should not be included. This is more a case for `First`, for the `mconcat` bits.
We may wat to take the last two samples in this example and include them in the second. Consolidating the (2). Have one example here also matches the rest of the docs.

dalefrancis88: I was going to move the second example to be the example for chain, thoughts?

src/Const/README.md
```diff
@@ -0,0 +1,215 @@
...
+Const c a ~> (a -> Const c b) -> Const c b
+```
+
+Combining a sequential series of transformations that capture disjunction can 
+be accomplished with `chain`. `chain` expects a unary, `Const` returning 
+function as its argument. When invoked the inner value will not be passed to 
+provided function. A new `Const` will be returned with the same inner value.
+
+```javascript
+import Const from 'crocks/Const'
```
evilsoft: I am thinking this example should merge with the first example, keep the first just append this on the end.
Then this chain can be similar to the `map` function, just for `chain`.

src/Const/README.md
```diff
@@ -0,0 +1,217 @@
...
+will be validated but it will not be evaluated. `map` will return a new `const`
+with the same left value.
+
+```javascript
+import { Const } from 'crocks'
+
+Const('initial')
+  .map(x => x.fakeProperty)
+//=> Const 'initial'
+
```
evilsoft: Extra new line here

src/Const/README.md
```diff
@@ -0,0 +1,217 @@
...
+Const c (a -> b) ~> Const c a -> Const c b
+```
+
+Short for apply, `ap` is normally used to apply a `Const` instance containing a 
+value to another `Const` instance that contains a function, resulting in new 
+`Const` instance with the result. However, due to the unique nature of `Const`
+the function will remain the active value in the `Const`.
+
+```javascript
+import { Const } from 'crocks'
```
evilsoft: need to fix the imports to point at `crocks/Const`
(so technically this *will* work for things like Rollup and Babel. but when we move to `esm` we will have to change all of these to things like `crocks/esm/Const/index.mjs`, but for now we will keep how we have been doing it)

src/Const/README.md
```diff
@@ -0,0 +1,217 @@
...
+function as its argument. When invoked the inner value will not be passed to 
+provided function. A new `Const` will be returned with the same inner value.
+
+```javascript
+import { Const } from 'crocks'
+
+Const('initial')
+  .chain(x => Const(x.fakeProperty))
+//=> Const 'initial'
+
```
evilsoft: Extra newline

src/Const/README.md
```diff
@@ -0,0 +1,217 @@
...
+Const c a ~> () -> c
+```
+
+`valueOf` is used as a means of extraction. This function
+is used primarily for convenience for some of the helper functions that ship
+with `crocks`. Calling `valueOf` on a `Const` instance will result in the
+underlying left value of the `Product` type.
+
+```javascript
+import { Const } from 'crocks'
```
evilsoft: need to fix the imports to point at `crocks/Const`

src/Const/README.md
```diff
@@ -0,0 +1,217 @@
...
+  updateField('Joey')
+    .map(toLower)
+    .chain(updateField)
+//=> Pair( Const "Joey", "joey" )
+
+resetField(changed)
+//=> Pair( Const "Joey", "Joey" )
+
+```
+<article id="topic-implements">
```
evilsoft: I was going to say that these are only for the `docs`...but it does not hurt to have them here...So nm. I will take a pass over the READMES to add these, makes it easier to maintain, the more that these two area have in common.

docs/src/pages/docs/crocks/Const.md
```diff
@@ -0,0 +1,217 @@
...
+title: "Const"
+description: "Const Monoid"
+layout: "guide"
+weight: 80
+---
+
+```haskell
+Const c a
+```
+`Const` is a function that takes a single value and returns a `Product` type
```
evilsoft: I would say something like:
> `Const` is a Product type the whose underlying value is fixed to the value it was originally constructed with. This allows to ensure that a desired value is immutable.

evilsoft:
@dalefrancis88 and @karthikiyengar @bennypowers 

To get this going, made some updates and addressed some of the previous PR feedback.
Please if you can give some additional 👀 over the changes please!

dalefrancis88:
Hey looks good, sorry i didn't get this done, but glad its done now :)

Issue #230, opened 2018-02-23
Add `Setoid` to most of the `Monoid`


evilsoft:
## Merging Equality
![image](https://user-images.githubusercontent.com/3665793/36575650-ec69abf6-1800-11e8-92e4-a0dc873dd946.png)

As brought up by @karthikiyengar, there is no reason for a Majority of the `Monoid`s to also be `Setoid`s. The following should be implemented:

- [x] `All`
- [x] `Any`
- [x] `Assign`
- [x] `First`
- [x] `Last`
- [x] `Max`
- [x] `Min`
- [x] `Prod`
- [x] `Sum`

Each will need to have the following additions:
* Add `equals` method to the instance
* Add the `fantasy-land/equals` to the instance
* Add `equals`  to `@@implements` function on the constructors
* specs for all.

evilsoft:
This is shipped and GTG

PR #250, opened 2018-03-28
Add equals to most Monoids (#230)

karthikiyengar:
This PR makes most of the monoids setoids as well, by implementing the `equals` function.

### DIFF
--- a/docs/src/pages/docs/monoids/First.md
+++ b/docs/src/pages/docs/monoids/First.md
@@ -138,6 +138,40 @@ concat(concat(a, b), c)
 //=> First( Just "c" )
 ```
 
+#### equals
+
+```haskell
+First a ~> b -> Boolean
+```
+
+Used to compare the underlying values of (2) `First` instances for equality by value, equals takes any given argument and returns `true` if the passed argument is a `First` with an underlying value equal to the underlying value of the `First` the method is being called on. If the passed argument is not a `First` or the underlying values are not equal, equals will return `false`.
+
+```javascript
+import First from 'crocks/First'
+
+import Maybe from 'crocks/Maybe'
+
+First(33)
+  .equals(First(33))
+//=> true
+
+First(42)
+  .equals(First(10))
+//=> false
+
+First({ a: 5 })
+  .equals({ a: 5 })
+//=> false
+
+First(95)
+  .equals(95)
+//=> false
+
+First(95)
+  .equals(Maybe.of(95))
+//=> false
+```
+
 #### option
 
 ```haskell
--- a/docs/src/pages/docs/monoids/Last.md
+++ b/docs/src/pages/docs/monoids/Last.md
@@ -136,6 +136,40 @@ concat(concat(a, b), c)
 //=> Last( Just "a" )
 ```
 
+#### equals
+
+```haskell
+Last a ~> b -> Boolean
+```
+
+Used to compare the underlying values of (2) `Last` instances for equality by value, equals takes any given argument and returns `true` if the passed argument is a `Last` with an underlying value equal to the underlying value of the `Last` the method is being called on. If the passed argument is not a `Last` or the underlying values are not equal, equals will return `false`.
+
+```javascript
+import Last from 'crocks/Last'
+
+import Maybe from 'crocks/Maybe'
+
+Last(33)
+  .equals(Last(33))
+//=> true
+
+Last(42)
+  .equals(Last(10))
+//=> false
+
+Last({ a: 5 })
+  .equals({ a: 5 })
+//=> false
+
+Last(95)
+  .equals(95)
+//=> false
+
+Last(95)
+  .equals(Maybe.of(95))
+//=> false
+```
+
 #### option
 
 ```haskell
--- a/src/All/index.js
+++ b/src/All/index.js
@@ -5,6 +5,7 @@ const VERSION = 1
 
 const _implements = require('../core/implements')
 const _inspect = require('../core/inspect')
+const _equals = require('../core/equals')
 const type = require('../core/types').type('All')
 const _type = require('../core/types').typeFn(type(), VERSION)
 const fl = require('../core/flNames')
@@ -29,6 +30,10 @@ function All(b) {
   const empty =
     _empty
 
+  const equals =
+    m => isSameType(All, m)
+      && _equals(x, m.valueOf())
+
   const inspect =
     () => `All${_inspect(valueOf())}`
 
@@ -42,8 +47,9 @@ function All(b) {
 
   return {
     inspect, toString: inspect,
-    valueOf, type, concat, empty,
+    valueOf, type, concat, empty, equals,
     ['@@type']: _type,
+    [fl.equals]: equals,
     [fl.concat]: concat,
     [fl.empty]: empty,
     constructor: All
@@ -51,7 +57,7 @@ function All(b) {
 }
 
 All['@@implements'] = _implements(
-  [ 'concat', 'empty' ]
+  [ 'concat', 'empty', 'equals' ]
 )
 
 All.empty = _empty
--- a/src/All/index.js
+++ b/src/All/index.js
@@ -46,8 +46,8 @@ function All(b) {
   }
 
   return {
-    inspect, toString: inspect,
-    valueOf, type, concat, empty, equals,
+    inspect, toString: inspect, valueOf,
+    equals, type, concat, empty,
     ['@@type']: _type,
     [fl.equals]: equals,
     [fl.concat]: concat,
@@ -57,7 +57,7 @@ function All(b) {
 }
 
 All['@@implements'] = _implements(
-  [ 'concat', 'empty', 'equals' ]
+  [ 'equals', 'concat', 'empty' ]
 )
 
 All.empty = _empty
--- a/src/Any/index.js
+++ b/src/Any/index.js
@@ -5,6 +5,7 @@ const VERSION = 1
 
 const _implements = require('../core/implements')
 const _inspect = require('../core/inspect')
+const _equals = require('../core/equals')
 const type = require('../core/types').type('Any')
 const _type = require('../core/types').typeFn(type(), VERSION)
 const fl = require('../core/flNames')
@@ -32,6 +33,10 @@ function Any(b) {
   const inspect =
     () => `Any${_inspect(valueOf())}`
 
+  const equals =
+    m => isSameType(Any, m)
+      && _equals(x, m.valueOf())
+
   function concat(m) {
     if(!isSameType(Any, m)) {
       throw new TypeError('Any.concat: Any required')
@@ -41,17 +46,18 @@ function Any(b) {
   }
 
   return {
-    inspect, toString: inspect,
+    inspect, toString: inspect, equals,
     valueOf, type, concat, empty,
     ['@@type']: _type,
     [fl.concat]: concat,
     [fl.empty]: empty,
+    [fl.equals]: equals,
     constructor: Any
   }
 }
 
 Any['@@implements'] = _implements(
-  [ 'concat', 'empty' ]
+  [ 'concat', 'empty', 'equals' ]
 )
 
 Any.empty = _empty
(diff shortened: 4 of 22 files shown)


bennypowers:
No docs updates?

src/Assign/Assign.spec.js
```diff
@@ -176,3 +193,19 @@ test('Assign empty functionality', t => {
 
   t.end()
 })
+
+test('Assign equals functionality', t => {
```
evilsoft: This block should be moved up and grouped with the Setoid validation.

src/First/First.spec.js
```diff
@@ -180,3 +197,19 @@ test('First empty properties (Monoid)', t => {
 
   t.end()
 })
+
+test('First equals functionality', t => {
```
evilsoft: This block should be moved up and grouped with the Setoid validation.

src/Last/Last.spec.js
```diff
@@ -178,3 +196,19 @@ test('Last empty properties (Monoid)', t => {
 
   t.end()
 })
+
+test('Last equals functionality', t => {
```
evilsoft: This block should be moved up and grouped with the Setoid validation.

src/Max/Max.spec.js
```diff
@@ -177,3 +195,19 @@ test('Max empty functionality', t => {
 
   t.end()
 })
+
+test('Max equals functionality', t => {
```
evilsoft: This block should be moved up and grouped with the Setoid validation.

src/Min/Min.spec.js
```diff
@@ -177,3 +195,19 @@ test('Min empty functionality', t => {
 
   t.end()
 })
+
+test('Min equals functionality', t => {
```
evilsoft: This block should be moved up and grouped with the Setoid validation.

src/Prod/Prod.spec.js
```diff
@@ -177,3 +195,19 @@ test('Prod empty functionality', t => {
 
   t.end()
 })
+
+test('Prod equals functionality', t => {
```
evilsoft: This block should be moved up and grouped with the Setoid validation.

src/Sum/Sum.spec.js
```diff
@@ -177,3 +195,19 @@ test('Sum empty functionality', t => {
 
   t.end()
 })
+
+test('Sum equals functionality', t => {
```
evilsoft: This block should be moved up and grouped with the Setoid validation.

src/Any/index.js
```diff
@@ -32,6 +33,10 @@ function Any(b) {
   const inspect =
     () => `Any${_inspect(valueOf())}`
 
+  const equals =
+    m => isSameType(Any, m)
```
dalefrancis88: Could this be made into a `isEqual` function and shared across each of these?

src/All/index.js
```diff
@@ -29,6 +30,10 @@ function All(b) {
   const empty =
     _empty
 
+  const equals =
+    m => isSameType(All, m)
```
dalefrancis88: Could this be made into a `isEqual` function and shared across each of these?

src/Assign/index.js
```diff
@@ -34,6 +35,10 @@ function Assign(o) {
   const inspect =
     () => `Assign${_inspect(valueOf())}`
 
+  const equals =
+    m => isSameType(Assign, m)
```
dalefrancis88: Could this be made into a `isEqual` function and shared across each of these?

src/First/index.js
```diff
@@ -30,6 +31,10 @@ function First(x) {
   const inspect =
     () => `First(${_inspect(maybe)} )`
 
+  const equals =
+  m => isSameType(First, m)
```
dalefrancis88: Could this be made into a `isEqual` function and shared across each of these?

src/Last/index.js
```diff
@@ -33,6 +34,10 @@ function Last(x) {
   const inspect =
     () => `Last(${_inspect(maybe)} )`
 
+  const equals =
+  m => isSameType(Last, m)
```
dalefrancis88: Could this be made into a `isEqual` function and shared across each of these?

src/Max/index.js
```diff
@@ -32,6 +33,10 @@ function Max(n) {
   const inspect =
     () => `Max${_inspect(valueOf())}`
 
+  const equals =
+    m => isSameType(Max, m)
```
dalefrancis88: Could this be made into a `isEqual` function and shared across each of these?

src/Min/index.js
```diff
@@ -32,6 +33,10 @@ function Min(n) {
   const inspect =
     () => `Min${_inspect(valueOf())}`
 
+  const equals =
+    m => isSameType(Min, m)
```
dalefrancis88: Could this be made into a `isEqual` function and shared across each of these?

src/Prod/index.js
```diff
@@ -32,6 +33,10 @@ function Prod(n) {
   const inspect =
     () => `Prod${_inspect(valueOf())}`
 
+  const equals =
+    m => isSameType(Prod, m)
```
dalefrancis88: Could this be made into a `isEqual` function and shared across each of these?

src/Sum/index.js
```diff
@@ -32,6 +33,10 @@ function Sum(n) {
   const inspect =
     () => `Sum${_inspect(valueOf())}`
 
+  const equals =
+    m => isSameType(Sum, m)
```
dalefrancis88: Could this be made into a `isEqual` function and shared across each of these?

evilsoft:
@karthikiyengar: if you wanna take @dalefrancis88 suggestion, you could make a helper in core like:

```js
const isSameType = require('./isSameType')
const equals = require('./equals')

const mequals = (x, y) => 
  isSameType(x, y)
    && equals(x.valueOf(), y.valueOf())
```

dalefrancis88:
@evilsoft Do you think it's a worthwhile change? I personally like the reduction in repeated tests but curious what you think

evilsoft:
@dalefrancis88 I like it.I do no think it would reduce the number of specs though as we would want to test each in context. Which specs do you think could be omitted?

karthikiyengar:
@dalefrancis88, @evilsoft - I do not see the value of adding the helper, especially in the cases of `First` and `Last` where the `mequals` won't work since the underlying construct is a `Maybe`. Also, we are getting rid of neither tests nor additional parameters with the helper. But, if it's unanimous, I can add the helper.

@evilsoft, @bennypowers - With respect to the docs, can we extract it to a separate issue? I am not sure how the release process works, but do we need to have complete documentation around issues? If not, then this would help us push out many improvements in with a new release that would be of great help, given that there are no more code related issues in the release board.

@evilsoft - I will reorder the testcases and push a commit soon.

dalefrancis88:
@evilsoft I guess when it comes down to what can/should be removed it's more a question on what you believe should be tested. 

@karthikiyengar The value in having the method is that it communicates intent. It also means that when anything deviates from that intent it should be for a notable reason. Having all the tests around it will let you define it as an aspect as well, equality is not unique to each monoid. Now when it comes to the `Maybe`, the question i guess is how often the equal with the Maybe will show or is already in the code.  You could have the func check and support maybe or have another func that expects maybes? 

Either way if you decide not to do it, let me know and i'll do it later this weekend. I'm finally going to get some time to work on this in more than 15 min blocks this weekend

bennypowers:
@karthikiyenga my concern is that docs will stay out of sync for a significant amount of time, and users like me won't even be aware of awesome features. OTOH I certainly appreciate the desire to ship features quickly, especially as we pursue f/l coverage.

Is there a middle ground that will allow us to gain ground at pace while still keeping docs in the picture? In my view, even stubs are better than nothing. Perhaps a `stub` pragma which renders both a link to edit docs and a link to relevant source code? This is a larger issue.

karthikiyengar:
@dalefrancis88 - I cannot see how in this particular case the words `isSameType` and `equals` in 2 lines of code fail to communicate intent.

@bennypowers - Agreed - I will spend some time on the docs on the long weekend ahead. Please feel free to contribute in case you some spare time.

dalefrancis88:
Its all good, I was more pointing it out because its a pattern thta matches other parts of the code-base, like isArray, or isFunction, these are wrappers around 1 line of code but are still good.
I believe this may be because if there was ever a reason to change how we determine those, we're only changing the in one place rather than everywhere. 

But like I said it's not a major concern for this PR, if you get the changes done by @evilsoft suggested i'll but a :shipit: on it.

karthikiyengar:
No problem at all, I'll take care of it with the next line of commits :-)

evilsoft:
@karthikiyengar how would you feel about merging this one in with the VERSION updates.
Then we can do the docs on another PR, and just wait to ship a new version after that is merged in. 
I ask because I have a big change PR that is going to 🌽flict with this one in every file. Would be easier for me to rebase these changes into mine.

docs/src/pages/docs/monoids/First.md
```diff
@@ -138,6 +138,40 @@ concat(concat(a, b), c)
 //=> First( Just "c" )
 ```
 
+#### equals
+
+```haskell
+First a ~> b -> Boolean
+```
+
+Used to compare the underlying values of (2) `First` instances for equality by value, equals takes any given argument and returns `true` if the passed argument is a `First` with an underlying value equal to the underlying value of the `First` the method is being called on. If the passed argument is not a `First` or the underlying values are not equal, equals will return `false`.
```
bennypowers: There's a convention in place elsewhere in the docs to start a document with a description. See for example https://github.com/evilsoft/crocks/blob/master/docs/src/pages/docs/monoids/Any.md which begins "`Any` is a `monoid` which..."

I think it would be good to stick to that convention here, too.

docs/src/pages/docs/monoids/Last.md
```diff
@@ -136,6 +136,40 @@ concat(concat(a, b), c)
 //=> Last( Just "a" )
 ```
 
+#### equals
+
+```haskell
+Last a ~> b -> Boolean
+```
+
+Used to compare the underlying values of (2) `Last` instances for equality by value, equals takes any given argument and returns `true` if the passed argument is a `Last` with an underlying value equal to the underlying value of the `Last` the method is being called on. If the passed argument is not a `Last` or the underlying values are not equal, equals will return `false`.
```
bennypowers: Same as above

docs/src/pages/docs/monoids/First.md
bennypowers: > ... equals takes any given argument and returns `true` if the passed argument is a `First` with an underlying value equal to the underlying value of the `First` the method is being called on.

A little wordy. You can cancel out the second mention of the argument something like

> `equals` is a unary method which returns `true` if its argument is a `First` with an underlying value equal to the underlying value of the `First` the method is being called on. It will return `false` for any other argument.

src/First/README.md
```diff
@@ -121,6 +121,40 @@ concat(concat(a, b), c)
 //=> First( Just "c" )
 ```
 
+#### equals
+
+```haskell
+First a ~> b -> Boolean
+```
+
+Used to compare the underlying values of (2) `First` instances for equality by value, equals takes any given argument and returns `true` if the passed arguments is a `First` with an underlying value equal to the underlying value of the `First` the method is being called on. If the passed argument is not a `First` or the underlying values are not equal, equals will return `false`.
```
bennypowers: comments from above apply here as well

docs/src/pages/docs/monoids/First.md
evilsoft: @bennypowers 
I am confused. This is just an addition to an existing document.
It looks like this is in the proper format starting on line 1.

bennypowers: You're right good catch!

karthikiyengar:
Consider it done. I will send in another PR in a few hours.

karthikiyengar:
Closing in favour of https://github.com/evilsoft/crocks/pull/258

docs/src/pages/docs/monoids/First.md
karthikiyengar: Thanks for the comment, but I guess `equals` is not really unary?

PR #252, opened 2018-04-03
Add details tags for the All Monoid signatures [docs]

rstegg:
This adds a sample taste for using `details` tags for alternative explanations the documentation.

This tag is not currently supported by IE or edge browsers.

### DIFF
--- a/src/All/README.md
+++ b/src/All/README.md
@@ -3,9 +3,12 @@
 ```haskell
 All Boolean
 ```
-<details>
-  <summary>Breakdown</summary>
-  The Monoid <code> All </code> outputs type <code> Boolean </code>
+<details style="margin:30px">
+  <summary style="margin-bottom:15px">Breakdown</summary>
+  <dl>
+    <dt><code>All</code></dt><dd>The Monoid container</dd>
+    <dt><code>Boolean</code></dt><dd>The contained type</dd>
+  </dl>
 </details>
 
 `All` is a `Monoid` that will combine (2) values of any type using logical
@@ -48,6 +51,14 @@ allGood([ 'nice', '00', null ])
 ```haskell
 All.empty :: () -> All
 ```
+<details style="margin:30px">
+  <summary style="margin-bottom:15px">Breakdown</summary>
+  <dl>
+    <dt><code>All.empty()</code></dt><dd>The constructor function</dd>
+    <dt><code>All</code></dt><dd>The output type</dd>
+  </dl>
+</details>
+
 <details>
   <summary>Breakdown</summary>
   The constructor <code> All.empty() </code> outputs type <code> All </code>
@@ -74,9 +85,12 @@ All(false).concat(All.empty())  //=> All false
 ```haskell
 All ~> All -> All
 ```
-<details>
-  <summary>Breakdown</summary>
-  <code> All.concat </code> inputs type <code> All </code> and outputs type <code> All </code>
+<details style="margin:30px">
+  <summary style="margin-bottom:15px">Breakdown</summary>
+  <dl>
+    <dt><code>All</code></dt><dd>The input type</dd>
+    <dt><code>All</code></dt><dd>The output type</dd>
+  </dl>
 </details>
 
 `concat` is used to combine (2) `Semigroup`s of the same type under an operation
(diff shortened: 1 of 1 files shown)


src/All/README.md
```diff
@@ -3,6 +3,10 @@
 ```haskell
 All Boolean
 ```
+<details>
```
evilsoft: We may want to do something like

```html
<details style="margin:30px">
  <summary style="margin-bottom:15px">Breakdown</summary>
  <p>The Monoid <code> All </code> outputs type <code> Boolean </code></p>
</details>
```
For the Markdown READMEs...Then we can add css to the Electric Docs so we can get better control on the responsiveness

src/All/README.md
```diff
@@ -66,6 +74,10 @@ All(false).concat(All.empty())  //=> All false
 ```haskell
 All ~> All -> All
 ```
+<details>
+  <summary>Breakdown</summary>
+  <code> All.concat </code> inputs type <code> All </code> and outputs type <code> All </code>
```
evilsoft: I am thinking instead of providing text blocks that the user needs to parse through, we could come up with a 🌽sistant  list format that lists each in order by arrow, with the last always being a return. so this could be:

```html
<dl>
  <dt>all<dt><dd>The <code>All</code> to combine<dd>
  <dt><i>Result</i><dt><dd>The combined <code>All</All><dd>
</dl>
```

src/All/README.md
```diff
@@ -3,6 +3,13 @@
 ```haskell
 All Boolean
 ```
+<details style="margin:30px">
```
evilsoft: Looks like github strips the style attrib.
Bummer.

src/All/README.md
```diff
@@ -44,6 +51,18 @@ allGood([ 'nice', '00', null ])
...
+  <summary style="margin-bottom:15px">Breakdown</summary>
+  <dl>
+    <dt><code>All.empty()</code></dt><dd>The constructor function</dd>
+    <dt><code>All</code></dt><dd>The output type</dd>
+  </dl>
+</details>
+
+<details>
```
evilsoft: Got an extra breakdown here.

src/All/README.md
evilsoft: Also this should go in a new constructor "heading" above `Constructor Methods`:

## Constructor
```haskell
All :: a -> All Boolean
```
<details>
  <summary style="margin-bottom:15px">Breakdown</summary>
  <dl>
    <dt><code>value</code></dt><dd>Value to wrap</dd>
    <dt><code>result</code></dt><dd>All wrapping the coerced provided value</dd>
  </dl>
</details>

evilsoft:
All of these breakdowns still contain Types in the term, that should actually be a name to identity what that argument is. The type is already communicated in the signature. A good example of this is in this suggestion here.:
## Constructor
```haskell
All :: a -> All Boolean
```
<details>
  <summary style="margin-bottom:15px">Breakdown</summary>
  <dl>
    <dt><code>value</code></dt><dd>Value to wrap</dd>
    <dt><code>result</code></dt><dd>All wrapping the coerced provided value</dd>
  </dl>
</details>

evilsoft:
Also how to read these type signatures will be covered in the Type Signature Guide.
So just having `All Boolean` at the top should be enough with that guide.

evilsoft:
And it looks like it also strips <br> tags.
Looks like we are stuck with how it lays out.

src/All/README.md
```diff
@@ -44,6 +44,13 @@ allGood([ 'nice', '00', null ])
 ```haskell
 All.empty :: () -> All
 ```
+<details>
+  <summary>Breakdown</summary>
+  <dl>
+    <dt>result</dt><dd><code>All</code> wrapping <code>true</code></dd>
```
evilsoft: This does have an input of Unit or () as that is to the left of the arrow.
so it could be

<dt>unit</dt><dd>nothing, all arguments passed will be ignored</dd>

rstegg: Is that necessary?

src/All/README.md
```diff
@@ -4,6 +4,20 @@
 All Boolean
 ```
 
+## Constructor
```
evilsoft: This should go right after Implements and before [Constructor Methods](https://github.com/evilsoft/crocks/pull/252#discussion_r178992967).
Then this should be GTG I believe.

evilsoft:
Closing for now, will still do these tho.

Issue #254, opened 2018-04-04
Search on Documentation is not working correctly


HenriqueLimas:
The search for documentation is not returning nice results. Trying to search for `assoc` for example, it doesn't return the correct place of the documentation for `assoc`. Another example is searching for `Maybe`, it should return the `Maybe` documentation.

evilsoft:
[PR](https://github.com/evilsoft/crocks/pull/255) was merged....keeping this open to track the by function name portion

evilsoft:
Actually going to close this for now. we can revisit when you start getting reports

PR #255, opened 2018-04-05
Fix search in docs (#254)

rstegg:
This change only uses `Title`, `Description`, and `Tags` as the search fields, and adds function names to the tags field of their relative documentation.

### DIFF
--- a/docs/src/pages/docs/functions/transformation-functions.md
+++ b/docs/src/pages/docs/functions/transformation-functions.md
@@ -2,11 +2,9 @@
 title: "Transformation Functions"
 description: "Transformation Functions API"
 layout: "notopic"
-tags: ["arraytolist", "eithertoasync", "eithertofirst", "eithertolast", "eithertomaybe", "eithertoresult", "firsttoasync", "firsttoeither", "firsttolast", "firsttomaybe", "firsttoresult", "lasttoasync", "lasttoeither", "lasttofirst", "lasttomaybe", "lasttoresult", "listtoarray", "maybetoasync", "maybetoeither", "maybetofirst","maybetolast", "maybetoresult", "resulttoasync", "resulttoeither", "resulttofirst", "resulttolast", "resulttomaybe", "writertopair"]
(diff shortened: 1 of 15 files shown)


evilsoft:
Is there a way to get function names into the the search, like with `assoc` as was mentioned by @HenriqueLimas in [this issue](https://github.com/evilsoft/crocks/issues/254).

Also it would be great to add some description on how this fixes the problem. Helps with knowledge sharing for people browsing the PRs.

evilsoft:
I would recommend adding a `functions` tag to the meta data.
Then on the pages that contain definitions for function, add a list of their names...
like for combinators this should be:

`functions: "applyTo composeB constant flip identity substitution"`

and for maybe this should be:

`functions: "find, prop, propPath, safe, safeAfter, safeLift, eitherToMaybe, firstToMaybe, lastToMaybe, resultToMaybe"`

Notice for the ADT, only for functions that are defined in that module, not the methods.

rstegg:
This is fixed with tags now.

docs/src/pages/docs/functions/transformation-functions.md
```diff
@@ -2,9 +2,12 @@
 title: "Transformation Functions"
 description: "Transformation Functions API"
 layout: "notopic"
+tags: ["arraytolist", "eithertoasync", "eithertofirst", "eithertolast", "eithertomaybe", "eithertoresult", "firsttoasync", "firsttoeither", "firsttolast", "firsttomaybe", "firsttoresult", "lasttoasync", "lasttoeither", "lasttofirst", "lasttomaybe", "lasttoresult", "listtoarray", "maybetoasync", "maybetoeither", "maybetofirst","maybetolast", "maybetoresult", "resulttoasync", "resulttoeither", "resulttofirst", "resulttolast", "resulttomaybe", "writertopair"]
 weight: 60
 ---
 
+
```
HenriqueLimas: Not sure we need that extra spaces

rstegg:
It doesn't seem like that is possible with the current search engine supplied.  Though, that would be nice to have.

HenriqueLimas:
Agree with @evilsoft, I think we can merge it. This will improve a lot the search. I will take a look if that is possible and if not, I will talk with guys (add an issue) on electric.js

Nice work @rstegg

evilsoft:
Have a Fix It Squirrel
image

PR #257, opened 2018-04-09
fixes missing curry in isSame

### LINKED ISSUE #256: Predicates Should be Pre-Curried
```js
const sameId = id =>
  compose(curry(isSame)(id), propOr(null, 'id'));
```
☝️  how we do now
👇 if we were auto curried

```js
const sameId = id =>
  compose(isSame(id), propOr(null, 'id'));
```

compare https://github.com/evilsoft/crocks/blob/309c9d9a898efe690058cfb5a5b485a12e9f8eeb/src/core/isSame.js#L13 to https://github.com/evilsoft/crocks/blob/309c9d9a898efe690058cfb5a5b485a12e9f8eeb/src/helpers/propOr.js#L29

bennypowers:
fixes #256

### DIFF
--- a/src/core/isSame.js
+++ b/src/core/isSame.js
@@ -1,6 +1,8 @@
 /** @license ISC License (c) copyright 2017 original and current authors */
 /** @author Ian Hofmann-Hicks (evil) */
 
+const curry = require('./curry');
+
 // isSame : (a, b) -> Boolean
 function isSame(x, y) {
   if(x === y) {
@@ -10,4 +12,4 @@ function isSame(x, y) {
   return x !== x && y !== y
 }
 
-module.exports = isSame
+module.exports = curry(isSame)
--- a/src/core/isSame.js
+++ b/src/core/isSame.js
@@ -1,8 +1,6 @@
 /** @license ISC License (c) copyright 2017 original and current authors */
 /** @author Ian Hofmann-Hicks (evil) */
 
-const curry = require('./curry');
-
 // isSame : (a, b) -> Boolean
 function isSame(x, y) {
   if(x === y) {
@@ -12,4 +10,4 @@ function isSame(x, y) {
   return x !== x && y !== y
 }
 
-module.exports = curry(isSame)
+module.exports = isSame
--- a/src/All/All.spec.js
+++ b/src/All/All.spec.js
@@ -1,4 +1,5 @@
 const test = require('tape')
+const MockCrock = require('../test/MockCrock')
 const helpers = require('../test/helpers')
 
 const bindFunc = helpers.bindFunc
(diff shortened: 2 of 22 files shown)


src/core/isSame.js
```diff
@@ -10,4 +12,4 @@ function isSame(x, y) {
   return x !== x && y !== y
 }
 
-module.exports = isSame
+module.exports = curry(isSame)
```
karthikiyengar: @bennypowers - We should probably do this in `src/predicates` rather than here for consistency and perf reasons.

bennypowers: np but in that case should we do the same with `isSameType`?

bennypowers:
```
# Result Err ap functionality
not ok 3659 returns a Result
  ---
    operator: ok
    expected: true
    actual:   false
    at: Test.<anonymous> (/home/travis/build/evilsoft/crocks/build/Result/Result.spec.js:682:5)
    stack: |-
      Error: returns a Result
          at Test.assert [as _assert] (/home/travis/build/evilsoft/crocks/node_modules/tape/lib/test.js:225:54)
          at Test.bound [as _assert] (/home/travis/build/evilsoft/crocks/node_modules/tape/lib/test.js:77:32)
          at Test.ok.Test.true.Test.assert (/home/travis/build/evilsoft/crocks/node_modules/tape/lib/test.js:346:10)
          at Test.bound [as ok] (/home/travis/build/evilsoft/crocks/node_modules/tape/lib/test.js:77:32)
          at Test.<anonymous> (/home/travis/build/evilsoft/crocks/build/Result/Result.spec.js:682:5)
          at Test.bound [as _cb] (/home/travis/build/evilsoft/crocks/node_modules/tape/lib/test.js:77:32)
          at Test.run (/home/travis/build/evilsoft/crocks/node_modules/tape/lib/test.js:96:10)
          at Test.bound [as run] (/home/travis/build/evilsoft/crocks/node_modules/tape/lib/test.js:77:32)
          at Immediate.next [as _onImmediate] (/home/travis/build/evilsoft/crocks/node_modules/tape/lib/results.js:71:15)
          at processImmediate [as _immediateCallback] (timers.js:383:17)
  ...
```
🤷‍♂️

evilsoft:
Sorry about that horrible suggestion.
But I would say `isSame` should remain free of curry in core (like you have it)  and we leave isSameType as is...could end up being a rabbit hole with no bit value.

bennypowers:
I'm fine with that, but I think it might be worthwhile leaving a comment in `core/isSameType` explaining why. Something like:
```js
// curried in core to support legacy uses
module.exports = curry(isSameType)
```

evilsoft:
image

Issue #234, opened 2018-02-26
Report Proper Function Name on Error


evilsoft:
Now that we are moving to FL integration through the use of pointfree and helper functions, we need a way to signal to the user the path in which the error occurred. If using the FL methods, it should be reported as such.

That means if an error occurred using the "public" or "fluent" version of the function it should read:
`Either.map: Function required`

But if using the FL method, typically through a helper/pointfree function, the error should read:
`Either.fantasy-land/map: Function required`

This will require that the type functions that have FL equivalents must be partial applied with the name that will be in the method name space:
`[Type].[methodName]: Error Message`

dalefrancis88:
@bennypowers are you still looking for something? what about this?

evilsoft:
This shipped. Closing

PR #260, opened 2018-04-11
Add ability to report fluent style and fantasy-land errors 

evilsoft:
## Error ALL the things
image

As the next step in :100: Fantasy-Land Integration, we need to report errors properly (when called with another lib pointfree, or our future pointfree functions). This is described in [this issue](https://github.com/evilsoft/crocks/issues/234), and addresses all throwing functions that are implemented currently. There will be another pass when we add `ap` and `traverse` integration.

We can capture those in another issues and this PR should effectively address the above issue.
Also cleaned up some older not Error specific specs and removed derived `ap` implementations in favor of the non-derived version. This will make it easier to handle the upcoming `fantasy-land/ap`  work.


evilsoft:
Gonna merge this in..
image

PR #261, opened 2018-04-12
Move to Standard Modules

bennypowers:
This work-in-progress pull requests aims to move the library to es modules
The goal is to future-proof against the day when node adopts esm without a flag, as well as to allow web developers to import crocks modules in the browser *today* using `<script type="module">`


bennypowers:
This first commit converts `require` expressions to `import` statements and `module.exports` to `export default` in a mechanical way.

I had to get a little creative with `types` imports.

Tests were not touched, nor the build system. The purpose is to elicit comment and get 'boots on the ground', so to speak.

src/transforms/index.js
```diff
@@ -1,30 +1,59 @@
...
-  maybeToFirst: require('../First/maybeToFirst'),
-  maybeToLast: require('../Last/maybeToLast'),
-  maybeToResult: require('../Result/maybeToResult'),
-  resultToAsync: require('../Async/resultToAsync'),
-  resultToEither: require('../Either/resultToEither'),
-  resultToFirst: require('../First/resultToFirst'),
-  resultToLast: require('../Last/resultToLast'),
-  resultToMaybe: require('../Maybe/resultToMaybe'),
-  writerToPair: require('../Pair/writerToPair')
+import arrayToList from '../List/arrayToList.js'
```
dalefrancis88: would export { default as arrayToList } from '../List/arrayToList.js' be better here? that way you don't have to import the default object and then extract it, it'll just be a import { arrayToList } from 'crocks/transforms'

src/predicates/isAlt.js
```diff
@@ -1,5 +1,6 @@
 /** @license ISC License (c) copyright 2017 original and current authors */
 /** @author Ian Hofmann-Hicks (evil) */
 
-module.exports =
-  require('../core/isAlt')
+import isAlt from '../core/isAlt.js'
+
+export default isAlt
```
dalefrancis88: could these be merged? export { default as isAlt } from '../core/isAlt'? all the ones below too

src/transforms/index.js
bennypowers: If it works, I'm down. This does change the behaviour though, You'd have to `import * as crocks from 'crocks'` if you wanted to destructure. What do you think? @evilsoft

src/predicates/isAlt.js
bennypowers: I'm ok, with above proviso.

bennypowers:
FYI `tape -r esm **/*.test.js` works. See https://github.com/substack/tape/issues/414

bennypowers:
Right then, got the tests working.

In addition to juggling syntax, this PR changes the build system

- rollup is preferred over webpack
- for the es modules build, `cp -r` is preferred over rollup 😉 
- buble and terser (the library formerly known as uglify-es) are brought in as rollup plugins
- `esm` package is sprinkled liberally throughout the build/test system
- a single-file cjs bundle is deposited in `build/cjs`, and the package's `main` field is changed to `build/cjs/crocks.js`
- a single-file minified umd bundle is plopped in `build/crocks.min.js` for people who are in to that kind of thing (although really, bundling should be the user's business)
- `module` and `jsnext:main` fields are added, pointing to `./index.js`

bennypowers:
While reviewing, please pay careful attention that nothing got lost during the rebase.

bennypowers:
Note: it's worth considering exporting named exports from `core`. We could provide the same external API but in a more idiomatic way with `import * from './foo/index.js'` for example

evilsoft:
@bennypowers is it possible to favor the `cjs` so we have a structure on npm where the root directory are the cjs files in their respective folders and then there is a structure for `esm` inside of an `esm` folder? So basically a swap of how it is now, with the addition of the esm folder having the import code?

It will be a larger install for the time being, but will be easier on current adopters to eventually migrate over when they see the value of modules. For me personally in my projects and projects at work, this is going to be a MAJOR refactor on our deps, for no real benefit. There may be others in this boat as well.

Also it appears that the `docs/electric.config.js` needs to remain `cjs` for its export.

bennypowers:
`main` still points to cjs so this will still work without using require hooks/runtime polyfills

```js
const { All, mreduce } = require('crocks')
```

I'm not sure this will preserve imports of the style

```js
const All = require('crocks/All')
const mreduce = require('crocks/pointfree/mreduce')
``` 

If not, it's likely that juggling the rollup config would do it (i.e. programatically generating a separate bundle for each non-test file in src)

If a user is importing by relative path from `node_modules` they'd get es modules. Is that a case you're concerned about covering?

-----

In some other libraries I've seen them provide standard modules in an `esm` directory. That strikes me as odd, since cjs is really the bespoke solution - modules are party if the language, not a runtime function.

I've also seen libraries publish a separate `foo-es` version of their package, which is still suboptimal, but strikes me as more reasonable. 

I think the ideal solution is to publish standard modules as the standard, and use package.json fields to route cjs users were they need to go, preferably, as you pointed out, without breaking existing app code.

bennypowers:
Of course, those users could also just staple on `-r esm` and have that cover all the problems for a nominal overhead. That's usually my step 0 when I start a node project. I agree though that it would be better to cover their cases without introducing dependencies.

bennypowers:
Rollup has a feature called experimentalPreserveModules that looks like it'd do the trick 

https://github.com/rollup/rollup/pull/1922

I'll look into adding that when I get a minute

bennypowers:
Ok I added that rollup feature, and if all my ducks are in a row, I think everything should Just Work™

dalefrancis88:
So this has become a very large PR. I am wondering if maybe there is a way it could be split up into successive smaller PRs? instead of trying to deal with all these conflicts maybe you could take the learnings from this and rebuild and deploy it in parts? We could even create a 1.0.0 branch to merge into

Just a suggestion, but it'll be completely up to you @bennypowers

bennypowers:
I'll get my bearings in the branch a little later on. Will probably just rebase onto this branch even if I end up starting from scratch. Thanks.

bennypowers:
Ok I made some progress, but got stuck on some test errors from `Maybe`

dalefrancis88:
KK i'll check it out shortly

bennypowers:
You're invited to the fork, so if you spot an easy fix, fire at will

dalefrancis88:
hahaha @bennypowers well that took me two hours to track down but found the issue. It'll be one you might have to keep an eye out for while you're doing the rest of the system.

Basically there was `Just` and `Maybe.Just`, you changed `Just` to `_Just` and `Maybe.Just` to `Just` but didn't update all the usages for the original `Just`

bennypowers:
That's great @dalefrancis88  thanks.

Now, for some reason, coverage is not running. I'll look into it.

bennypowers:
Some things to note

The `es`, `cjs`, and `system` builds have `preserveModules` set, so we could, say, publish
  - `build/es` as `@crocks/crocks`
  - `build/cjs` as `@crocks/legacy`
  - `build/system` as `@crocks/systemjs`
  - `build/umd` as `@crocks/bundled`

Rollup made this incredibly easy to pull off.
tape and nyc both support the `esm` require hook.

I changed the `npm test` script to run `spec:src` instead of `spec:build`. My assumption here is that rollup won't break tested src in build.

bennypowers:
@coveralls yalla. does tagging you unstick the build?

dalefrancis88:
Should we recreate this PR on the 1.0 branch?

bennypowers:
changed the base to `release/1.0`

bennypowers:
is this still on the slate for 1.0? the project milestone one has this one issue. I'd like to save it for last as it's largely mechanical

dalefrancis88:
This was accidently closed :/ sorry, i did not realise deleting the 1.0 project would close this. That is sooo stupid

dalefrancis88:
I closed 1.0 because it looks like we'll do a 13, maybe this could go into that?

bennypowers:
Works for me. Came back to this because I realised it needs to be part of a well-defined milestone

dalefrancis88:
Yeah i'd agree, i think we need for it to be a concerted effort to close all branches, do this commit, then continue after that

bennypowers:
https://twitter.com/sindresorhus/status/1349294527350149121?s=20

PR #263, opened 2018-04-18
Move to one truth for documentation.

evilsoft:
## There can be only ONE
image

As brought @karthikiyengar (and other), it would make sense to have a single truth for our documents. In order to get that to work we had to do some moving around of things. Also did some final work on the Setoid Monoid docs (adding equals to some Monoids). This PR does the following:

* Unignore the `crocks` and `monoids` markdown docs for eslint (used for linting the coding examples). Going to continue to ignore the `functions` folder until they have a pass to be cleaned up.
* Add HTML comment linting rule exceptions for some of the code examples (`Async` and `ReaderT`) in the main docs
* Extend the markdown renderer to remove those eslint comments (and all html comments).
* Remove all the old READMEs in the `src` folder, now the `docs` will be the main source of truth.
* Add a Markdown template to the project for aiding with composing documentation.

Addtions for Adding Setoid to some Monoids:
* Move to 80 char per line for `equals` description.
* Added links on the Pointfree Functions docs for all equals references


docs/src/pages/docs/monoids/All.md
```diff
@@ -99,7 +99,11 @@ All(false).concat(All(false)) //=> All false
 All a ~> b -> Boolean
 ```
 
-Used to compare the underlying values of (2) `All` instances for equality by value, equals takes any given argument and returns `true` if the passed argument is a `All` with an underlying value equal to the underlying value of the `All` the method is being called on. If the passed argument is not a `All` or the underlying values are not equal, equals will return `false`.
+Used to compare the underlying values of (2) `All` instances for equality by
+value, equals takes any given argument and returns `true` if the passed argument
```
HenriqueLimas: I think we can format `equals` as code (put btw ` char).

docs/src/pages/docs/monoids/All.md
```diff
@@ -99,7 +99,11 @@ All(false).concat(All(false)) //=> All false
...
+is a `All` with an underlying value equal to the underlying value of the `All` the
```
HenriqueLimas: Maybe a little typo here `is an` instead of `is a` (Not a native language, I can be wrong)

docs/src/pages/docs/monoids/Any.md
```diff
@@ -102,7 +102,11 @@ Any(false).concat(Any(false)) //=> Any false
 Any a ~> b -> Boolean
 ```
 
-Used to compare the underlying values of (2) `Any` instances for equality by value, equals takes any given argument and returns `true` if the passed argument is a `Any` with an underlying value equal to the underlying value of the `Any` the method is being called on. If the passed argument is not a `Any` or the underlying values are not equal, equals will return `false`.
+Used to compare the underlying values of (2) `Any` instances for equality by
+value, equals takes any given argument and returns `true` if the passed argument
```
HenriqueLimas: Same here for `equals`

docs/src/pages/docs/monoids/Assign.md
```diff
@@ -104,7 +104,11 @@ Assign({ b: 4 })
 Assign a ~> b -> Boolean
 ```
 
-Used to compare the underlying values of (2) `Assign` instances for equality by value, equals takes any given argument and returns `true` if the passed argument is a `Assign` with an underlying value equal to the underlying value of the `Assign` the method is being called on. If the passed argument is not a `Assign` or the underlying values are not equal, equals will return `false`.
+Used to compare the underlying values of (2) `Assign` instances for equality by
+value, equals takes any given argument and returns `true` if the passed argument
```
HenriqueLimas: And here 😅 , `equals` again you get the point 😄

docs/src/pages/docs/monoids/All.md
evilsoft: Ohhh. thought I got all of those. Good find.

docs/src/pages/docs/monoids/Assign.md
evilsoft: Will replace them all. thanks for the 👀

evilsoft:
Alrighty...This has been updated
When this is merged we should be able to cut a release!!

evilsoft:
image

PR #264, opened 2018-04-18
Create es-modules branch on upstream

bennypowers:
The purpose of this PR is to move work on the es-modules branch from my fork to the main repo, in order to facilitate collaboration.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,2 +1,3 @@
-module.exports =
-  require('./src')
+import crocks from './src/index.js'
+
+export default crocks
--- a/src/All/index.js
+++ b/src/All/index.js
@@ -3,16 +3,17 @@
 
 const VERSION = 2
 
-const _implements = require('../core/implements')
-const _inspect = require('../core/inspect')
-const _equals = require('../core/equals')
-const type = require('../core/types').type('All')
-const _type = require('../core/types').typeFn(type(), VERSION)
-const fl = require('../core/flNames')
+import _implements from '../core/implements.js'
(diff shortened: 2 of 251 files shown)


karthikiyengar:
@bennypowers - I was wondering what the status of this PR is. Anything I could do to help get this to master?

bennypowers:
I would love to revive this project.

Last I recalled this was slated for the next breaking change, but I'm not sure when that Is @evilsoft  can help estimate?

I made some attempts at publishing an esm build of crocks, but they've fallen short for the time being (see https://github.com/esm-bundle/crocks/pull/11#issuecomment-605512670)

So I'd really like to see crocks shipping esm sooner rather than later, as it remains a blocker for some at https://github.com/apollo-elements/apollo-elements/issues/57

If there's consensus, we can try a non-breaking in-place transform by undertaking the following steps:
1. Rewrite all src to es modules, as per this PR
2. Use rollup to output CJS in-place as `*.cjs` files. e.g. under `predicates`, we'd have:
  - hasProp.js - es module
  - hasProp.cjs - cjs module
  - hasProp.spec.cjs - cjs tests 
3. Publish a point release

That last point I'm not 100% sure about. It may require (pun!) cjs-only users to `require('crocks/**/*.cjs')` *with* extension. We can experiment and try it out.

Having poked around a little bit, we could supply a snippet, or a side-effecting module which registers `.cjs` for older node versions in [require.extensions](https://gist.github.com/jamestalmage/df922691475cff66c7e6)

PR #265, opened 2018-04-25
Make inspect method stringify objects (#251)

### LINKED ISSUE #251: All inspect methods should provide stringified representation for objects
The current behaviour when we `console.log` an object in a container is like so

```
const object = { a: 5, b: Just(10) }
console.log(Maybe.of(object))
=> Just {}
```
The developer experience could be greatly enhanced if we also display what the object contains like so: 
```
Just { a: 5, b: Just 10 }
```

This would involve modifying the inspect method of every ADT.

karthikiyengar:
This PR fixes #251 by recursively calling `inspect` on object properties.


evilsoft:
image
(shipping)

PR #266, opened 2018-04-25
Initialize Async cancel state in call to fork

NiallArkEnergy:
I came across this pain point today. When I wrap a function the state is initialised.
```javascript
const CrockFetcher = Async((rej, res) => fetch("/myApi")) // state { cancelled = false }
```
If I fork I can then cancel
```javascript
CrockFetcher.fork(() => console.log("fail"), () => console.log("success"))() // state { cancelled = true }
```
Later when I fork the fork callbacks won't run
```javascript
CrockFetcher.fork(() => console.log("no run"), () => console.log("no run")) // state { cancelled = true }
```
This pull request aims to only set state when the fork is called. Allowing the wrapped function to be forked again after a previous cancel


NiallArkEnergy:
Any suggestions on how to clean up the other references to cancel?

evilsoft:
GREAT FIND. Although I am sorry this is causing a pain point for you...I think the problem here is keeping that state itself...Going to think of a way to maybe use a union type here that just gets passed down (or up?)  the "cancel" chain. That way it can have a ripple effect without state.

I have always hated that state.

I will get back to you for sure with a couple ideas

evilsoft:
@NiallArkEnergy After playing with some options, we will need to get that state out of there...So the best way I can think we can address this is to make `Async` something like `Async e a () =  Rejected e | Resolved a | Cancelled`.

NiallArkEnergy:
I can't picture how that union type would work. Other libraries I've been looking at expect the function provided to Async to return a cleanup function in this case it's trivial to cancel. 

```javascript
const timer = ms => val => Async(function(rej, res){
  const id = setTimeout(()=>res(val), ms);
  return () => clearTimeout(id)
})

const timer1000 = timer(1000);
const delayed = timer1000('Delayed');
const cancel = delayed.fork(()=>console.log('fail'),()=>console.log('pass'))
cancel()

// delayed should work if I call fork here
```

crocks already uses this mechanism to guard the reject and resolve callbacks. I think adding the capture of the function dispose function removes the need for a parentCancel

This should simplify composing methods, no need to pass a cancel, just return the result of the call to fork
```javascript
function map(method) {
    return function(mapFn) {
      if(!isFunction(mapFn)) {
        throw new TypeError(`Async.${method}: Function required`)
      }

      return Async(function(reject, resolve) {
        return fork(reject, compose(resolve, mapFn))
      })
    }
  }
```

evilsoft:
Gotcha.
The `Cancelled` tag will never have to be explicitly constructed, this is mostly an internal tag, that is a noop through a "chain" of Asyncs. When the function is called it would chain a move to the `Cancelled` tag while one of the Asyncs in the chain is "in-flight" that way once it completed it will throw away the result and pass `Cancelled ()` down the line.

That was my thinking on adding another tag to Async.

I am curious about your solution, I do not see how it could cancel a wrapped Promise or Node CPS function in a chain like:

```js
promAsync
  .chain(anotherProm)
  .chain(someNodeAsync)
```

what would happen if that `promAsync` has the cancel logic and it has already been completed and `anotherProm` is in-flight when cancelled is called? I would expect it would never run `someNodeAsync` and never call res or rej functions passed to fork. I may be missing something with that approach, could you demonstrate how this could cancel the whole lot of them?

NiallArkEnergy:
Is there any use case I'm not thinking about that needs the parentCancel function. Those are the failing tests that I have at the moment. I will add tests also to check cancel functions are called

evilsoft:
Awesome. @NiallArkEnergy will get some 👀on this today.

evilsoft:
More like a Promise.

evilsoft:
Code wise this looks GREAT thanks for the PR. One last thing that will be needed is to show the ability to return an internal cancel function in the docs.
It can be added here in the description and example for fork:
https://github.com/evilsoft/crocks/blob/master/docs/src/pages/docs/crocks/Async.md

Also would you mind if we added you to the Contributors section on the main README?

NiallArkEnergy:
Finally got back to this. Sorry it took so long. Hope this is enough to explain my changes. Yeah please add me to the Contributors section, that sounds great

evilsoft:
Looks pretty darn good, here have a squirrel:
image

PR #268, opened 2018-04-27
Time to Upgrade

evilsoft:
## All Shiny and new!
image

With the release of node `10.x`, we can now remove the build dep on node `4.2`. Which in turn means we can FINALLY upgrade to webpack 4. 🎉 

This means that CONTRIBUTOR node must be `6.x`+, many of the updates here will not work with anything less.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: node_js
 
 node_js:
+  - '10'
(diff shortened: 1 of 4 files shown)


evilsoft:
Gonna merge this.
image

Issue #269, opened 2018-05-05
Add a Tuple Type


evilsoft:
A simple Tuple type that lets us define a tuple of any size would be amazing.
To get a constructor, the size of the tuple will need to be specified to get back an instance constructor, fixed to the size specified:

```js
const Triple = Tuple(3)

// t :: Tuple String Boolean Number
const t = Triple('a', true, 97)
```
*Implementation detail*: When providing the constructor, we should return (up to Tuple-10) an anonymous function with the length of the function matching the size of the tuple. Providing the proper length for the function allows us to integrate with other libs curry and partial application helpers.

```js
// suggested/possible implementation

function _Tuple(n) {
  if(!(isInteger(n) || n > 1)) {
    throw
  }
  const _type = () =>
     `${n}-Tuple`

  // instance impl
  function Tuple(n, ...parts) {
    if(parts.length !== n) {
       throw
    }

    // method defs
  
    return {
        inspect, toString: inspect, type: _type, concat, map,
        mapTuple, toArray, project, constructor: Tuple, merge
    }
  }

  Tuple['@@implements'] = () =>
    [ 'concat', 'map' ]

  Tuple.type = _type

  Tuple['@@type'] =
    `crocks/${n}-Tuple`
 
  switch(n) {
    case 1: return function(a) { return Tuple(n, a) }
    case 2: return function(a, b) { return Tuple(n, a, b) }
    case 3: return function(a, b, c) { return Tuple(n, a) }
    case 4: return function(a, b, c, d) { return Tuple(n, a, b, c, d) }
    case 5: return function(a, b, c, d, e) { return Tuple(n, a, b, c, d, e) }
    case 6: return function(a, b, c, d, e, f) { return Tuple(n, a, b, c, d, e, f) }
    case 7: return function(a, b, c, d, e, f, g) { return Tuple(n, a, b, c, d, e, f, g) }
    case 8: return function(a, b, c, d, e, f, g, h) { return Tuple(n, a, b, c, d, e, f, g, h) }
    case 9: return function(a, b, c, d, e, f, g, h, i) { return Tuple(n, a, b, c, d, e, f, g, h, i) }
    case 10: return function(a, b, c, d, e, f, g, h, i, j) { return Tuple(n, a, b, c, d, e, f, g, h, i, j) }
  }

  return function (...parts) { return _Tuple(n, ...parts) }
}
```

### Instance Methods
* `concat`: when all parts are a semi-group concat them together, if not throw
[...]

PR #275, opened 2018-05-24
Add a Tuple Type (#269)

karthikiyengar:
This PR introduces a Tuple (#269) type and a pointfree `project` function. 
Breaking Change: `Pair/merge` has now moved to `pointfree`


src/pointfree/project.js
```diff
@@ -0,0 +1,14 @@
+/** @license ISC License (c) copyright 2016 original and current authors */
+/** @author Karthik Iyengar (karthikiyengar) */
+
+const isFunction = require('../core/isFunction')
+
+function project(m) {
```
evilsoft: This will also need to take an Integer as its first argument, but we should only throw if it is not an Integer, no need to check the range between 1-10, we can let the underlying `m` handle that.

src/Tuple/index.js
```diff
@@ -0,0 +1,152 @@
...
+      && _equals(parts, m.toArray())
+    }
+
+    function concat(method) {
+      return function(t) {
+        if (!isSameType({ type }, t)) {
+          throw new TypeError(`${n}-Tuple.${method}: Tuple of the same length required`)
+        }
+        const a = t.toArray()
+        return _Tuple(n)(
```
evilsoft: Do we need to go through `_Tuple` for 🌽struction here? All other functions seem to use `Tuple`

src/Tuple/index.js
```diff
@@ -0,0 +1,152 @@
...
+    const inspect = () => `Tuple(${parts.map(_inspect).join(',')} )`
+
+    const type = constant(`${n}-${_type()}`)
+
+    function map(method) {
+      return function(fn) {
+        if (!isFunction(fn)) {
+          throw new TypeError(`${n}-Tuple.${method}: Function required`)
+        }
+        return _Tuple(n)(
```
evilsoft: Any reason we need to go all the way through `_Tuple` here? We could probably just go through `Tuple`

evilsoft: Also a newline above may make this function easier for the reader to parse.

src/Tuple/index.js
```diff
@@ -0,0 +1,152 @@
...
+  case 5: return function(a, b, c, d, e) { return Tuple(n, arguments) }
+  case 6: return function(a, b, c, d, e, f) { return Tuple(n, arguments) }
+  case 7: return function(a, b, c, d, e, f, g) { return Tuple(n, arguments) }
+  case 8: return function(a, b, c, d, e, f, g, h) { return Tuple(n, arguments) }
+  case 9: return function(a, b, c, d, e, f, g, h, i) { return Tuple(n, arguments) }
+  case 10: return function(a, b, c, d, e, f, g, h, i, j) { return Tuple(n, arguments) }
+  }
+  /* eslint-enable no-unused-vars */
+
+  function Tuple(n, args) {
```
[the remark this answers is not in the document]
evilsoft: Also `type` needs to be available on the constructor.
Need to be able to do:
```js
const Tuple3 = Tuple(3)
const instance = Tuple3(1, 2, 'b')

isSameType(Tuple3, instance)
//=> true
```

src/pointfree/mapAll.js
```diff
@@ -0,0 +1,21 @@
...
+  })
+  return (m) => {
+    if(!(m && isFunction(m.mapAll))) {
+      throw new TypeError('mapAll: Tuple required')
+    }
+    return m.mapAll(...parts)
+  }
+}
+
+module.exports = mapAll
```
evilsoft: We should make this curry-able and would make more sense living in `src/Tuple` as it only works on tuple types.
If we rename this to `nMap` the implementation should look like this:

```js
const validTuple = (n, m) =>
  isSameType(Tuple(n), m)

function runMap(m, fns) {
  const n = fns.length

  if (!validTuple(n, m)) {
    throw new TypeError(`nMap: ${n}-Tuple required`)
  }

  fns.forEach(fn => {
    if (!isFunction(fn)) {
      throw new TypeError('nMap: Functions required')
    }
  })

  return m.mapAll(...fns)
}

function dispatch(n) {
  switch (n) {
  case 1:
    return (a, m) =>
      runMap(m, [ a ])
  case 2:
    return (a, b, m) =>
      runMap(m, [ a, b ])
  case 3:
    return (a, b, c, m) =>
      runMap(m, [ a, b, c ])
  case 4:
    return (a, b, c, d, m) =>
      runMap(m, [ a, b, c, d ])
  case 5:
    return (a, b, c, d, e, m) =>
      runMap(m, [ a, b, c, d, e ])
  case 6:
    return (a, b, c, d, e, f, m) =>
      runMap(m, [ a, b, c, d, e, f ])
  case 7:
    return (a, b, c, d, e, f, g, m) =>
      runMap(m, [ a, b, c, d, e, f, g ])
  case 8:
    return (a, b, c, d, e, f, g, h, m) =>
      runMap(m, [ a, b, c, d, e, f, g, h ])
  case 9:
    return (a, b, c, d, e, f, g, h, i, m) =>
      runMap(m, [ a, b, c, d, e, f, g, h, i ])
  case 10:
    return (a, b, c, d, e, f, g, h, i, j, m) =>
      runMap(m, [ a, b, c, d, e, f, g, h, i, j ])
  }
}

function nMap(n) {
  if (!isInteger(n) && (n < 1 || n > 10)) {
    throw new TypeError(
      'nMap: Integer between 1 and 10 required for first argument'
    )
  }

  return dispatch(n)
}

module.exports =
  curry(nMap)
```

It is a doozy, but we allows for currying, and all the partial application goodness that comes along with it.

src/Tuple/index.js
```diff
@@ -0,0 +1,152 @@
...
+      inspect, toString: inspect, merge,
+      project, mapAll, toArray,
+      type, equals,
+      map: map('map'),
+      concat: concat('concat'),
+      [fl.map]: map(fl.map),
+      [fl.concat]: concat(fl.concat),
+      [fl.equals]: equals,
+      ['@@type']: typeString,
+      ['@@implements']: _implements([ 'map', 'concat', 'equals' ]),
```
evilsoft: This need to live on the constructor, not the instance

src/Tuple/index.js
```diff
@@ -0,0 +1,152 @@
...
+    return {
+      inspect, toString: inspect, merge,
+      project, mapAll, toArray,
+      type, equals,
+      map: map('map'),
+      concat: concat('concat'),
+      [fl.map]: map(fl.map),
+      [fl.concat]: concat(fl.concat),
+      [fl.equals]: equals,
+      ['@@type']: typeString,
```
evilsoft: This needs to also live on the constructor as well as the instance.

src/Tuple/index.js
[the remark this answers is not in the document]
karthikiyengar: I did try the spread operator out, but the behaviour seems unexpected. Here's a REPL: https://repl.it/@KarthikIyengar/spread-operator-quirk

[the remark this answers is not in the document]
karthikiyengar: If I understand you correctly,  we want `Tuple(3).type() => 3-Tuple`.

This really wouldn't be clean to achieve with the `switch/case` unless I'm missing something. It'd require that we set a type property on whatever function the `case` is returning.

src/Tuple/index.js
karthikiyengar: @evilsoft - Great catch :-)

src/pointfree/mapAll.js
karthikiyengar: @evilsoft - I will wait for our decision regarding getting rid of the `switch/case` before I go ahead and fix this. I think it would make an order of magnitude difference in getting rid of the indirection and improving readability.

src/pointfree/project.js
karthikiyengar: Could you please elaborate? I do not see any checks here except the `isFunction` check - everything else is indeed being deferred to the underlying `m`. Are you suggesting that we add another integer check here?

evilsoft: So how do we pass in WHICH element we want?
`m.project` take a number 1-10 yeah?

karthikiyengar: Hah, sorry - totally missed that. I shall add the integer!

karthikiyengar:
@evilsoft - Okay, took another run at it and hopefully addressed your review points. Could you please take a look?

src/Tuple/index.js
```diff
@@ -0,0 +1,162 @@
...
+  function Tuple(n, args) {
+    const parts = [].slice.call(args)
+    if (n !== parts.length) {
+      throw new TypeError(
+        `${n}-Tuple: Expected ${n} values, but got ${parts.length}`
+      )
+    }
+
+    const inspect =
+      () => `Tuple(${parts.map(_inspect).join(',')} )`
```
evilsoft: should be `${n}-Tuple(${parts.map(_inspect).join(',')} )`

src/Tuple/index.js
```diff
@@ -0,0 +1,162 @@
...
+  case 3: return withType(function(a, b, c) { return Tuple(n, arguments) })
+  case 4: return withType(function(a, b, c, d) { return Tuple(n, arguments) })
+  case 5: return withType(function(a, b, c, d, e) { return Tuple(n, arguments) })
+  case 6: return withType(function(a, b, c, d, e, f) { return Tuple(n, arguments) })
+  case 7: return withType(function(a, b, c, d, e, f, g) { return Tuple(n, arguments) })
+  case 8: return withType(function(a, b, c, d, e, f, g, h) { return Tuple(n, arguments) })
+  case 9: return withType(function(a, b, c, d, e, f, g, h, i) { return Tuple(n, arguments) })
+  case 10: return withType(function(a, b, c, d, e, f, g, h, i, j) { return Tuple(n, arguments) })
+  default:
+    throw new TypeError(
```
evilsoft: 👍

src/Tuple/index.js
```diff
@@ -0,0 +1,162 @@
...
+const isSameType = require('../core/isSameType')
+const isSemigroup = require('../core/isSemigroup')
+
+const constant = x => () => x
+
+function _Tuple(n) {
+  const type =
+    constant(`${n}-${_type()}`)
+
+  const withType = fn => {
```
evilsoft: 👍 
could use this function to add `@@type` and `@@implements`

src/Tuple/nMap.js
```diff
@@ -0,0 +1,68 @@
...
+  fns.forEach(fn => {
+    if (!isFunction(fn)) {
+      throw new TypeError('nMap: Functions required')
+    }
+  })
+
+  return m.nMap(...fns)
+}
+
+function nMap(n) {
```
evilsoft: following the 🌽vention of `bimap`, `contramap`, `promap`... I am thinking this should follow suit and become `nmap`?

src/Tuple/mapAll.js
```diff
@@ -0,0 +1,21 @@
+/** @license ISC License (c) copyright 2018 original and current authors */
+/** @author Karthik Iyengar (karthikiyengar) */
```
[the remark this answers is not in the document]
karthikiyengar: Not sure what I'm missing here, but the current implementation is indeed curried. I just got rid of the dispatch and added the default case to keep it consistent with the constructor.

For instance `nmap(3)(f => f)(f => f)(f => f)(q)` works fine.

I'd be happy to change things if you see any further issues / stylistic problems.

src/Tuple/index.spec.js
```diff
@@ -0,0 +1,407 @@
...
+
+  t.ok(isFunction(m.inspect), 'provides an inpsect function')
+  t.equal(m.inspect, m.toString, 'toString is the same function as inspect')
+  t.equal(m.inspect(), '2-Tuple( 0, "nice" )', 'returns inspect string')
+
+  t.end()
+})
+
+test('Tuple type', t => {
+  const m = Tuple(1)(0)
```
evilsoft: Gonna wanna write specs for each `n`, up to `10`.

src/Tuple/index.spec.js
```diff
@@ -0,0 +1,407 @@
...
+
+  t.equal(f('equals'), true, 'implements equals func')
+  t.equal(f('map'), true, 'implements map func')
+  t.equal(f('concat'), true, 'implements concat func')
+
+  t.end()
+})
+
+test('Tuple inspect', t => {
+  const m = Tuple(2)(0, 'nice')
```
evilsoft: Gonna wanna write specs for each `n`, up to `10`.

src/Tuple/index.spec.js
```diff
@@ -0,0 +1,407 @@
...
+test('Tuple @@type', t => {
+  const m = Tuple(1)( 0)
+
+  t.equal(m['@@type'], 'crocks/1-Tuple@1', 'returns crocks/1-Tuple@1 as Monoid Type')
+
+  t.end()
+})
+
+test('Tuple project', t => {
+  const tuple = Tuple(3)('zalgo', 'is', 'back')
```
evilsoft: gonna wanna go to the max `10-tuple`.
 and project on each position.

src/Tuple/index.js
```diff
@@ -0,0 +1,163 @@
...
+
+    function merge(fn) {
+      if (!isFunction(fn)) {
+        throw new TypeError(`${n}-Tuple.merge: Function required`)
+      }
+
+      return fn(...parts)
+    }
+
+    function nmap(...args) {
```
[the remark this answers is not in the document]
karthikiyengar: I thought about this and then I noticed that all other pointfree functions are named after instance methods. I'd naturally reach out for `pointfree/mapAll` if I'm trying to convert a `Tuple(..).mapAll(..)` to a pointfree compose chain.

That being said, I feel your reasoning is valid because `nmap` does take context. The only other way we can call the pointfree function `mapAll` would be to make it smarter by detecting `Tuple` lengths (https://github.com/evilsoft/crocks/pull/275/commits/197c9d1d9a16f0c934c06eb668f0be0d5ce861fd) - but since we decided against that, I'm going to go ahead and make this change.

karthikiyengar:
It's up 🐿! Please don't hesitate to let me know if something needs to be changed.

PR #276, opened 2018-05-29
Add construction section for all Monoids

evilsoft:
## Lets talk 🌽struction
image

One thing I think that will help people out is to provide a ***Construction*** section to all Monoids and Crocks as part of the documentation. Sometimes is can be very difficult to sus out how to construct on of these things based on examples and signatures alone.

These sections will provide any construction information as well as an example (or many, depending on special features of the type) that should convay not so much how to use the type, but how to construct one.

The general example (first example) in the docs should speak to general usage, this section should speak to construction.

This PR provides the construction section for all Monoids as well as an update to the Docs template for future use. Will follow this up with a the Crocks updates on another PR.


docs/src/pages/docs/monoids/First.md
```diff
@@ -65,6 +66,45 @@ chooseFirst([ 'a', 'b', 'c' ])
...
+
+## Construction
+
+```haskell
+First :: a -> First (Maybe a)
+First :: Maybe a -> First (Maybe a)
+```
+
+There are two ways to construct a `First` instance using the constructor. The
+first requires that you provide any value (excluding a `Maybe a`) and you will
```
karthikiyengar: first here should be `First`

evilsoft: well it is the "first" of 2 ways of constructing.
Maybe the english should be a bit better.

docs/src/pages/docs/monoids/Sum.md
```diff
@@ -42,6 +42,41 @@ sumByTen([ 2, 2 ])
...
+
+Passing `undefined`, `NaN` or `null` to the constructor will result in an
+`empty` instance, or an instance containing a `0`.
+
+```javascript
+import Sum from 'crocks/Sum'
+
+import equals from 'crocks/pointfree/equals'
+
+Sum(41)
```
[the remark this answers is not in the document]
evilsoft: Oh yes I did. Glad you caught that, did not know if anyone would.

evilsoft:
@bennypowers and @karthikiyengar
Thanks for the 👀. If you could, would it be possible to get another pass and make sure I addressed all of your 🌽cerns!

evilsoft:
image

PR #279, opened 2018-06-10
Add Fantasy Land Compliance for Associated Predicates

evilsoft:
## Fantasy All the THINGS!!
image

As a first step to [this issue](https://github.com/evilsoft/crocks/issues/202) we need to get the Fantasy Land Methods/Function specific predicates to respect the fantasy-land spec. This PR adds the following to all fantasy-land predicates:
* Check for method on passed object (Constructor/Instance), favoring the fantasy-land prefix over non-decorated.

* Check for method on `constructor` for methods specific to the `TypeRep`. `crocks` provides these functions on the instance as well as the `TypeRep`, but not all libs do that, so we need to check the `TypeRep` of the instance.

* Check for `@@implements` method. There are functions in crocks that can take a `TypeRep` as arguments and need to use that `TypeRep` for typeclass validation. For instance `isFunctor(Identity)` would return true. This is a sole feature of `crocks` and is not part of the fantasy-land spec. As a result only `crocks` or outside types that implement the `@@implements` method will return true when checked against a given `TypeRep`

* Also due to the lack of `ap` and `traverse`, there are a few predicates that will return false for fantasy-land types until those specs are fully implemented:
  * `isAlternative`( must be an `Applicative`)
  * `isApplicative`
  * `isApply`
  * `isChain` (must be an `Apply`)
  * `isMonad` (must be a `Chain`)
  * `isTraversable`

### DIFF
--- a/src/test/helpers.js
+++ b/src/test/helpers.js
@@ -1,8 +1,19 @@
+const fl = require('../core/flNames')
+
 const id =
   x => x
 
 const slice =
   x => Array.prototype.slice.call(x)
+
+const contains = xs => x =>
+  xs.indexOf(x) !== -1
+
+const repFuncs =
+  [ 'id', 'empty', 'of', 'zero' ]
+
+const isRepFunc =
+  contains(repFuncs)
 
 function bindFunc(fn) {
   return function() {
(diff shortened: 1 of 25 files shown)


src/test/helpers.js
```diff
@@ -1,29 +1,55 @@
+const fl = require('../core/flNames')
+
 const id =
   x => x
 
 const slice =
   x => Array.prototype.slice.call(x)
 
+const contains = xs => x =>
+  xs.indexOf(x) !== -1
```
dalefrancis88: given that this is a helper, do we care that xs could be null here? I am thinking no because being that its a test, it's going to very publicly break

evilsoft: I agree with you.

evilsoft:
image

PR #280, opened 2018-06-14
Add Fantasy Land Compliance for Pointfree Functions

evilsoft:
## The Latest in Point Free Energy
image

This PR finishes off [this issue](https://github.com/evilsoft/crocks/issues/202) by adding a preference to fantasy-land methods/functions over non-decorated methods/functions.

Like with the predicates, this covers all of the methods/functions that are implemented at this time. When we add the `ap` and `traverse` implementations, we will go back and add both the assosiated predicates and pointfree functions.

Tested these against `folktale`, `Fluture`, `creed` and `most`

### DIFF
--- a/src/core/equals.js
+++ b/src/core/equals.js
@@ -53,7 +53,7 @@ function equals(a, b) {
   }
 
   if(hasAlg('equals', a)) {
-    return (b[fl.equals] || b.equals).bind(b)(a)
+    return (b[fl.equals] || b.equals).call(b, a)
   }
 
   return (strats[type(a)] || comp)(a, b)
--- a/src/core/equals.js
+++ b/src/core/equals.js
@@ -5,6 +5,7 @@ const isSameType = require('./isSameType')
 const isSame = require('./isSame')
 const hasAlg = require('./hasAlg')
 const type = require('./type')
+const fl = require('./flNames')
 
 const comp = (a, b) =>
   a.valueOf() === b.valueOf()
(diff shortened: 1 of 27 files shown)


src/core/equals.js
```diff
@@ -52,8 +53,9 @@ function equals(a, b) {
   }
 
   if(hasAlg('equals', a)) {
-    return a.equals(b)
+    return (b[fl.equals] || b.equals).bind(b)(a)
```
HenriqueLimas: IMHO `apply` or `call` are more appropriate here, I think `bind` creates a function every time.

src/core/equals.spec.js
```diff
@@ -22,19 +31,31 @@ test('equals primatives', t => {
   t.end()
 })
 
-test('equals adts', t => {
-  const value = 'value'
-  const spy = sinon.spy(() => value)
+test('equals Setoid', t => {
```
HenriqueLimas: I think we should add some test related to the `this` context. Since someone can remove the bind and the test will still work.

```js
const b = {
  prop: 'yay',
  equals(arg) {
    this.foo === arg
  }
};
```

evilsoft: I do not understand what you mean.
Does ln 42 not cover this?

HenriqueLimas: Oh, I mean like use some object that has a method `equals` and inside it use some value that is a property of the object. Like the variable `b` that I mention. It has a property `prop` and inside the `b.equals` method we are checking against that property using `this.prop`. if we remove the `bind` in the implementation of the `equals`, `this` will be undefined.

HenriqueLimas: Oooh, you are right, sorry, did not get it.

rpearce:
👏

PR #283, opened 2018-06-25
Constant should be Curried

evilsoft:
## Kurry That Kestrel
image

When playing with Combinators, found that is it super silly to NOT have `constant` curried.

So this PR provides a Curried Constant


evilsoft:
image

PR #284, opened 2018-06-25
Support arbitrary length Tuples (#269)

karthikiyengar:
This PR adds support for arbitrary length Tuples beyond the initial 10 element limit.


evilsoft:
image

PR #295, opened 2018-07-21
Added pointfree for toPromise

dalefrancis88:
Added to help enable composition

### DIFF
--- a/src/pointfree/toPromise.js
+++ b/src/pointfree/toPromise.js
@@ -1,0 +1,14 @@
+/** @license ISC License (c) copyright 2018 original and current authors */
+/** @author Dale Francis */
+
+const isFunction = require('../core/isFunction')
+
+function toPromise(m) {
+  if(!(m && isFunction(m.toPromise))) {
+    throw new TypeError('toPromise: Async required')
+  }
+
+  return m.toPromise()
+}
+
+module.exports = toPromise
--- a/src/pointfree/toPromise.spec.js
+++ b/src/pointfree/toPromise.spec.js
@@ -1,0 +1,22 @@
+const test = require('tape')
+
+const isFunction  = require('../core/isFunction')
+
+const Async = require('../Async')
+const toPromise = require('./toPromise')
+
+test('toPromise pointfree', t => {
+  t.plan(3)
+
+  const val = 1337
+
+  const rej = y => x => t.equal(x, y, 'rejects a rejected Async')
+  const res = y => x => t.equal(x, y, 'resolves a resolved Async')
+
+  const testFunc = a => toPromise(a).then(res(val)).catch(rej(val))
+
+  t.ok(isFunction(toPromise), 'is a function')
+
+  testFunc(Async.Rejected(val))
+  testFunc(Async.Resolved(val))
+})


bennypowers:
LGTM but should maybe add a unit test for the case where `f` is not an async

dalefrancis88:
Yeah fair point, let me add that now

dalefrancis88:
Do you want to merge this @bennypowers? :) I can't it seems

src/pointfree/toPromise.spec.js
```diff
@@ -0,0 +1,39 @@
...
+
+const bindFunc = helpers.bindFunc
+
+const isFunction  = require('../core/isFunction')
+
+const Async = require('../Async')
+const toPromise = require('./toPromise')
+
+test('toPromise pointfree', t => {
+  t.plan(13)
```
karthikiyengar: Usually for pointfree functions, we only test if the underlying function on the container has been called using `sinon`. If I am not wrong, functionality tests are to be reserved for the crocks themselves.

@evilsoft - Thoughts?

bennypowers:
I'll defer to the big dogs, but thanks 🌟😉

dalefrancis88:
I agree, good point :)

Change committed

src/pointfree/toPromise.spec.js
```diff
@@ -0,0 +1,39 @@
...
+const bindFunc = helpers.bindFunc
+
+const isFunction  = require('../core/isFunction')
+
+const constant = x => () => x
+
+const toPromise = require('./toPromise')
+
+test('toPromise pointfree', t => {
+  t.plan(13)
```
karthikiyengar: We do not need `t.plan` anymore because it's no longer async.

dalefrancis88: ahh ofcourse.... this is what i get for coding quickly and not checking :)

dalefrancis88:
Well spotted! I thought that was the first thing i did

karthikiyengar:
Also, I noticed that we also don't have `fromPromise` and `fromNode` pointfree. Would you have the time to take that up as well?

dalefrancis88:
Yeah can do, something i've done has screwed with the build wierdly. I have no problem locally though. Can you see what it is by looking?

src/pointfree/toPromise.js
```diff
@@ -0,0 +1,14 @@
+/** @license ISC License (c) copyright 2018 original and current authors */
+/** @author Dale Francis */
+
+const isFunction = require('../core/isFunction')
+
+function toPromise(m) {
+  if(!(m && isFunction(m.toPromise))) {
```
evilsoft: As this is only for the `Async` type, this should live in `src/Async/toPromise`. Then as it is in the same folder `isSameType` should be used like so:

```js
const Async = require('./index.js')
const isSameType = require('../core/isSameType.js')

function toPromise(m) {
  if(!isSameType(Async, m)) {
    throw new TypeError('toPromise: Async required')
  }
}
```
While you the type usually come in as `require('./')`, due to the work started by you and @bennypowers around modules, we should start being very explicit about file names and extensions going forward. You do not have to do it for this PR mind you, just something to keep in mind as we move forward.

src/pointfree/index.js
```diff
@@ -39,6 +39,7 @@ module.exports = {
   snd: require('../Pair/snd'),
   swap: require('./swap'),
   tail: require('./tail'),
+  toPromise: require('./toPromise'),
```
evilsoft: This will need to point to `'../Async/toPromise.js'`, when it is moved over to the `Async` folder.

src/pointfree/index.spec.js
```diff
@@ -42,6 +42,7 @@ const sequence = require('./sequence')
 const snd = require('../Pair/snd')
 const swap = require('./swap')
 const tail = require('./tail')
+const toPromise = require('./toPromise')
```
evilsoft: This will need to point to `'../Async/toPromise.js'`, when it is moved over to the `Async` folder.

src/pointfree/toPromise.spec.js
```diff
@@ -0,0 +1,37 @@
+const test = require('tape')
```
evilsoft: This should be moved over to the `Async` folder with the `toPromise` function.

src/pointfree/toPromise.js
dalefrancis88: Haha this is funny, thats where i had it originally using isSameType and then changed it because i thought you'd want it like the other ones like run or map :)

evilsoft: if it was a generic pointfree function, then I would say this is the proper inplementation. You totally nailed it. But it is specific to one type then the pattern should be using isSameType like [here](https://github.com/evilsoft/crocks/blob/master/src/Async/race.js#L10)

docs/src/pages/docs/crocks/Async.md
```diff
@@ -1257,6 +1257,69 @@ timeout(slow)
 
 ## Transformation Functions
 
+#### asyncToPromise
```
evilsoft: Need to add `asynctoproimse` to the functions collection at the top of this file

docs/src/pages/docs/crocks/Async.md
```diff
@@ -1257,6 +1257,69 @@ timeout(slow)
...
+
+`crocks/Async/asyncToPromise`
+
+```haskell
+asyncToPromise :: Async e a -> Promise a e
```
evilsoft: Need to provided the dual signature here. like the examples below this function

docs/src/pages/docs/functions/transformation-functions.md
```diff
@@ -138,6 +138,7 @@ bad
 | Transform | ADT signature | Function Signature | Location |
 |---|:---|:---|:---|
 | `arrayToList` | `[ a ] -> List a` | `(a -> [ b ]) -> a -> List b` | `crocks/List` |
+| [`asyncToPromise`][async-promise] | `Async e a -> Promise a e` | `crocks/Async` |
```
evilsoft: Need to add the function siggy here, like the other transformations.

src/Async/asyncToPromise.js
```diff
@@ -0,0 +1,26 @@
...
+  if(!isSameType(Async, m)) {
+    throw new TypeError('asyncToPromise: Async or a function returning an Async required')
+  }
+
+  return m.toPromise()
+}
+
+// asyncToPromise :: Async e a -> Promise a e
+// asyncToPromise :: (a -> Async e b) -> a -> Promise b e
+function asyncToPromise(async) {
```
evilsoft: probably do not want to use the reserved word `async` here, typically we use `m` for an ADT.

src/Async/asyncToPromise.js
```diff
@@ -0,0 +1,26 @@
...
+  return isFunction(async) ?
+    x => toPromise(async(x)) :
```
evilsoft: Mostly a matter of style, and have not found a rule yet in lint, but could you change this up to read:
```js
return isFunction(m)
  ? x => toProimse(m(x))
  : toPromise(m)
```

src/Async/asyncToPromise.js
dalefrancis88: updated the siggy

docs/src/pages/docs/crocks/Async.md
dalefrancis88: added the new sig

docs/src/pages/docs/crocks/Async.md
dalefrancis88: done

PR #297, opened 2018-07-23
Add Tuple documentation

evilsoft:
## The Product of Types

image

This PR is the final outstanding portion for the upcoming anticipated release and has all the Documentation changes/additions from the addition of the new `Tuple type`. For the most part, this PR does the following:
* Remove `merge` from the `Pair` portion of the docs as it is now used for both `Pair` and `Tuple`
* Update code references to `crocks/Pair/merge` in all the code examples in the docs.
* Fix a search/replace mistake from a previous PR. `(twoals -> equals)`.
* Add `Tuple` documentation.
* Add `Tuple` links to the `crocks` index page to the main article and functions.
* Add `merge` and `nmap` to the search list for the pointfree page.
* Update the path for `merge` to read the general `crocks/pointfree` on the pointfree index page
* Add links to `Tuple` pointfree targets on the pointfree index page
* Remove old links to `merge` on the pointfree index page
* Add entry and link for `nmap` in the `Tuple` folder on the pointfree index page.

### DIFF
--- a/docs/src/pages/docs/crocks/Tuple.md
+++ b/docs/src/pages/docs/crocks/Tuple.md
@@ -212,14 +212,11 @@ Semigroup s1, s2 => 2-Tuple s1 s2 ~> 2-Tuple s1 s2 -> 2-Tuple s1 s2
 ...
 ```
 
-When a given `n-Tuple` contains `Semigroup` instances in all defined `n`
-positions. It can be concatenated with another `n-Tuple` instance containing
-instances of the same `Semigroup`s in the same positions as the `n-Tuple`
-instance it is being concatenated with.
-
-`concat` will be called on each contained `Semigroup` instance with the instance
-corresponding to the other `n-Tuple` instance. The result of each concatenation
-will be provided in a new `n-Tuple` instance of the same `Semigroup`s.
+Two instances of the same `n-Tuple` can be combined using concatenation, as
+long as both have instances of the same `Semigroup`s in the same
+position. `concat` will be called on each contained `Semigroup` instance with
+the instance corresponding to the other `n-Tuple` instance. The result of each
+concatenation will be provided in a new `n-Tuple` instance.
 
 ```javascript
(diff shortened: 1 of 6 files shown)


docs/src/pages/docs/crocks/Tuple.md
```diff
@@ -0,0 +1,480 @@
...
+#### concat
+
+```haskell
+Semigroup s1 => 1-Tuple s1 ~> 1-Tuple s1 -> 1-Tuple s1
+Semigroup s1, s2 => 2-Tuple s1 s2 ~> 2-Tuple s1 s2 -> 2-Tuple s1 s2
+...
+```
+
+When a given `n-Tuple` contains `Semigroup` instances in all defined `n`
+positions. It can be concatenated with another `n-Tuple` instance containing
```
HenriqueLimas: IMHO, little change but should be better to use a `,` instead of `.` here. It makes the read more easy.
> When a given `n-Tuple` contains `Semigroup` instances in all defined `n`
> positions, it can be concatenated with another `n-Tuple`...

docs/src/pages/docs/crocks/Tuple.md
```diff
@@ -0,0 +1,480 @@
...
+2-Tuple a b ~> (b -> c) -> 2-Tuple a c
+...
+```
+
+Used to lift a single function into a given `n-Tuple` to map the rightmost
+value. `map` takes a function `(a -> b)` and will return a
+new `n-Tuple` instance with the result of mapping the rightmost value from its
+original `a` type to the resulting `b`.
+
+This method will only apply to the rightmost value, if the need to map one or
```
HenriqueLimas: Here we can use the `.` instead of `,`. And also I think we can restructure this a bit something like:
> This method will only apply to the rightmost value. 
> [`mapAll`](#mapall) can be used to map one or many
> of the values in the various other positions.

evilsoft:
Gonna merge this.
image

PR #298, opened 2018-07-23
Support inspecting Symbols

### LINKED ISSUE #296: inspect code fails when Symbol provided
Example failure:
```js
const { Right } = require('crocks/Either')
Right(Symbol('42'))
```
gives the error
```
/path/to/node_modules/crocks/core/inspect.js:39
  return (" " + x)
              ^

TypeError: Cannot convert a Symbol value to a string
    at inspect (/path/to/node_modules/crocks/core/inspect.js:39:15)
    at /path/to/node_modules/crocks/Either/index.js:72:41
    at /path/to/node_modules/crocks/core/defineUnion.js:26:25
    at Object.caseOf (/path/to/node_modules/crocks/core/curry.js:25:12)
    at either (/path/to/node_modules/crocks/Either/index.js:81:20)
    at Either (/path/to/node_modules/crocks/Either/index.js:70:5)
    at /path/to/node_modules/crocks/core/compose.js:8:12
    at Object.<anonymous> (/path/to/index.js:5:3)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
```
when I think it should return
```js
Right Symbol('42')
```

Here is what I believe a simple fix is, though it'd be pulled out to `ifSymbol(x)`:
```js
if (typeof x === "symbol") {
  return " " + x.toString()
}
```

rpearce:
Closes #296


evilsoft:
This looks pretty awesome.
What do you think about exposing `isSymbol` to the outside world as new predicate function?

If so it will have to be added to the predicate folder, and brought in from the core implementation. Then it would need to be exposed to the main crocks index through predicates index and onto the index inside of `src/`?

src/core/isSymbol.js
```diff
@@ -0,0 +1,9 @@
+/** @license ISC License (c) copyright 2016 original and current authors */
```
evilsoft: This should be 2018

rpearce: Hmmm I amended a commit that I don't think pushed. Will redo

rpearce:
@evilsoft I think that's a fair thing to expose

evilsoft:
This is GTG. Merging and please by all means, have a squirrel:

image

PR #300, opened 2018-07-25
Fix up the Fantasy Land `bimap` method on `Pair`

evilsoft:
Looks like we missed one of the fantasy land methods on `Pair`.
This PR just fixes up the method so it properly applies the method name for error reporting.


evilsoft:
image

Merging for quick release!

rpearce:
:100: for the testing

Issue #302, opened 2018-07-29
Add some additional logic and predicates from live code sesh


evilsoft:
After doing a live code sesh, found some value in adding the following:

- [x] `logic/implies :: (a -> Boolean) -> (a -> Boolean) -> a -> Boolean`
- [x] `predicates/propSatisfies :: (a -> Boolean) -> String -> Object -> Boolean`
- [x] `predicates/propPathSatisfies :: (a -> Boolean) -> [ String ] -> Object -> Boolean`

evilsoft:
This has been released, closing!

PR #303, opened 2018-08-04
Add `implies` Logic Combinator

evilsoft:
## What are you implying?
image

This addresses the first item from [this issue](https://github.com/evilsoft/crocks/issues/302) and adds the `implies` logic combinator and its documentation.


evilsoft:
image

PR #305, opened 2018-08-06
Add `propSatisfies` and `propPathSatisfies` to the Predicates

evilsoft:
## Packed with peanuts
image

Addressing the last (2) items on [this issue](https://github.com/evilsoft/crocks/issues/302), this PR add the predicate functions: `propSatisfies` and `propPathSatisfies` that allows one to check a value on an object to see if it passes a provided predicate (function or Pred Type).

### DIFF
--- a/src/predicates/propPathSatisfies.js
+++ b/src/predicates/propPathSatisfies.js
@@ -3,7 +3,6 @@
 
 const curry = require('../core/curry')
 const isArray = require('../core/isArray')
-const isDefined = require('../core/isDefined')
 const isEmpty  = require('../core/isEmpty')
 const isInteger = require('../core/isInteger')
 const isNil = require('../core/isNil')
@@ -45,10 +44,6 @@ function propPathSatisfies(keys, pred, x) {
     }
 
     target = target[key]
-
-    if(!isDefined(target)) {
-      return false
-    }
   }
 
   return !!predOrFunc(pred, target)
--- a/src/predicates/propPathSatisfies.js
+++ b/src/predicates/propPathSatisfies.js
@@ -1,0 +1,57 @@
+/** @license ISC License (c) copyright 2018 original and current authors */
+/** @author Ian Hofmann-Hicks (evilsoft) */
+
+const curry = require('../core/curry')
+const isArray = require('../core/isArray')
+const isDefined = require('../core/isDefined')
+const isEmpty  = require('../core/isEmpty')
+const isInteger = require('../core/isInteger')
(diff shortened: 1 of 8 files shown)


src/predicates/propPathSatisfies.js
```diff
@@ -0,0 +1,57 @@
...
+      )
+    }
+
+    if(isNil(target)) {
+      return false
+    }
+
+    target = target[key]
+
+    if(!isDefined(target)) {
```
karthikiyengar: Should we really be defaulting to `false` in such cases? 

I understand that it's an edge case, but there are certain libs (mainly ORMs) that treat `undefined` as a legit value. The same goes for `hasPropPath` - should we consider piping `undefined` to the pred instead?

src/predicates/propPathSatisfies.js
```diff
@@ -0,0 +1,57 @@
...
+    )
+  }
+
+  if(!isPredOrFunc(pred)) {
+    throw new TypeError(
+      'propPathSatisfies: Pred or predicate function required for second argument'
+    )
+  }
+
+  if(isNil(x)) {
```
karthikiyengar: Should we be throwing in this case? `if (!isObject(x)) { throw new TypeError( 'propPathSatisfies: Object required for third argument') } `

evilsoft: No. because the path does not exist.
AND Predicates do not throw in crocks (or shouldn't) they either are or are not.

evilsoft:
For predicates I think that is fine to add prop based functions.
For helpers though, I think we should hold off.
That is a great function though!

src/predicates/propPathSatisfies.js
evilsoft: Okay, after fiddling with ways to get this to make sense, there is no way to tell if the key exists with an `undefined` value when the path is the same depth of the target path. This was the same problem we ran into with the other path functions.

@karthikiyengar
I say we keep it in until people start reporting their specific edge cases with the specific libs they are using and address them on a case-by-case basis. we could check `Object.keys` but I do not want to add extra cycles until a lot of users need it.

karthikiyengar: How about this before `target = target[key]`?

```
if (target.hasOwnProperty(key) && !target[key]) {
  return !!predOrFunc(pred, target[key])
}
```

evilsoft:
Okay @karthikiyengar, just removed the defined check.
This should be GTG now.

evilsoft:
image

PR #310, opened 2018-08-17
Add `foldMap` pointfree and `List`/`Array` implementations

evilsoft:
image

After being pointed to a @drboolean [egghead lesson](https://egghead.io/lessons/javascript-real-world-example-pt3) by @michael-wolfenden in the [gitter chat](https://gitter.im/crocksjs/crocks), it was painfully clear that we needed this function so we can do things like:

```js
const { Sum, foldMap, isNumber, safeLift } = require('../crocks')

// fn :: a -> Maybe Sum
const fn =
  safeLift(isNumber, Sum)

foldMap(fn, [ 1, 2, 4, 3 ])
//=> Just Sum 10

foldMap(fn, [ 1, '2', 4, 3 ])
//=> Nothing
```

This PR just add a pointfree function that will call `foldMap` on any container that has it available. Also implemented it on `List` and the `Array` implementations.

Also noticed that we implemented `fold` for `Array` in the pointfree version which is Bad News Bears :bear: as it goes against the pattern we established after its initial implementation. So I moved that over as part of this PR.

Thanks @michael-wolfenden for the nudge in the proper direction.


evilsoft:
image

PR #311, opened 2018-08-19
Seperate function refs in the main index

evilsoft:
## Gotta keep em separated
egg

This PR moves around the references to better reflect how these references will be used when we move to a mono-repo.

No code what changed, just references to where things live.


evilsoft:
This is not breaking.
All it does is remove how the references were added to the main API.
None of the file locations changed. Now it *could* be breaking if we ever communicates something like:
```js
const helpers = require('crocks/helpers')
```

which would being in the index file, but that was never officially supported or communicated as officially supported. So in terms of how we communicated the API to be used this is non-breaking. It was just a fluke that this made it in and was never intended to behave this way

HenriqueLimas:
Perfect explanation. I was just wondering if someone was using in that way. But since we do not communicate that as you said, that is perfect. Thanks 👍

evilsoft:
You know what though @HenriqueLimas, there is no reason for us to NOT mark this as breaking, I mean technically it is not, BUT I say because someone decided to use that pattern does not mean we throw them in the penalty box.
We are going to release a major breaking change after this next one, So I say we throw a breaking label on this and send it out with the breaking changes.

Good call! Thanks for noticing that!

evilsoft:
Merging
image

Issue #318, opened 2018-10-02
tryCatch should work with nAry arguments


karthikiyengar:
Currently, `tryCatch` always returns unary functions even if the input passed to it is nAry.

We can change this behaviour with the proposed implementation, if this doesn't cause any breaking problems.

```javascript
const tryCatch =  function tryCatch(fn) {
  if(!isFunction(fn)) {
    throw new TypeError('tryCatch: Function required for first argument')
  }

  const safe = function(...args) {
    try { return Ok(fn(...args)) }
    catch(e) { return Err(e) }
  }

  Object.defineProperty(safe, "length", { value: fn.length })
  
  return safe
}


const add = tryCatch((a, b) => {
  return a + b
})

add(1, 2) => 3
curry(add)(1)(2) => 3
```

evilsoft:
Think this is a good idea. Good call.
Is that defineProperty going to work consistently across contexts on a function?

karthikiyengar:
I don't see problems with the `defineProperty` itself, but the spread operator is going to destroy context. Edited to use the old-school way.

karthikiyengar:
In hindsight, I don't think that the `arguments` change is going to make a major difference. Looks like our curry always causes functions to lose context. Is this by design?

```
const inc = function() { return ++this.a }
const incCurried = curry(inc)

const obj = {
  a: 0,
  inc,
  incCurried,
}

obj.inc() // => 1
obj.incCurried() // => NaN
```

PR #320, opened 2018-10-05
Convert tryCatch to nAry (#318)

karthikiyengar:


karthikiyengar:
Sure, I'll get that in today.

src/Result/tryCatch.spec.js
```diff
@@ -59,5 +59,13 @@ test('tryCatch functionality', t => {
   t.equals(good, 'Ok', 'returns an Ok when no error')
   t.equals(bad.message, msg, 'returns an Err with error on error')
 
+  const x = (a, b, c) => ({ a, b, c })
+
+  const noncurried = tryCatch(x, 1,2,3).either(identity, identity)
+  const curried = tryCatch(x, 1)(2)(3).either(identity, identity)
+
+  t.equals(equals(noncurried, { a: 1, b: 2, c: 3 }), true, 'preserves the arity of the passed function')
```
evilsoft: instead of equals with an equals function checking for true, tape provides `same`.
But this is okay, I have done this a lot myself and it still is a vaild test

PR #321, opened 2018-10-06
Update Dependencies

evilsoft:
image

We were a little outdated on some deps and it was not only causing a couple vulnerabilities to slip into our Dev stack, but also made it so node:6 specs would not run in CI.

This PR updates those all deps (except for `husky` and `nodemon`, still waiting on some things to be fixed up in those packages).


evilsoft:
image

PR #322, opened 2018-10-07
Add `setPath` helper function

evilsoft:
## The path less traveled
image

This PR adds a function that adds the functionality from the `assocPath` function provided by `ramda`. This was motivated by the documentation on the [meiosis project](https://github.com/foxdonut/meiosis) from @foxdonut.

This PR signals a change to some of the existing functions to match this new format for dealing with prop and propPath. The following functions will be addressed in the next breaking change release, although they will be deprecated and their original names will still be available until we hit `1.0.0`:
* `Maybe/propPath` -> `Maybe/getPath`
* `Maybe/prop` -> `Maybe/getProp`
* `helpers/assoc` -> `helpers/setProp` (will work with `Array` as well)
* `helpers/dissoc` -> `helpers/unsetProp` (will work with `Array` as well)
* `helpers/propOr` -> `helpers/getPropOr`
* `helpers/propPathOr` -> `helpers/getPathOr`


src/helpers/setPath.js
```diff
@@ -0,0 +1,56 @@
...
+
+  if(!isValid(obj)) {
+    throw new TypeError('setPath: Object or Array required for third argument')
+  }
+
+  const key = path[0]
+  let newVal = val
+
+  if(path.length > 1) {
+    const next = isNil(obj[key]) || !isValid(obj[key])
```
karthikiyengar: The `isNil` check looks redundant to me - I can't think of any case that isn't covered by `isValid`. Am I missing something?

src/helpers/setPath.js
```diff
@@ -0,0 +1,56 @@
...
+const isString = require('../core/isString')
+const object = require('../core/object')
+
+const isValid = x =>
+  isObject(x) || isArray(x)
+
+const pathError =
+  'setPath: Non-empty Array of non-empty Strings and/or Integers required for first argument'
+
+// setPath :: [ String | Number ] -> a -> Object -> Object
```
karthikiyengar: Just nitpicking at this point, but the signature is different from the docs.

src/helpers/setPath.js
evilsoft: Good catch!

PR #323, opened 2018-10-07
Add `unsetPath` helper function

evilsoft:
## A path a little less traveled

image

This PR adds a function that provides the functionality from the `dissocPath` function provided by `ramda`. This was motivated by the documentation on the [meiosis project](https://github.com/foxdonut/meiosis) from @foxdonut.

This PR signals a change to some of the existing functions to match this new naming convention for dealing with functions like: `prop` and `propPath`. The following functions will be addressed in the next breaking change release, although they will be deprecated and their original names will still be available until we hit `1.0.0`:
* `Maybe/propPath` -> `Maybe/getPath`
* `Maybe/prop` -> `Maybe/getProp`
* `helpers/assoc` -> `helpers/setProp` (will work with `Array` as well)
* `helpers/dissoc` -> `helpers/unsetProp` (will work with `Array` as well)
* `helpers/propOr` -> `helpers/getPropOr`
* `helpers/propPathOr` -> `helpers/getPathOr`

### DIFF
--- a/docs/src/pages/docs/functions/helpers.md
+++ b/docs/src/pages/docs/functions/helpers.md
@@ -2,7 +2,7 @@
 title: "Helpers"
 description: "Helper functions"
 layout: "notopic"
-functions: ["assign", "assoc", "binary", "compose", "composek", "composep", "composes", "curry", "defaultprops", "defaultto", "dissoc", "fanout", "frompairs", "lifta2", "lifta3", "liftn", "mapprops", "mapreduce", "mconcat", "mconcatmap", "mreduce", "mreducemap", "nary", "objof", "omit", "once", "partial", "pick", "pipe", "pipek", "pipep", "pipes", "propor", "proppathor", "tap", "unary", "unit"]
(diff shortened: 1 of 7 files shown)


docs/src/pages/docs/functions/helpers.md
```diff
@@ -983,11 +983,50 @@ applies (2) arguments to a given function.
...
+`crocks/helpers/unsetPath`
+
+```haskell
+unsetPath :: [ String | Integer ] -> (Object | Array) -> (Object | Array)
+```
+
+Used to remove a property or index on a deeply nested `Object`/`Array`.
+`unsetPath` is will return a new instance with the property or index removed.
+
+The provided path can be a mixture of either `Integer`s or `String`s to allow
```
karthikiyengar: Looks like we've taken a different direction from Ramda here?

```javascript
dissocPath([1], { 1: {} } ) // []
unsetPath([1], { 1: {} } ) // { '1': {} }
```

karthikiyengar: Makes sense. Thanks for the explanation.

evilsoft:
image

PR #324, opened 2018-10-14
Add `Either` docs

evilsoft:
## This or That?
image

This PR adds the Either docs as requested in [this issue](https://github.com/evilsoft/crocks/issues/42).
Nothing too fancy, Just the same ol' stuff as the other docs PRs.

EDIT: Although, I did sneak in a few documentation :lipstick: and :nail_care: as I found them.


docs/src/pages/docs/crocks/Async.md
```diff
@@ -1266,17 +1266,19 @@ eitherToAsync :: Either b a -> Async b a
...
+original [`Right`][right]. If a [`Left`][left] is provided,
+then `eitherToAsync` will return a [`Rejected`](#rejected) instance, wrapping
+the original [`Left`][left] value.
 
 Like all `crocks` transformation functions, `eitherToAsync` has two possible
-signatures and will behave differently when passed either an `Either` instance
-or a function that returns an instance of `Either`. When passed the instance,
-a transformed `Async` is returned. When passed an `Either` returning function,
-a function will be returned that takes a given value and returns an `Async`.
+signatures and will behave differently when passed either
```
HenriqueLimas: IMHO there are to much either here 😄 , I think we can remove the "when passed **either**" here

docs/src/pages/docs/crocks/Either.md
```diff
@@ -0,0 +1,1346 @@
...
+it provides either a `Left` value or a `Right` value and not both.
+
+Unlike [`Maybe`][maybe], which fixes its "left side", or [`Nothing`][nothing] to
+a `()` (unit), `Either` is a functor in both it's `Left` and `Right` sides. This
+allows for the ability to vary the type on either side and is akin to the
+imperative `if...else` trees that are common in programming.
+
+Like most other types in `crocks`, `Either` has a right bias in regard to
+instance methods like `map`, `ap` and `chain`. This behavior can be used to
+formally capture Error handling as the `Left` value will be maintained and most
```
HenriqueLimas: The `most` here is a little confusing. Is it `most of the future` or `most important` or maybe I misunderstood 😄. Maybe we could remove it.

docs/src/pages/docs/crocks/Either.md
```diff
@@ -0,0 +1,1346 @@
...
+
+```haskell
+Either.Left :: c -> Either c a
+```
+
+Used to construct a `Left` instance of an `Either` that represents the
+`false` portion of a disjunction. The `Left` constructor takes a value of any
+type and returns a `Left` instance wrapping the value passed to the constructor.
+
+When an instance is a `Left`, most other methods `Either` returning methods on
```
HenriqueLimas: `most other methods Either returning methods` I think should be `other Either returning methods`

docs/src/pages/docs/crocks/Either.md
```diff
@@ -0,0 +1,1346 @@
...
+
+```haskell
+Either c a ~> (a -> Either c b) -> Either c b
+```
+
+Combining a sequential series of transformations that capture disjunction can be
+accomplished with `chain`. `chain` expects a unary, `Either` returning function
+as its argument. When invoked on a `Left`, `chain` will not run the function,
+but will instead return another `Left` wrapping the original value. When called
+on a `Left` however, the inner value will be passed to provided function,
```
HenriqueLimas: Maybe: "the inner value will be passed to **the** provided function"

docs/src/pages/docs/crocks/Either.md
```diff
@@ -0,0 +1,1346 @@
...
+```
+
+#### either
+
+```haskell
+Either c a ~> ((c -> b), (a -> b)) -> b
+```
+
+Used as a means to map and extract a value from an `Either` based on the
+context. `either` takes two functions as its arguments. The first will map any
```
HenriqueLimas: comma instead of point after `context` maybe? Just to maintain the same style of the other ones.

docs/src/pages/docs/crocks/Either.md
```diff
@@ -0,0 +1,1346 @@
...
+firstToEither :: c -> (a -> First b) -> a -> Either c a
+```
+
+Used to transform a given [`First`][first] instance to an `Either`
+instance, `firstToEither` will turn a non-empty instance into
+a [`Right`](#right) wrapping the original value contained within
+the [`First`][first].
+
+The [`First`][first] datatype is based on a [`Maybe`][maybe] and as such its
+is fixed to a `()` type. As a means to allow for convenient
```
HenriqueLimas: can we add here `(unit)` after `()`, the same we did before.

docs/src/pages/docs/crocks/Either.md
```diff
@@ -0,0 +1,1346 @@
...
+lastToEither :: c -> (a -> Last b) -> a -> Either c a
+```
+
+Used to transform a given [`Last`][last] instance to
+an `Either`, `lastToEither` will turn a non-empty [`Last`][last] instance into
+a [`Right`](#right) instance wrapping the original value contained in the
+original non-empty.
+
+The [`Last`][last] datatype is based on a [`Maybe`][maybe] and as such its left
+or empty value is fixed to a `()` type. As a means to allow for convenient
```
HenriqueLimas: Can we add `(unit)` here too after `()`?

Issue #278, opened 2018-05-30
Const and its Applicative constraints.


neonphlux:
Hello.

Firstly, I would like to thank you for providing this library, it's lovely.

I'm learning Javascript so maybe you are already aware of this and haven't implemented it for reasons that are beyond my knowledge. 
I noticed that the Const doesn't behave as expected, since it doesn't 'enforce' the monoid contraint. For the Const to be able to be an Applicative, it has to provide a pure/return function:

```
class Functor f => Applicative f where
  pure :: a -> f a
  ...

``` 

`data a b = Const a`

Looking at "pure"'s signature, we observe that is impossible to provide a value for the monoid, since it only accepts one parameter. But by enforcing the first type variable of Const to be a monoid, we can conjure a value for Const out of nothing, by just providing a monoid type.

` repl> pure "hello" :: Const <which monoid?> String`

But with monoid:

```
repl> pure "hello" :: Const (Sum Int) String
repl> Const (Sum 0) // Sum Int! I know how to provide you with the empty value of that!

```
That also allows gives us other abilities like:

```
repl> liftA3(_ => _ => _ => undefined, Const(10), Const(30), Const(50))
repl> Const (90)
```

Again, thank you so much for the open-source.

evilsoft:
So in the beginning I went back and forth on this. At the time we did not really have a notion of Type Constructors (this was before Writer, and please do not look at that one, there will be a bunch of breaking changes going out on that after the FL stuff is finished). So at the time I did not want to limit it to just work with the Monoid 🌽straint.

Also we did not have `First` implemented, so it made sense to make `Const` more like `Constant` and allow it to act as a poor-person's `First` and just lock down to a value. I figured that I would address this when the time came.

With this issue, it appears that the time has come. 😄.

So as the person making the request, how do you think we should implement, here are a couple options (and I am way open to more), because of the need for `of` and `empty` we will need to def go the type constructor route:

* Lock `Const` down to always use a `Monoid`  by taking a `Monoid` TypeRep. If a non-Monoid TypeRep is provided we throw a TypeError.

* Allow for all TypeReps (including constructors of JS types like `Array`, `Object`, `String`) and if `of`, `ap`, `empty` or `concat` is used we Throw a type error. (although we could probably allow `concat` for `Semigroups`, but that is debatable.

In both of the solutions we will be getting rid of `chain`, that part was from my lack of knowledge at the time.

evilsoft:
[...]

PR #327, opened 2018-10-20
Make `Const` a "true" Applicative

evilsoft:
## The only constant is change
image

This PR address [this issue](https://github.com/evilsoft/crocks/issues/278) and changes the implementation of the `Const` type to act more like the delta functor it is based on. In order to do that, this PR proposes the following changes (mostly breaking):
* Initial constructor is a type constructor that requires a crocks instance constructor or a JS constructor. This will provide an instance constructor that will be fixed to the provided type.
* Remove `chain` method, made no sense to begin with
* Add `empty` and `of` construction helpers to instance constructor. These will only apply when the type is fixed to a `Monoid`
* change behavior of `ap` to concat values when type is fixed to a `Semigroup`


docs/src/pages/docs/crocks/Const.md
```diff
@@ -9,65 +9,76 @@ weight: 30
 Const c a
 ```
 
-Const is a Product type the whose underlying left-most value is fixed to the
-value it was originally constructed with. This ensures that a desired value is
-immutable. While its right portion can still be mapped over, when observed, all
-information on the right will be discarded, leaving only the initial fixed
-value `c`.
+Commonly known as the Delta (Δ) Functor, `Const` is a Product type the whose
```
dalefrancis88: I think this is a much nicer description. Good job

PR #330, opened 2018-10-22
Documentation for Identity

dalefrancis88:
This is the initial commit for the `Identity` documentation for #42


docs/src/pages/docs/crocks/Identity.md
```diff
@@ -0,0 +1,274 @@
...
+
+```haskell
+Identity (a -> b) ~> Identity a -> Identity b
+```
+
+`ap` allows for values wrapped in a `Identity` to be applied to functions also
+wrapped in a `Identity`. In order to use `ap`, the `Identity` must contain a
+function as its value. Under the hood, `ap` unwraps both the function
+and the value to be applied and applies the value to the function. Finally it
+will wrap the result of that application back into a `Identity`. It is required
```
evilsoft: a `Identity` -> an `Identity`

docs/src/pages/docs/crocks/Identity.md
```diff
@@ -0,0 +1,274 @@
...
+#### sequence
+
+```haskell
+Apply f => Identity (f a) ~> (b -> f b) -> f (Identity a)
+Applicative f => Identity (f a) ~> TypeRep f -> f (Identity a)
+```
+
+When an instance of `Identity` wraps an `Apply` instance, `sequence` can be used to
+swap the type sequence. `sequence` requires either an `Applicative TypeRep` or
+an `Apply` returning function is provided for its argument.
```
evilsoft: is provided -> to be provided
(may need to fix this in a couple places if this was copied)

docs/src/pages/docs/crocks/Identity.md
```diff
@@ -0,0 +1,274 @@
...
+
+#### chain
+
+```haskell
+Identity a ~> (a -> Identity b) -> Identity b
+```
+
+Normally one of the ways `Monad`s like `Identity` are able to be combined and 
+have their effects applied is through `chain`. However `Identity` is different
+because there are no effects to apply. `chain` will simply take a func that
```
evilsoft: func -> function

docs/src/pages/docs/crocks/Identity.md
```diff
@@ -0,0 +1,274 @@
...
+description: "Identity Crock"
+layout: "guide"
+weight: 60
+---
+
+```haskell
+Identity a
+```
+
+`Identity` is one of the most versatile `monads`. Although it does not have any
```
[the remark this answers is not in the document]
dalefrancis88: Haha all good. Never thought the roles would be reversed and you'd tell me i was being too theoretical :)

docs/src/pages/docs/crocks/Identity.md
dalefrancis88: This came from `Maybe`. After this commit i'll make that channge

docs/src/pages/docs/crocks/Identity.md
```diff
@@ -0,0 +1,344 @@
...
+```
+
+Used to apply the "effect" of an `Apply` to a value inside of a `Identity`,
+`traverse` combines both the "effects" of the `Apply` and the `Identity` by
+returning a new instance of the `Apply`, wrapping the result of the
+`Apply`s "effect" on the value in the `Identity`.
+
+`traverse` requires either an `Applicative TypeRep` or an `Apply` returning
+function as its first argument and a function that is used to apply the "effect"
+of the target  `Apply` to the value inside of the `Identity`. Both arguments must provide
```
HenriqueLimas: Extra space after `target`

dalefrancis88: damn man, nice 👀!

Issue #329, opened 2018-10-21
Rename `prop` and `propPath` related functions


evilsoft:
In order for better :corn:sistancy, it makes sense to move to more explicit names for some of the prop, propPath based functions. With the merger of [this](https://github.com/evilsoft/crocks/pull/322) and [this](https://github.com/evilsoft/crocks/pull/323), it makes sense to go with the `get`, `set`, `unset` convention and allow us to drop the prefixing of `prop` to `path` based functions.

All existing functions will remain, but will be deprecated and completely removed at version `1.0.0`. The code should exist in one location and have the deprecated function reference the new version, The functions documentation should be moved to the new name for the description and code examples. The index views should point to the new function name, with some indication of its deprecation.

The following function are to be affected by this work:
- [x] `helpers/assoc` -> `helpers/setProp`
- [x] `helpers/dissoc` -> `helpers/unsetProp`
- [x] `helpers/propOr` -> `helpers/getPropOr`
- [x] `helpers/propPathOr` -> `helpers/getPathOr`
- [x] `Maybe/prop` -> `Maybe/getProp`
- [x] `Maybe/propPath` -> `Maybe/getPath`
- [x] `predicates/propPathSatisfies` -> `predicates/pathSatisfies`
- [x] `predicates/propPathEq` -> `predicates/pathEq`

evilsoft:
These are all complete and shipped to production as part of `0.12.0`
Closing.

PR #331, opened 2018-10-22
Move `assoc` to be `setProp`

evilsoft:
## PROPS
image

This PR adds a new `setProp` function that implements was `assoc` was, with the exception that it will work with `Array` as well as `Object`. Also this is sets the pattern for how we may want to do deprecation in this lib. 

This is the first part of [this issue](https://github.com/evilsoft/crocks/issues/329)


docs/src/pages/docs/functions/helpers.md
```diff
@@ -979,6 +963,46 @@ setPath([ 'list', 'a' ], 'ohhh, I see.', { list: [ 'string', 'another' ] })
...
+```haskell
+setProp :: (String | Integer) -> a -> (Object | Array) -> (Object | Array)
+```
+
+Used to set a given value for a specific key or index of
+an `Object` or `Array`. `setProp` takes either a `String` or `Integer` value
+as its first argument and a value of any type as its second. The third parameter
+is dependent of the type of the first argument. When a `String` is provided, the
+third argument must be an `Object`. Otherwise if the first argument is
+an `Integer`, then the third must be an `Array`.
```
dalefrancis88: Is this behaviour what we want, javascript has an interesting feature where keys are forced to be strings regardless of whether they're a number or not. This prevents a scenario where you've created a dictionary for fast access that could be based on an int id where the input is then given as an int. Could we do it the other direction? if it's an array it'll try to make the first arg a number and use is, same for object/string?

src/helpers/setProp.js
```diff
@@ -0,0 +1,50 @@
...
+const array = require('../core/array')
+const curry = require('../core/curry')
+const isArray = require('../core/isArray')
+const isInteger = require('../core/isInteger')
+const isObject = require('../core/isObject')
+const isString = require('../core/isString')
+const object = require('../core/object')
+
+function fn(name) {
+  function setProp(key, val, x) {
```
dalefrancis88: See above comment

src/helpers/assoc.js
```diff
@@ -1,21 +1,7 @@
...
-  }
-
-  if(!isObject(obj)) {
-    throw new TypeError('assoc: Object required for third argument')
-  }
-  return object.assign({ [key]: val }, obj)
-}
-
-module.exports = curry(assoc)
+module.exports =
```
dalefrancis88: Although i understand that this will create something none pure, do we want to do some console logging here to inform the dev?

evilsoft: Well no, I do not think that would make sense as we do not allow for dev vs prod implementations. If we log a warning that means that we are polluting their production space with development messages.

dalefrancis88:
Lol what was the 🌽cern i had? i forget now 😅

docs/src/pages/docs/functions/helpers.md
evilsoft: Hmm. let me think about that one for a bit.
My knee jerk reaction is to keep it explicit as the user should know what they want. OR do not include `Array` as a valid type and keep it as is, only using Array for `Path` functions

evilsoft: Nah. I think we should be good with doing it how you suggest.
will update.

evilsoft:
Okay @dalefrancis88 addressed that 🌽cern with Array and moved back to having the functions ONLY handle `String`/`Object`. That means that this is no longer a breaking change, as everything still works like it did with `assoc`.

Ready for some additional 👀

src/helpers/assoc.js
dalefrancis88: yeah fair point

dalefrancis88:
for the 👀 did you mean you want me to take another look or waiting for someone else to review as well?

evilsoft:
@dalefrancis88 other :eyes: would be always welcome, but mostly to make sure that the changes I made are in line with accepted expectations in response to your PR comments.

evilsoft:
And if good, change your review from Comment to Approved.

src/helpers/setProp.js
```diff
@@ -0,0 +1,36 @@
...
+    if(isString(key)) {
+      if(isObject(x)) {
+        return object.set(key, val , x)
+      }
+
+      throw new TypeError(
+        `${name}: Object required for third argument`
```
karthikiyengar: I see that `setPath` still allows arrays and coerces them into objects. Do you think that it should throw for Objects too, to keep behaviour consistent?

evilsoft: Good 👁!
Are you proposing that if the end of the path is an `Array` we throw?

karthikiyengar: Yeah pretty much, for consistency's sake.

evilsoft:
Okay @karthikiyengar and @dalefrancis88 this has been updated to work with `Arrays` and as such is back to a breaking change.

@karthikiyengar will follow this up with another PR that will add the `Integer`/`Array` :corn:straint to `setPath`.

:red_circle: :eyes: for some more :eyes:

docs/src/pages/docs/functions/helpers.md
```diff
@@ -2,7 +2,7 @@
 title: "Helpers"
 description: "Helper functions"
 layout: "notopic"
-functions: ["assign", "assoc", "binary", "compose", "composek", "composep", "composes", "curry", "defaultprops", "defaultto", "dissoc", "fanout", "frompairs", "lifta2", "lifta3", "liftn", "mapprops", "mapreduce", "mconcat", "mconcatmap", "mreduce", "mreducemap", "nary", "objof", "omit", "once", "partial", "pick", "pipe", "pipek", "pipep", "pipes", "propor", "proppathor", "setpath", "tap", "unary", "unit", "unsetpath"]
+functions: ["assign", "assoc", "binary", "compose", "composek", "composep", "composes", "curry", "defaultprops", "defaultto", "dissoc", "fanout", "frompairs", "lifta2", "lifta3", "liftn", "mapprops", "mapreduce", "mconcat", "mconcatmap", "mreduce", "mreducemap", "nary", "objof", "omit", "once", "partial", "pick", "pipe", "pipek", "pipep", "pipes", "propor", "proppathor", "setpath", "setprop", "tap", "unary", "unit", "unsetpath"]
```
dalefrancis88: Do we want to leave `assoc` in here?

evilsoft: Yep because we have an entry in the table!

evilsoft: I will be doing another PR for dissoc.

Issue #29, opened 2017-01-18
Every function that takes an Array should also allow an Iterable


evilsoft:
So each one of these should also take an Iterable:

### Helpers
- [ ] `mconcat`
- [ ] `mconcatMap`
- [ ] `mreduce`
- [ ] `mreduceMap`
- [ ] `take` **(new)**

## Pointfree
- [ ] `filter`
- [ ] `head`
- [ ] `map`
- [ ] `reduce`
- [ ] `sequence`
- [ ] `tail`
- [ ] `traverse`

Since we are now exposing the possibility of infinite values, makes sense to add a `take` function at this point.

dalefrancis88:
how would this be implemented without changing the type or mutating the value? the value of using the known functions is that they take care of creating the correct type but i think filter/map/reduce won't work without mutation or changing type

I think though most of the others should be fine

dalefrancis88:
hey @evilsoft I am wondering if you have some magical idea for this? cause i can't quite think of how to do many of these without returning an array instead of the original `iterable`?

HenriqueLimas:
I think we need to define a function that `clone` the object in a way that it doesn't loose the structure. We need some checks for base constructors like (Date, Array, String etc..) and clone it in the right way, and when its a normal object use something like Object.create and assign.
Something like (pretty basic without the base constructor checks)
```js
let object = Object.create(Object.getPrototypeOf(source))
Object.assign(object, source)

let symbols = Object.getOwnPropertySymbols(source)
symbols.forEach(key => object[key] = source[key])

return object
```

dalefrancis88:
@HenriqueLimas @evilsoft I think we may have made a mistake on this one when we did head. I am thinking that the behavior we started with in `head` is wrong. I'm thinking that it should be if a generator function is passed in, then we will invoke it and iterate, however if you pass in the result of a generator call then your expectation is that you want it iterated over. 
What do you guys think?

PR #332, opened 2018-10-24
Added an isIterable predicate

dalefrancis88:
This is a PR to add an `isIterable` predicate in preparation for work on #29

This is something we'll obviously need internally, however I am unsure on the value of adding this as something that is part of the api, thoughts?

### DIFF
--- a/src/core/isIterable.js
+++ b/src/core/isIterable.js
@@ -2,9 +2,10 @@
 /** @author Dale Francis (dalefrancis88) */
 
 const isFunction = require('./isFunction')
+const isNil = require('./isNil')
 
 function isIterable(iterable) {
-  return iterable !== null && iterable !== undefined && isFunction(iterable[Symbol.iterator])
+  return !isNil(iterable) && isFunction(iterable[Symbol.iterator])
 }
 
 module.exports = isIterable
--- a/src/core/isIterable.js
+++ b/src/core/isIterable.js
@@ -1,0 +1,10 @@
+/** @license ISC License (c) copyright 2018 original and current authors */
+/** @author Dale Francis (dalefrancis88) */
+
+const isFunction = require('./isFunction')
+
+function isIterable(iterable) {
+  return iterable !== null && iterable !== undefined && isFunction(iterable[Symbol.iterator])
+}
+
+module.exports = isIterable
--- a/src/core/isIterable.spec.js
+++ b/src/core/isIterable.spec.js
@@ -1,0 +1,44 @@
+const test = require('tape')
+
+const isFunction = require('./isFunction')
+
+const identity = x => x
+
+const isIterable = require('./isIterable')
+
+const testIterable = {
+  from: 1,
+  to: 5,
+
+  [Symbol.iterator]() {
+    this.current = this.from
+    return this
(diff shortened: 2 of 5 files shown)


src/predicates/isIterable.js
```diff
@@ -0,0 +1,5 @@
+/** @license ISC License (c) copyright 2018 original and current authors */
+/** @author Dale Francis (dalefrancis88) */
+
+module.exports =
+  require('../core/isPromise')
```
evilsoft: This is including `isPromise`

src/predicates/isIterable.spec.js
```diff
@@ -0,0 +1,10 @@
+const test = require('tape')
+
+const isFunction = require('../core/isFunction')
+
+const isMonad = require('./isIterable')
```
evilsoft: this should be named `isIterable`

src/core/isIterable.spec.js
```diff
@@ -0,0 +1,44 @@
...
+  t.equal(isIterable(undefined), false, 'returns false for undefined')
+  t.equal(isIterable(null), false, 'returns false for null')
+  t.equal(isIterable(0), false, 'returns false for falsey number')
+  t.equal(isIterable(1), false, 'returns false for truthy number')
+  t.equal(isIterable(false), false, 'returns false for false')
+  t.equal(isIterable(true), false, 'returns false for true')
+  t.equal(isIterable({}), false, 'returns false for an object')
+  t.equal(isIterable(identity), false, 'returns false for function')
+
+  t.equal(isIterable([]), true, 'returns false for an array')
```
evilsoft: text should read 'returns *true* for an array'

src/core/isIterable.spec.js
```diff
@@ -0,0 +1,44 @@
...
+  t.equal(isIterable('string'), true, 'returns true for truthy string')
```
evilsoft: We should also add a test for falsey string, just to be safe.

src/core/isIterable.js
```diff
@@ -0,0 +1,10 @@
+/** @license ISC License (c) copyright 2018 original and current authors */
+/** @author Dale Francis (dalefrancis88) */
+
+const isFunction = require('./isFunction')
+
+function isIterable(iterable) {
+  return iterable !== null && iterable !== undefined && isFunction(iterable[Symbol.iterator])
```
evilsoft: with `NaN` also not being iterable, could we use the `isNil` predicate here?

src/predicates/isIterable.js
dalefrancis88: That's what i get for copying, fail :)

src/predicates/isIterable.spec.js
dalefrancis88: another fail copy :/

src/core/isIterable.spec.js
dalefrancis88: ok i need to stop copying :/

src/core/isIterable.js
dalefrancis88: cool, good idea, will do it

PR #333, opened 2018-10-25
Add converge (Phoenix) to combinators.

amsross:
Add a Phoenix (Starling Prime) combinator.

`phoenix :: (b -> c -> d) -> (a -> b) -> (a -> c) -> a -> d`

re:
* https://hackage.haskell.org/package/data-aviary-0.4.0/docs/Data-Aviary-Birds.html#v:phoenix
* https://github.com/fantasyland/fantasy-birds#phoenix--b---c---d---a---b---a---c---a---d
* https://ramdajs.com/docs/#converge

eg:
```
  const result = converge(concat, reverse, toUpperCase, 'foo') // => 'oofFOO'
```


amsross:
I named this `converge` because that's what it's called in [ramda](https://ramdajs.com/docs/#converge). I'm not married to that name.

evilsoft:
Sweet!
we have this with
`compose(merge(binaryFunc), fanout(f, g))`
And the only reason I avoided `converge` before is because sometimes you wanna keep that `Pair` for further extension and then merge later.

BUT as the Phoenix this makes sense and I ❤️ that it it keeps it binary, if we need a Triple, the same pattern works

```js
const right =
 converge(liftA2(assign), func, another)

const flow =
  converge(liftA2(assign), Maybe.of, right)
```

evilsoft:
`converge` is an excellent name.
Just a couple things on this PR:
* The copy write should go to you, if you want it, otherwise you can default it to me.
* You can add yourself to the README by running: `yarn contributors:add amsross code`, if you want to.
* Need to expose it to the main `crocks` api (check out [`src/combinators/index.js`](https://github.com/evilsoft/crocks/blob/master/src/combinators/index.js) and [`src/index.js`](https://github.com/evilsoft/crocks/blob/master/src/index.js))
* need to add documentation for this on these files:
  * [functions-index](https://github.com/evilsoft/crocks/blob/master/docs/src/pages/docs/functions/index.md)
  * [combinator-functions](https://github.com/evilsoft/crocks/blob/master/docs/src/pages/docs/functions/combinators.md)

Even though none of the other combinators do not have example code, as of late we have been requiring that for all new documented functions we provide a simple code example.

evilsoft:
Also thank you so much for taking the time for this wonderful addition to `crocks`

amsross:
@evilsoft it looks to me like I've squashed everything in https://github.com/evilsoft/crocks/pull/333#issuecomment-433152028. Let me know if [those docs](https://github.com/evilsoft/crocks/blob/f6ac272a50b9102a1486face30251b8314b4cc1e/docs/src/pages/docs/functions/combinators.md#converge) are sufficient (or even too much).

src/combinators/converge.js
```diff
@@ -0,0 +1,17 @@
...
+const isFunction = require('../predicates/isFunction')
+
+// converge (Phoenix or Starling Prime)
+// (b -> c -> d) -> (a -> b) -> (a -> c) -> a -> d
+function converge(f, g, h, x) {
+  if(!isFunction(f) || !isFunction(g) || !isFunction(h)) {
+    throw new TypeError('converge: Functions required for first two arguments')
+  }
+
+  return curry(f)(g(x), h(x))
```
evilsoft: BTW. I ❤️ that you thought to do this.

src/combinators/converge.js
```diff
@@ -0,0 +1,17 @@
+/** @license ISC License (c) copyright 2016 original and current authors */
+/** @author Matt Ross (amsross) */
+
+const curry = require('../helpers/curry')
```
evilsoft: I am sorry that I also missed this in my original review:
both `curry` and `isFunction` should come from `require('../core/curry')` and `require('../core/isFunction')`.
Once we move to a mono-repo, `@crocksjs/core` will have to be installed to keep the other packages light.

Again, I am sorry I missed this.

src/combinators/converge.js
```diff
@@ -0,0 +1,17 @@
+/** @license ISC License (c) copyright 2016 original and current authors */
```
evilsoft: The year here should be `2018`, another thing I failed to find.

src/combinators/converge.js
```diff
@@ -0,0 +1,17 @@
...
+/** @author Matt Ross (amsross) */
+
+const curry = require('../helpers/curry')
+const isFunction = require('../predicates/isFunction')
+
+// converge (Phoenix or Starling Prime)
+// (b -> c -> d) -> (a -> b) -> (a -> c) -> a -> d
+function converge(f, g, h, x) {
+  if(!isFunction(f) || !isFunction(g) || !isFunction(h)) {
+    throw new TypeError('converge: Functions required for first two arguments')
```
evilsoft: Should read "for first *three* arguments"

src/combinators/converge.spec.js
```diff
@@ -0,0 +1,56 @@
...
+test('converge (Big Phi or S\' combinator)', t => {
+  const s = bindFunc(sub)
+  const x = 67
+  const f = x => y => x + y
+  const g = x => x - 1
+  const h = x => x + 1
+
+  t.ok(isFunction(sub), 'is a function')
+
+  t.throws(s(undefined, g, h, x), TypeError, 'throws with first arg undefined')
```
evilsoft: Another thing in the specs, typically we want to test for the specific Error Message that we provide so you could do
```js
const err = /converge: Functions required for first three arguments$/
```

and do a search and replace for `:%s/TypeError/err/g`

src/combinators/converge.spec.js
```diff
@@ -0,0 +1,56 @@
+const test = require('tape')
+const helpers = require('../test/helpers')
+
+const bindFunc = helpers.bindFunc
+const isFunction = require('../predicates/isFunction')
```
evilsoft: this should also come from `core`

amsross:
Got 'em, thanks!

I initially branched from a _very_ outdated `master`, so it didn't make for a perfect template. When I rebased against an updated `master` I didn't go back and match the updated style.

src/combinators/converge.js
amsross: Honestly, I copied the approach directly out of `substitution`.

docs/src/pages/docs/functions/combinators.md
```diff
@@ -48,6 +48,55 @@ This is a very handy dandy function, used a lot. Pass it any value and it will
...
+import { Just } from 'crocks/Maybe'
+import { alt } from 'crocks/pointfree'
+import { converge } from 'crocks/combinators'
+import { prop } from 'crocks/helpers'
+
+const divide = x => y => x / y
+const sum = xs => xs.reduce((m, n) => m + n, 0)
+const length = xs => xs.length
+
+converge(divide, sum, length)([1, 2, 3, 4, 5])
```
evilsoft: it may be beneficial to create an `average` function for the benifiet of the siggy like

```js
// data :: [ Number ]
const data =
  [ 1, 2, 3, 4, 5 ]

// average :: [ Number ] -> Number
const average =
  converge(divide, sum, length)

average(data)
//=> 3
```

amsross: I didn't declare variables/signatures for the objects in the second half of the example because it seemed too verbose. I can do it if that's desired, though.

docs/src/pages/docs/functions/combinators.md
```diff
@@ -48,6 +48,55 @@ This is a very handy dandy function, used a lot. Pass it any value and it will
...
+converge :: (b -> c -> d) -> (a -> b) -> (a -> c) -> a -> d
+```
+
+Provides a means of passing an acculumating function and two branching functions.
+A value can be applied to the resulting function which will then be applied to
+each branching function, the results of which will be applied to the accumulating
+function.
+
+```javascript
+import { Just } from 'crocks/Maybe'
```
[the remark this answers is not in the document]
amsross: I'm in node 100% of the time, so you could've fooled me!

docs/src/pages/docs/functions/combinators.md
```diff
@@ -48,6 +48,71 @@ This is a very handy dandy function, used a lot. Pass it any value and it will
...
+const sum = xs => xs.reduce((m, n) => m + n, 0)
+// length :: [ a ] -> Number
+const length = xs => xs.length
+// average :: [ Number ] -> Number
+const average = converge(divide, sum, length)
+
+average(data)
+//=> 3
+
+// maybeGetDisplay :: Object -> Maybe String
```
evilsoft: `prop` can actually take any `a` because of the `Maybe`, so these should be `a -> Maybe String`

docs/src/pages/docs/functions/combinators.md
```diff
@@ -48,6 +48,71 @@ This is a very handy dandy function, used a lot. Pass it any value and it will
...
+// maybeGetFirst :: Object -> Maybe String
+const maybeGetFirst = prop('first')
+// maybeGetLast :: Object -> Maybe String
+const maybeGetLast = prop('last')
+// maybeConcatStrings :: Maybe String -> Maybe String -> Maybe String
+const maybeConcatStrings = x => y => Just(x => y => x + ' ' + y).ap(x).ap(y).alt(x).alt(y)
+// maybeMakeDisplay :: Object -> Maybe String
+const maybeMakeDisplay = converge(maybeConcatStrings, maybeGetFirst, maybeGetLast)
+// maybeGetName :: Object -> Maybe String
+const maybeGetName = converge(alt, maybeGetDisplay, maybeMakeDisplay)
```
evilsoft: These functions could be made easier on the 👀 with some newlines between them.

docs/src/pages/docs/functions/combinators.md
amsross: Should this maybe be `a -> Maybe b` since the property value might not necessarily be a string?

evilsoft: Good call you are right!!

evilsoft:
Amazing! Thanks for this addition.
Here have a Strong Squirrel for such a Strong PR:
image

docs/src/pages/docs/functions/combinators.md
```diff
@@ -48,6 +48,79 @@ This is a very handy dandy function, used a lot. Pass it any value and it will
...
+maybeGetName({ first: 'J', last: 'S' })
+//=> Just('J S')
+
+maybeGetName({ display: 'Jack Sparrow', first: 'J', last: 'S' })
+//=> Just('Jack Sparrow')
+
+maybeGetName({ first: 'J' })
+//=> Just('J')
+
+maybeGetName({ first: 'S' })
```
amsross: This should read `last: 'S'`. oops 😢

Issue #237, opened 2018-03-01
Update `filter` functions and methods to be `Filterable` instead of `Foldable` for FL


evilsoft:
Documentation for the `filter` pointfree function needs to have its signature updated to reflect `Filterable`. Currently it takes a `Foldable`.
Also, `List` needs to be updated to expose `fantasy-land/filter`

dalefrancis88:
Looking at the PR, this appears to be complete @karthikiyengar @evilsoft is there anything left?

PR #341, opened 2018-10-29
Change Foldable to Filterable on List.filter for FL Compliance

karthikiyengar:
This PR addresses #237 

I'm not sure if we need to change this in the pointfree docs - I see that most functions don't represent a typeclass. Thoughts?


evilsoft:
image

Issue #337, opened 2018-10-27
Move `fanout` to `src/Pair`


evilsoft:
### Description/Location

**location:** 'src/helpers/fanout.js'
**description:**
`fanout` is out of place from our pattern of putting functions into the folders of the type they return/construct outside of normal JS types. As `fanout` NEEDS to have a `Pair` it should live inside of `src/Pair`.

### Task List

* [ ] move `fanout` from `src/helpers/fanout.js` into `src/Pair`
* [ ] move `fanout` from `src/helpers/fanout.spec.js` into `src/Pair`
* [ ] Update references to main api references
* [ ] Update the documentation

For more information, see the `CONTRIBUTING` guide.

evilsoft:
it has not shipped yet. I like to keep them open until released.

rpearce:
Got it! Thanks for clarifying

evilsoft:
Is now in production!
Closing!

Issue #307, opened 2018-08-06
Move `project` pointfree function into `Tuple` folder


evilsoft:
`project` pointfree should live in `Tuple` and also needs to be documented.

* Move `project` to `Tuple` folder
* Make sure it is still accessible from the `crocks` main api.
* Add full documentation to the `Tuple` docs
* Add documentation entry on `Pointfree Functions` page.

karthikiyengar:
@evilsoft - I was checking the existing docs and I do see the documentation for `project`

Additionally, one of the examples for project implies that it also works with a pair:

```
Pair('one', 'two')
  .project(2)
//=> "two"
```

Are we going to implement this? And if so, do we still move `project` around?

evilsoft:
That example is working on a `2-Tuple` named `Pair`.

dalefrancis88:
Closing as this appears to be complete

PR #342, opened 2018-10-29
Move fanout to the Pair folder

karthikiyengar:
This PR addresses #337 

Okay, so I have a couple of questions on how we're going to approach this.

I see that as a part of #307, we're still exposing `project` and `nmap` in `pointfree`. I followed the same line of reasoning and I've exposed this under `helpers`. 

I was thinking about it and I'm not completely sure why we've exposed them in different folders to begin with. Also, considering our monorepo move, I'm assuming both PRs need to be updated to not expose them in these folders?


src/helpers/index.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
   defaultProps: require('./defaultProps'),
   defaultTo: require('./defaultTo'),
   dissoc: require('./dissoc'),
-  fanout: require('./fanout'),
+  fanout: require('../Pair/fanout'),
```
evilsoft: I think you were pointing at an old ref in your comment.
We no longer export cross folder in those index files.
This will need to be removed from this index file and referenced solely in the main crocks index file.

docs/src/pages/docs/functions/helpers.md
```diff
@@ -293,7 +293,7 @@ functions, `dissoc` will remove any `undefined` values from the result.
 
 #### fanout
 
-`crocks/helpers/fanout`
+`crocks/Pair/fanout`
```
evilsoft: This documentation should be moved to the `Pair` docs, and as part of that move will now require an example

evilsoft: That also means that `fanout` will need to be removed from the tags list above, and added to the tags list in `Pair` (check out `branch` for an example.)

karthikiyengar:
Thanks for the review. Addressed!

evilsoft:
LGTM. With this being a breaking change, lets get this in, in the next cycle, keep this next release an additions and docs release for this week.

but here is a nice squirrel for all your efforts!
image

Issue #203, opened 2018-02-12
Add flatting examples to natural transformation examples


evilsoft:
When looking over the docs we already have, noticed that the Natural Transformation Examples do not show flattening. Basically they should provide something like this:
```js
const { Nothing, Just } = require('crocks/Maybe')

const { Left, Right } = require('crocks/Either')

const eitherToMaybe = require('crocks/Maybe/eitherToMaybe')

Just(Right(34))
  .chain(eitherToMaybe)
// Just 34

Just(Left(false))
  .chain(eitherToMaybe)
// Nothing
```

dalefrancis88:
To update the currently live documentation the following need to be done:

[ ] Async
[ ] Either
[ ] Maybe
[ ] Pair
[ ] First
[ ] Last

dalefrancis88:
First and last not valid because they don't have a `chain` method

dalefrancis88:
Going to close this issue off, @evilsoft we can re-open if you feel there is still more to do

PR #343, opened 2018-10-30
Added flattening examples to the Maybe docs

dalefrancis88:
Adding in the examples per #203

### DIFF
--- a/docs/src/pages/docs/crocks/Maybe.md
+++ b/docs/src/pages/docs/crocks/Maybe.md
@@ -1452,6 +1452,10 @@ Just([ 'first', 'second', 'third' ])
 Just(First('first'))
   .chain(firstToMaybe)
 //=> Just "first"
+
+Just(First.empty())
+  .chain(firstToMaybe)
+//=> Nothing
 ```
 
 #### lastToMaybe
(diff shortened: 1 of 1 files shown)


docs/src/pages/docs/crocks/Maybe.md
```diff
@@ -1438,6 +1448,14 @@ Just([])
 Just([ 'first', 'second', 'third' ])
   .chain(firstToMaybe(firstValue))
 //=> Just "first"
+
+Just(First('first'))
+  .chain(firstToMaybe)
+//=> Just "first"
+
+Just(First.empty())
```
HenriqueLimas: Should we also use an example using Last? What this will return ?

docs/src/pages/docs/crocks/Maybe.md
```diff
@@ -1492,6 +1511,14 @@ Just([])
 Just([ 'first', 'second', 'third' ])
   .chain(lastToMaybe(lastValue))
 //=> Just "third"
+
+Just(Last('last'))
+  .chain(lastToMaybe)
+//=> Just "last"
+
+Just(Last.empty())
```
HenriqueLimas: Same here but with First.

docs/src/pages/docs/crocks/Maybe.md
dalefrancis88: Hey @HenriqueLimas this would throw an error `TypeError: firstToMaybe: First or First returning function required`, i'm not sure it makes sense unless you mean to want to show that?

dalefrancis88:
Going to merge this actually and then if we want to add the error example will create a new PR

PR #345, opened 2018-11-02
Added flattening examples to Async

dalefrancis88:
This is for #203, I've deliberately left out examples for `asyncToPromise`. I don't see any value in it or think it would be a common thing. Happy to hear otherwise though

### DIFF
--- a/docs/src/pages/docs/crocks/Async.md
+++ b/docs/src/pages/docs/crocks/Async.md
@@ -1331,7 +1331,7 @@ eitherToAsync :: (a -> Either c b) -> a -> Async c b
 ```
 
 Used to transform a given [`Either`][either] instance to an `Async` instance or
-flatten a `Async` of `Either` into a `Async` when chained, `eitherToAsync` will
+flatten an `Async` of `Either` into an `Async` when chained, `eitherToAsync` will
 turn a [`Right`][right] instance into a [`Resolved`](#resolved) instance
 wrapping the original value contained in the original [`Right`][right]. If a 
 [`Left`][left] is provided, then `eitherToAsync` will return a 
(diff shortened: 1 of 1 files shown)


docs/src/pages/docs/crocks/Async.md
```diff
@@ -1583,6 +1603,16 @@ Resolved([ 'cat', 'bat', 'imp' ])
...
   .fork(log('rej'), log('res'))
 //=> rej: "Nothing Found"
+
+Resolved(Last.empty())
+  .chain(lastToAsync('Left'))
+  .fork(log('rej'), log('res'))
+//=> rej: "Left"
+
+Resolved(Last('too know!'))
+  .chain(lastToAsync('Error'))
```
HenriqueLimas: Can we use `Left` her instead of `Error`, so we maintain the same logic btw the examples and it's easier to compare.

docs/src/pages/docs/crocks/Async.md
```diff
@@ -1492,6 +1502,16 @@ Resolved([ 'cat', 'bat', 'imp' ])
...
   .fork(log('rej'), log('res'))
 //=> rej: "Nothing Found"
+
+Resolved(First.empty())
+  .chain(firstToAsync('Left'))
+  .fork(log('rej'), log('res'))
+//=> rej: "Left"
+
+Resolved(First(42))
+  .chain(firstToAsync('Error'))
```
HenriqueLimas: Can we use Left her instead of Error, so we maintain the same logic btw the examples and it's easier to compare

docs/src/pages/docs/crocks/Async.md
dalefrancis88: Yeah good idea

docs/src/pages/docs/crocks/Async.md
dalefrancis88: Yeah not sure why i changed it lol

dalefrancis88:
This one was pretty straight forward now, I think i'll do the next four in one PR. I am generally a fan of small deliberate PRs but i think this'll be fine

PR #346, opened 2018-11-02
Added flattening examples

dalefrancis88:
This is for adding flattening examples for `Either`, `Pair` for #203

### DIFF
--- a/docs/src/pages/docs/crocks/Either.md
+++ b/docs/src/pages/docs/crocks/Either.md
@@ -13,16 +13,16 @@ Either c a = Left c | Right a
 `Either` is the canonical `Sum` type and is the base of all other `Sum` types
 included in `crocks`, such as [`Maybe`][maybe] and [`Async`][async] to name a
 few. Defined as a tagged union type, it captures the essence of disjunction as
-it provides either a `Left` value or a `Right` value and not both.
+it provides either a [`Left`](#left) value or a [`Right`](#right) value and not both.
 
 Unlike [`Maybe`][maybe], which fixes its "left side", or [`Nothing`][nothing] to
-a `()` (unit), `Either` is a functor in both it's `Left` and `Right` sides. This
+a `()` (unit), `Either` is a functor in both it's [`Left`](#left) and [`Right`](#right) sides. This
 allows for the ability to vary the type on either side and is akin to the
 imperative `if...else` trees that are common in programming.
 
 Like most other types in `crocks`, `Either` has a right bias in regard to
-instance methods like `map`, `ap` and `chain`. This behavior can be used to
-formally capture Error handling as the `Left` value will be maintained
+instance methods like [`map`](#map), [`ap`](#ap) and [`chain`](#chain). This behavior can be used to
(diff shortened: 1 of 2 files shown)


docs/src/pages/docs/crocks/Either.md
```diff
@@ -944,10 +944,10 @@ incValue({ a: 44 })
 Either c a ~> ((c -> d), (a -> b)) -> Either b d
 ```
 
-Used to map the value of an `Either` instance and transform a `Left` into a
-`Right` or a `Right` into a `Left`, `swap` takes two functions as its arguments.
-The first function is used to map and transform a `Left` into a `Right`,
-while the second maps and transforms a `Right` into a `Left`. If no mapping of
+Used to map the value of an `Either` instance and transform a [`Left`](#left) into a
+`Right` or a [`Right`](#right) into a [`Left`](#left), `swap` takes two functions as its arguments.
```
HenriqueLimas: Should we add on this `right` also?

docs/src/pages/docs/crocks/Either.md
```diff
@@ -996,7 +996,7 @@ Either c a ~> ((c -> b), (a -> b)) -> b
 
 Used as a means to map and extract a value from an `Either` based on the
 context, `either` takes two functions as its arguments. The first will map any
```
HenriqueLimas: Should `either` be capitalized?

docs/src/pages/docs/crocks/Either.md
```diff
@@ -1037,10 +1037,10 @@ firstToEither :: c -> First a -> Either c a
 firstToEither :: c -> (a -> First b) -> a -> Either c a
 ```
 
-Used to transform a given [`First`][first] instance to an `Either`
-instance, `firstToEither` will turn a non-empty instance into
-a [`Right`](#right) wrapping the original value contained within
-the [`First`][first].
+Used to transform a given [`First`][first] instance to an `Either` instance or
```
HenriqueLimas: Should we be
Have a `#` before `first`?

docs/src/pages/docs/crocks/Either.md
```diff
@@ -1113,10 +1121,10 @@ lastToEither :: c -> Last a -> Either c a
 lastToEither :: c -> (a -> Last b) -> a -> Either c a
 ```
 
-Used to transform a given [`Last`][last] instance to
-an `Either`, `lastToEither` will turn a non-empty [`Last`][last] instance into
-a [`Right`](#right) instance wrapping the original value contained in the
-original non-empty.
+Used to transform a given [`Last`][last] instance to an `Either` or flatten an
```
HenriqueLimas: Here too for `last`, maybe it's missing `#`

docs/src/pages/docs/crocks/Either.md
dalefrancis88: The first `Either` is capatilized because it's the type and the second `either` isn't because it's the function

evilsoft:
image

PR #348, opened 2018-11-06
Simplify `isNil` predicate for better backward compatability.

evilsoft:
## Well this is just silly
image

Even though we have to shut off eslint here, I much prefer this implementation, saves a compare for the `null` and `undefined` and works better with legacy browsers with the `isNaN` detector.

Thoughts?


HenriqueLimas:
I am wondering if we should consider `NaN` as a nil value. It seems that [lodash](https://lodash.com/docs/4.17.11#isNil) and [ramda](https://ramdajs.com/docs/#isNil) do not consider it as a nil value.

evilsoft:
@HenriqueLimas Yeah I get that. My reasoning is as follows as to why it will stay in `isNil`

 `NaN` is a value of `Number` as evidenced by the fact of `typeof NaN === 'number'`, But it is a distinguished element not only in `Number` but in all of `Javascript. It is special in that it is the ONLY value in which `x => x ===  x` will return `false`. This makes it a singleton in the language of JS. All singletons are isomorphic to every other singleton. So because the other (2) singletons are `undefined` and `null`, `NaN` is isomorphic to them, and as such it is just a matter of labeling at this point and all of them can be labeled as `Nil`.

That is the argument  that is on the table that should be disputed and not an Appeal to Authority. Do you have a counter to the above argument?

evilsoft:
image

PR #349, opened 2018-11-08
Added support for iterables to `head`

dalefrancis88:


src/pointfree/head.spec.js
```diff
@@ -72,3 +72,19 @@ test('head pointfree List', t => {
...
+  const one = testIterable(1, 1, 1)
+  const two = testIterable(1, 2, 1)
+
+  t.equals(head(empty).type(), M.type(), 'returns a Maybe on empty')
+  t.equals(head(one).type(), M.type(), 'returns a Maybe on one element array')
+  t.equals(head(two).type(), M.type(), 'returns a Maybe on two element array')
+
+  t.equals(head(empty).option('Nothing'), 'Nothing', 'returns a Nothing on empty')
+  t.equals(head(one).option('Nothing'), 1, 'returns `1` on single element iterable')
+  t.equals(head(two).option('Nothing'), 1, 'returns `1` on dual element iterable')
```
HenriqueLimas: IMHO opinion it should be better to test a different value than 1 for the iterator with two elementas. The same idea we test on array and string.

dalefrancis88:
@HenriqueLimas can you also check out https://github.com/evilsoft/crocks/issues/29 this is the issue this originally came from. I am trying to think what I could do for the other functions

src/pointfree/head.spec.js
dalefrancis88: updated

HenriqueLimas:
@dalefrancis88 I found it a great way to do without changing type. Do you have concerns about this solution in the other functions?

dalefrancis88:
Yeah i this is fine but for something like `map` it'll either be mutating the object or changing returning an array right?

PR #352, opened 2018-11-28
Added a test for cloning arrays

dalefrancis88:
@evilsoft I think @HenriqueLimas may have been correct in adding in support directly for Arrays. It appears that they don't clone well. It all looked fined until I went to use it in head and it didn't work. It appears that the context for the iterator is lost and so when calling `next` you just get `{ value: undefined, done: true }`. We can get around this by putting the original code back in for `array` and `string` but i thought it best to re-open this discussion. 

You can remove the array code in cloneIterable to see the failure

### DIFF
--- a/src/core/cloneIterable.js
+++ b/src/core/cloneIterable.js
@@ -1,7 +1,13 @@
 /** @license ISC License (c) copyright 2018 original and current authors */
 /** @author Henrique Limas (HenriqueLimas) */
 
+const isArray = require('../core/isArray')
+
 function cloneIterable(source) {
+  if(isArray(source)) {
+    return source.slice(0)
+  }
+
   let copy = Object.create(Object.getPrototypeOf(source))
   Object.assign(copy, source)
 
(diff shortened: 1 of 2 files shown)


evilsoft:
We should stick with using Array directly even when in the iterable functions.
So for things like `take` on an Array that would do an `m.slice(0, num - 1)` when it is an Array. 
What are some of the cases that we would want to clone an Array?

dalefrancis88:
I'm not sure we ever will want to clone it. This came more from the side of treating all iterables as the same so that way we didn't have to have code that checked for string or array then checked iterable but after playing with it a bit and doing research i'm quite happy to leave as is and do the check for array/string then iterable in each part. Although it might be worthwhile creating a branching helper for that given we'll be doing it in a few spots

HenriqueLimas:
Agree, I think we should separately check it. Probably it will be repetitive, but at least we would have a consistent way of deal with that. Sometimes we need a copy like `map`, but others we do not.

PR #353, opened 2018-11-28
Update for head that now uses the new cloneIterable function

dalefrancis88:
This may change depending on the outcome of https://github.com/evilsoft/crocks/pull/352 however this works with the current version now


dalefrancis88:
And will also work with the new version

dalefrancis88:
@evilsoft any thoughts on this?

dalefrancis88:
I'm not sure i get what you mean?

HenriqueLimas:
Ouch, sorry. So I mean that cloneIterable, will clone the reference of the iterator but not the reference of the elements inside it. So the first element of the cloned iterable is the exact one of the original iterable that is why we shouldn't need to clone the iterable and than return the first element since it will be the same as returning the first element of the original iterator. Hope I was able to explain myself

PR #356, opened 2018-12-06
Moved List outside of core. Fixes #336

### LINKED ISSUE #336: Remove `List` from `src/core`
## Move `List` out of `src/core`

**location:** 'src/core/List.js'
**description:**
There is the only reason `List` is in `core` is for some specs and use in `src/Pair/toPairs`. It actually makes more sense to have `toPairs` return an Array of Pairs. With that done

### Task List

* [ ] Alter `toPairs` to return `Array` instead `List`
* [ ] Remove/Replace specs using `List` inside of `src/Async/Async.spec.js`
* [ ] Remove/Replace specs using `List` inside of `src/Maybe/find.spec.js`
* [ ] Move implementation from `src/core/List.js` and `src/core/List.spec.js` to `src/List/index.js` and `src/List/List.spec.js`
* [ ] Update the documentation

aecepoglu:
I couldn't have git realize the file move (because I was overwriting the older file within the same commit)


src/Async/Async.spec.js
```diff
@@ -4,7 +4,7 @@ const helpers = require('../test/helpers')
 
 const bindFunc = helpers.bindFunc
 
-const List = require('../core/List')
+const List = require('../List')
```
evilsoft: So this is interesting, cannot believe I missed this in the original definition. Sorry to add more work to this.

There are a couple things here that will need to be changed with the `Async` Type.
First of all in `Async`, for `Async.all` we check for a foldable type, and we also need to check with `isTraversable`. Which means that `isTraversable` will then also need to be moved core folder so it can be referenced in `Async`
Then on top of that, `sequence` is not defined by the fantasy-land spec, so we would need to change the `all` implementation to be: `asyncs.traverse(Async.of, x => x)`, but `array` can stay as is.

Then after all of that, the spec for `Async` for `all` working on Lists, should work with any Traversable containing Async. then the [`MockCrock`](https://github.com/evilsoft/crocks/blob/master/src/test/MockCrock.js) can be in place of `List` for the `all` spec.

So basically the following needs to happen:
* Move functionality for `isTraversable` into core.
* Update `isTraversable` in predicates to use core version.
* In `Async.all` implementation, add a check for `isTraversable` before traversing.
* update error to mention `Traversable` as well as foldable.
* change from sequence to traverse for the non-array case.
* update the `VERSION` of `Async` as this is an api change.
* update all spec in Async to use the `MockCrock`

So the implementation of Async.all should look like:
```js
const allErr = 'Async.all: Foldable Traversable structure of Asyncs required'
function all(asyncs) {
  if(!(isFoldable(asyncs) && allAsyncs(asyncs))) {
    throw new TypeError(allErr)
  }

  if(isArray(asyncs)) {
    return array.sequence(Async.of, asyncs)
  }

  if(isTraverable(asyncs)) {
    return asyncs.traverse(Async.of, x => x)
  }

  throw new TypeError(allErr)
}
```

package-lock.json
```diff
@@ -2712,12 +2712,14 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true
+          "dev": true,
```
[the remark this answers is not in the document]
aecepoglu: true, I didn't realize I committed this.

src/Async/Async.spec.js
aecepoglu: Sounds doable.

docs/src/pages/docs/crocks/Pair.md
```diff
@@ -887,14 +887,12 @@ const record = {
 const joinField = (key, value) =>
   `${key}:${value}`
 
-// joinRecord :: List String -> String
-const joinRecord = list =>
-  list.toArray()
-    .join('|')
+// joinRecord :: String -> [String] -> String
+const join = separator => array => array.join(separator)
```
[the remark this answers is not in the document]
aecepoglu: Ah, `joinWith` is my preferred way to name a method like this but I assumed you would choose differently :)

src/Async/Async.spec.js
aecepoglu: Still sounds doable but I don't believe I have time in the short term future to do this. Any takers?

dalefrancis88:
I'll have a look at this and see if I can finish the last few review comments

PR #359, opened 2018-12-08
Adding examples for the `And` function

dalefrancis88:
Added examples for And. LMK if this works


dalefrancis88:
Hahaha for us too, Aussies only have to be 18 but i figured can't miss if you do 21 😄

Issue #355, opened 2018-12-05
Test for monoid empty values in `isEmpty`


jasminabasurita:
Currently passing an empty from any monoid to `isEmpty` returns false. I think it would make sense if `isEmpty` included tests for empty monoids 

###### Example:
```js
//current
isEmpty(Min.empty()) // => false
//desired
isEmpty(Min.empty()) // => true
```

dalefrancis88:
I really like this idea. How were thinking it'd best be implemented with being able to detect that it is an `empty` of a particular monoid?

karthikiyengar:
This would be great! Would we also be including zero (https://github.com/fantasyland/fantasy-land#plus) in this check? This would mean that `isEmpty(Maybe.zero())` would be true.

dalefrancis88:
I am not sure we should consider `zero` an `empty` value if we start to do that, we're a step away from considering `Nothing` an `empty` value. I think `zero` itself holds an intrinsic value above that of something that is `empty`

jasminabasurita:
> I really like this idea. How were thinking it'd best be implemented with being able to detect that it is an `empty` of a particular monoid?

I was thinking that perhaps the different monoids could include an `isEmpty` function that when invoked would return the expected boolean. Then all we'd need to do within the pointfree `isEmpty` is:
1. check that the function exists and is a function
2. invoke it and return the value

dalefrancis88:
so all monoids values include `empty` as a function, so a `Min(1)` contains a method `empty` and so does a `Min.empty()` do you think there'd be a way to use that to only need to update the `isEmpty` implementation?

karthikiyengar:
What do you think about `isMonoid(x) && equals(x.empty(), x)?`

dalefrancis88:
@jasminabasurita This is now merged into master and will be available in the next release :)

PR #364, opened 2018-12-19
Adding in check for empty monoid

dalefrancis88:
This is a change to allow checking for `empty` monoids as requested by @jasminabasurita on https://github.com/evilsoft/crocks/issues/355


src/core/isEmpty.spec.js
```diff
@@ -23,6 +28,7 @@ test('isEmpty predicate function', t => {
   t.equal(isEmpty(''), true, 'returns true with empty string')
   t.equal(isEmpty('string'), false, 'returns false with non-empty string')
 
+  t.equal(isEmpty(Fake.empty()), true, 'returns true with empty monoid')
```
karthikiyengar: Might want to add `returns false for non-empty monoids` for consistency. I think that's what the CI is crying about as well?

src/core/isEmpty.js
```diff
@@ -1,8 +1,14 @@
 /** @license ISC License (c) copyright 2016 original and current authors */
 /** @author Ian Hofmann-Hicks (evil) */
 const isObject = require('./isObject')
+const isMonoid = require('./isMonoid')
+const equals = require('./equals')
 
 function isEmpty(x) {
+  if(isMonoid(x)) {
+    return equals(x, x.empty())
```
evilsoft: to make this compatible with FL as well as crocks types, we should grab the empty off of the constructor doing `x.constructor.empty()`.

src/test/helpers.js
```diff
@@ -27,7 +27,7 @@ function makeFake(algs, useFl) {
 
   const inst = xs.reduce((o, alg) => {
     const fn = useFl ? fl[alg] : alg
-    if(!isRepFunc(alg) && fn) {
+    if((!isRepFunc(alg) || alg === 'empty') && fn) {
```
evilsoft: Why is this needed?

src/core/isEmpty.js
```diff
@@ -1,8 +1,14 @@
 /** @license ISC License (c) copyright 2016 original and current authors */
 /** @author Ian Hofmann-Hicks (evil) */
 const isObject = require('./isObject')
+const isMonoid = require('./isMonoid')
+const equals = require('./equals')
 
 function isEmpty(x) {
+  if(isMonoid(x)) {
+    return equals(x, x.constructor.empty())
```
evilsoft: Sorry, I totally forgot about fantasy land name prefixing,
One last change and then I think this will be GTG.

So this should be something like:
```js
const fl = require('../flNames')

return equals(x, (x.constructor[fl.empty] || x.empty)())
```

evilsoft: Actually to account for pre 1.x fl it should probably be:
```js
return equals(x, (x.constructor[fl.empty] || x.constructor.empty || x.empty)()
```

src/test/helpers.js
dalefrancis88: This was needed because of the way that i was calling empty off of the instance. This code makes sure that the is an empty func on the instance and the constructor, however now that we've changed it this is very likely not needed. I just didn't think to remove it

dalefrancis88:
Hey team, didn't realise there was an issue with the code coverage. Should be good to go now @evilsoft @HenriqueLimas

PR #365, opened 2018-12-20
Result Documentation

dalefrancis88:
This is to add the documentation for `Result` per https://github.com/evilsoft/crocks/issues/42

**Not ready for review, but happy for input**


dalefrancis88:
Hey @evilsoft or @HenriqueLimas I'd love an initial set of 👀 on this. I think everything is ready for review except the `tryCatch` example which i'll get onto tomorrow, assuming I don't REALLY enjoy new years 😄

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1422 @@
...
+//=> Err "down is not a valid number"
+```
+
+#### alt
+
+```haskell
+Result e a ~> Result e a -> Result e a
+```
+
+Providing a means for a fallback or alternative value, `alt` combines two
```
evilsoft: One thing that is missed with `alt` is that fact that `Err` of `Semigroup` will accumulate like `ap` does.
Should be mentioned in the text and shown in the examples.

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1422 @@
...
+functions: ["trycatch", "eithertoresult", "firsttoresult", "lasttoresult", "maybetoresult"]
+weight: 130
+---
+
+```haskell
+Result e a = Err e | Ok a
+```
+
+Result is a Sum Type a step above [`Maybe`][maybe]. With a [`Maybe`][maybe] the
+left side represents the lack of a valu, where as with a `Result` the left
```
evilsoft: "valu" -> "value"

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1422 @@
...
+contains
+the error information from an operation. `Result` is well suited for capturing
+disjunction when the cause of the "error" case needs to be communicated. For
+example, when executing a function and you exception is important or useful.
+
+A `Result` represents disjunction by using two constructors, [`Err`](#err) and
+[`Ok`](#ok). An [`Ok`](#ok) instance represents the positive result while
+[`Err`](#err) is considered the negative. With the exception of
+[`coalesce`](#coalesce), [`swap`](#swap) and [`bimap`](#bimap), all `Result`
```
evilsoft: One thing that needs to be kept in mind with the limiting to 80 chars is that lines cannot end or begin with a code tick. The space between words cannot be sussed out. So in these descriptions the starts and ends be code ticked text.

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1422 @@
...
+// gte :: Number -> Number -> Boolean
+const gte =
+  y => x => x >= y
+
+// lte :: Number -> Number -> Boolean
+const lte =
+  y => x => x <= y
+
+// between :: (Number, Number) -> Boolean
+const between = (x, y) => and(gte(x), lte(y))
```
evilsoft: wanna avoid using single line functions.  These should include the arguments on the first line and the body on the next line indented to keep is :corn:sistant with the rest of the functions. So this would be something like:
```js
const between = (x, y) =>
  and(gte(x), lte(y))
```

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1422 @@
...
+
+// ensure :: (a -> bool) -> (a -> Result a a)
+const ensure = pred => ifElse(pred, Ok, Err)
```
evilsoft: Same as above

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1422 @@
...
+
+// inRange :: Number -> Result
+const inRange = ensure(between(10, 15))
```
evilsoft: same as above

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1422 @@
...
+inRange(12)
+//=> Ok 12
+
+inRange(25)
+//=> Err 25
+
+inRange('Test')
+//=> Err "Test"
+
+const prop =
```
evilsoft: The following example would be better for showing `maybeToResult` and may be a better suited as that functions example.

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1422 @@
...
+layout: "guide"
+functions: ["trycatch", "eithertoresult", "firsttoresult", "lasttoresult", "maybetoresult"]
+weight: 130
+---
+
+```haskell
+Result e a = Err e | Ok a
+```
+
+Result is a Sum Type a step above [`Maybe`][maybe]. With a [`Maybe`][maybe] the
```
evilsoft: It should really be compared with `Either` instead of `Maybe`. With the difference being that you can accumulate `Err`s with `ap` and `alt` when using `Err`s of like `Semigroup`s. Should be mentioned in the body and replace the last code example with a simple example showing that accumulation with either `alt` or `ap`. Does not need to be real world, can save the real world for the method examples for each.

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1422 @@
...
+
+Ok('some string')
+//=> Ok "some string"
+
+Ok(null)
+//=> Ok null
+
+Ok(undefined)
+//=> Ok undefined
+//=> Ok undefined
```
evilsoft: Extra output here.

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1422 @@
...
+## Constructor Methods
+
+#### Err
+
+```haskell
+Result.Err :: e -> Result e a
+```
+
+Used to construct an [`Err`](#err) instance that represents the "false" or
+"Negative" portion of a disjunction. When an instance is an [`Err`](#err), most
```
evilsoft: I think the use of "Negative" my be confusing. What do you think about sticking with just "false"

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1422 @@
...
+`Result` returning methods will just return a new [`Err`](#err) instance with
+the same containing value.
+
+The power of the [`Err`](#err) as opposed to a [`Nothing`] is that it can hold
```
evilsoft: `Result` is better compared to `Either` instead of `Maybe`. May want to reword this to include the bits about accumulation of `Err` of `Semigroup`s for `alt` and `ap`. Basically, if the reader does not need this accumulation, they should stick with `Either`

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1422 @@
...
+import chain from 'crocks/pointfree/chain'
+import bimap from 'crocks/pointfree/bimap'
+import isNumber from 'crocks/predicates/isNumber'
+import ifElse from 'crocks/logic/ifElse'
+import compose from 'crocks/core/compose'
+
+const { Ok, Err } = Result
+
+// ensure :: (a -> bool) -> (a -> Result a a)
+const ensure = pred => ifElse(pred, Ok, Err)
```
evilsoft: the signature need to be fixed up and the function body moved down a line:
```js
// ensure :: (a -> Boolean) -> a -> Result a a
const ensure = pred =>
  ifElse(pred, Ok, Err)
```

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1422 @@
...
+
+// buildError :: String -> String
+const buildError = x => `${x} is not a valid number`
```
evilsoft: While it is not too bad, we probably wanna shy away from saying it is okay to put values in error messages. Imagine if we were checking a password or pulling in critical or private data, this would allow for private data to trickle into logs that may or may not be scrubbed or secured. May wanna change what this error reports. Also should be in two line format:

```js
// buildError :: () -> String
const buildError = () =>
  'number not provided'
```

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1422 @@
...
+import bimap from 'crocks/pointfree/bimap'
+import identity from 'crocks/combinators/identity'
+
+const { Ok, Err } = Result
+
+// ensure :: (a -> bool) -> (a -> Result a a)
+const ensure = pred => ifElse(pred, Ok, Err)
+
+// buildError :: String -> String
+const buildError = x => `${x} is not a valid string`
```
evilsoft: Same comment as `buildError` above.

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1422 @@
...
+
+// protectedAdd10 :: a -> Result String Number
```
evilsoft: wrong signature for this function.

docs/src/pages/docs/crocks/Result.md
dalefrancis88: I thought it did but when i tested it that didn't seem to happen so i was confused. More than likely i made a mistake so will go back and try again

dalefrancis88:
@evilsoft Thanks for looking at this. I'll get the review items fixed up and the `tryCatch` examples in :)

docs/src/pages/docs/crocks/Result.md
dalefrancis88: I see where i made the mix-up, i knew there was this behaviour but i thought it was on `concat`. Will update `alt` and add the examples

docs/src/pages/docs/crocks/Result.md
dalefrancis88: So what you're saying is if you have something like "[`Ok`](#ok)" then it has to have a word in front of it if it's the beginning of a line?

docs/src/pages/docs/crocks/Result.md
dalefrancis88: As in don't re-create the `prop` function, use the real one and then use this for `maybeToResult`?

docs/src/pages/docs/crocks/Result.md
dalefrancis88: I'm not fussed either way, i thought i'd float the idea. My main thought was that it could be a little more comfortable for the non-fp people and was trying to differentiate a bit from `Either`
But i think consistency is key, lets stick with `false`

docs/src/pages/docs/crocks/Result.md
dalefrancis88: it's this signature because the resulting func is curried right?

docs/src/pages/docs/crocks/Result.md
dalefrancis88: Yeah i get where you're coming from here for sure. This was more to show the difference from a `Nothing` in that it held a value. Would it be better to just change the example to something different that shows that? or are we happy to show an example without the value?

docs/src/pages/docs/crocks/Result.md
evilsoft: but the parentheses are not needed and just obfuscates the siggy

docs/src/pages/docs/crocks/Result.md
evilsoft: I think it should be fine without the value.

docs/src/pages/docs/crocks/Result.md
evilsoft: Sorry the name.

docs/src/pages/docs/crocks/Result.md
evilsoft: The difference between Result and Either is the accumulation of `Err` on `ap` and `alt`. Other than that it could still be used for exclusive disjunction just like `Either`

docs/src/pages/docs/crocks/Result.md
dalefrancis88: That was meant to be more of a question, as in, is the difference between your sig and min because it's curried? Was more trying to understand it so i don't make the mistake in the future

docs/src/pages/docs/crocks/Result.md
evilsoft: yeppers or a word after it if it is at the end.

docs/src/pages/docs/crocks/Result.md
evilsoft: HAHA. sorry. your signature is correct at well. They both say the same thing. The only difference is the lack of parentheses.

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1566 @@
...
+saviour in this situation is `tryCatch`. You can wrap this funcition and it
+will always return a `Result`
+
+It is important to note that, as of now, `tryCatch` does not currently support
+manually or externally curried functions. When using it with a function that
+has been `curry`'d by `crocks` all arguments must be passed in one go. It will
+not support partial application. `tryCatch` applies all arguments to the top-level
+function, which works well for `curry`, and the protection only applies to the
+first activation in the run.
+
```
dalefrancis88: For this third paragraph i'm not sure if it's to wordy, let me know thoughts

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1566 @@
...
+be returned that takes a given value and returns a `Result`.
+
+```javascript
+import Result from 'crocks/Result'
+
+import assign from 'crocks/helpers/assign'
+import composeK from 'crocks/helpers/composeK'
+import composeB from 'crocks/combinators/composeB'
+import Either from 'crocks/Either'
+import eitherToResult from 'crocks/Either/eitherToResult'
```
evilsoft: should be:
```js
import eitherToResult from 'crocks/Result/eitherToResult'
```

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1566 @@
...
+import fanout from 'crocks/Pair/fanout'
+import isNumber from 'crocks/predicates/isNumber'
+import liftA2 from 'crocks/helpers/liftA2'
+import map from 'crocks/pointfree/map'
+import maybeToEither from 'crocks/Result/maybeToEither'
```
evilsoft: Should be:
```js
import maybeToEither from 'crocks/Either/maybeToEither'
```

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1566 @@
...
+import merge from 'crocks/Pair/merge'
```
evilsoft: Should be:
```js
import merge from 'crocks/pointfree/merge'
```

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1566 @@
...
+
+// incThem :: a -> Result [ String ] Object
+const incThem = composeB(
+  merge(liftA2(assign)),
+  fanout(incResult('b'), incResult('a'))
+)
+
+Result.of({})
+  .chain(eitherToResult(incThem))
+//=> Err [ "b is not valid", "a is not valid" ]
```
evilsoft: This should be:
```js
//=> Err [ "b is not valid" ]
```

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1566 @@
...
+firstToResult :: e -> (a -> First b) -> a -> Result e b
+```
+
+Used to transform a given [`First`][first] instance to a `Result`
+instance or flatten a `Result` of [`First`][first] into a `Result` when chained, 
+`firstToMaybe` will turn a non-empty instance into an [`Ok`](#ok) wrapping
+the original value contained within the [`First`][first]. All empty instances
+will map to an [`Err`](#err) with the given value.
+
+Like all `crocks` transformation functions, `firstToMaybe` has two possible
```
evilsoft: `firstToMaybe` -> `firstToResult`

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1566 @@
...
+const { empty } = First
+
+// Person :: { name: String, age: Number }
+
+// createPerson :: (String, Number) -> Person
+const createPerson = (name, age) => ({
+  name, age
+})
+
+// liftName :: Person -> First (Maybe String)
```
evilsoft: As all `First` instances wrap a `Maybe` it is not needed in the siggy. this can just be
```js
// liftName :: Person -> First String
```

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1566 @@
...
+const liftName = composeB(
+  First,
+  prop('name')
+)
+
+// mergeFirstName :: [ Person ] -> First (Maybe String)
```
evilsoft: Can just be:
```js
// mergeFirstName :: [ Person ] -> First String
```

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1566 @@
...
+const isEvenOrOdd = n =>
+  n % 2 === 0 ? 'Even' : 'Odd'
+
+// getParity :: Number -> Object
+const getParity = composeB(
+  objOf('parity'),
+  isEvenOrOdd
+)
+
+// getInfo :: a -> Object
```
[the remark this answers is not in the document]
dalefrancis88: Yeah i like it, when it comes to siggy's i'm still not sure so i tend to do my best and then learn from you each time. Hopefully you're finding that there is less and less review comments needed :)

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1566 @@
...
+
+Err('Error 1.')
+  .alt(Err('Error 2.'))
+//=> Err "Error 1.Error 2."
+
+find(gte(41), [ 17, 25, 38, 42 ])
+//=> Ok 42
+
+find(gte(11), [ 1, 2, 3, 4 ])
+//=> Err "Not found"
```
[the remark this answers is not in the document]
dalefrancis88: I apologize for this one, big mistake and shouldn't have been made

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1566 @@
...
+  ensure(isNumber)
+)
+
+// hasError :: Boolean -> Object -> Object
+const hasError =
+  setProp('hasError')
+
+// ResultObject :: { result: Number, hasError: Boolean, error: String }
+
+// buildResult :: (String, Boolean) -> a -> ResultObject
```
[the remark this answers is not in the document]
dalefrancis88: Sorry @evilsoft  this is probably bad naming, especially given that this is docs for `Result`. This is referring to the type alias above. I'll try to think of a better name

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1566 @@
...
+const buildResult = (key, isError) =>
+  composeB(hasError(isError), objOf(key))
+
+// finalize :: Result e a -> Result ResultObject
```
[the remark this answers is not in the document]
dalefrancis88: same as above

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1566 @@
...
+
+// gte :: Number -> Number -> Boolean
+const lte = y => x =>
+  x <= y
+
+// ensure :: (a -> Boolean) -> a -> Result a
+const ensure = pred =>
+  ifElse(pred, Ok, Err)
+
+// tallyOf :: a -> [ a ]
```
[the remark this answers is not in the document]
dalefrancis88: I stole this from the `Either` docs so i'll fix it there too

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1566 @@
...
+updateSmall()
+  .chain(updateSmall)
+  .runWith(3)
+//=> Pair( Err 12, 12 )
+
+updateSmall()
+  .chain(updateSmall)
+  .chain(updateSmall)
+  .runWith(3)
+//=> Pair( Err 12, 12 )
```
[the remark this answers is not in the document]
dalefrancis88: I think you mean the one above this, but either way, fixed :)

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1566 @@
...
+// ensure :: (a -> Boolean) -> a -> Result a
+const ensure = pred =>
+  ifElse(pred, Ok, buildError)
+
+// fromNumber :: a -> Result String Number
+const fromNumber =
+  ensure(isNumber)
+
+// prop :: (String | Number) -> Object -> Result String a
+const prop = name => compose(
```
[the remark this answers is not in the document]
dalefrancis88: I'm always one for simplifying things, i like it

docs/src/pages/docs/crocks/Result.md
dalefrancis88: Good to know

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1557 @@
...
+
+// safeInc :: a -> Maybe Number
+const safeInc =
+  safeLift(isNumber, x => x + 1)
+
+// incProp :: String -> a -> Maybe Number
+const incProp = key =>
+  composeK(safeInc, prop(key))
+
+// incResult :: String -> a -> Result [ String ] Object
```
evilsoft: I believe this returns an `Either`.

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1557 @@
...
+const incResult = key => maybeToEither(
+  [ `${key} is not valid` ],
+  composeB(
+    map(objOf(key)),
+    incProp(key)
+  )
+)
+
+// incThem :: a -> Result [ String ] Object
+const incThem = composeB(
```
evilsoft: I believe this returns an `Either`.

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1557 @@
...
+const mergeFirstName = composeB(
+  firstToResult('(No name found)'),
+  mapReduce(liftName, flip(concat), empty())
+)
+
+mergeFirstName([
+  createPerson('John', 30),
+  createPerson('Jon', 33)
+])
+//=> Ok "Jon"
```
evilsoft: This resolves to:
```js
//=> Ok "John"
```

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1557 @@
...
+
+// Outcome :: { result: Number, hasError: Boolean, error: String }
+
+// buildResult :: (String, Boolean) -> a -> Outcome
+const buildResult = (key, isError) =>
+  composeB(hasError(isError), objOf(key))
+
+// finalize :: Result e a -> Result Outcome
+const finalize = bimap(
+  buildResult('error', true),
```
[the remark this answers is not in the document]
dalefrancis88: I like it :)

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1557 @@
...
+Result e a ~> ((e -> b), (a -> b)) -> b
+```
+
+Used to provide a means to map a given `Result` instance folding it out of its
+container. `either` expects two functions as its arguments. The first is a
+function that will be used to map an [`Err`](#err). While the second
+will map the value wrapped in a given [`Ok`](#ok) and return the result of that
+mapping.
+
+By using composing `either` you can create functions that us the power of
```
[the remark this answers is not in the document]
dalefrancis88: yeah it doesn't quite work does it

docs/src/pages/docs/crocks/Result.md
```diff
@@ -0,0 +1,1557 @@
...
+// ensure :: (a -> Boolean) -> a -> Result String a
+const ensure = pred =>
+  ifElse(pred, Ok, buildError)
+
+// fromNumber :: a -> Result String Number
+const fromNumber =
+  ensure(isNumber)
+
+// swapWithDefault :: Result e a -> Result String Number
+const swapWithDefault =
```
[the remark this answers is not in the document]
dalefrancis88: I've written a different version, hopefully it's a bit better, lmk

evilsoft:
Think this will do it. Amazing work on this @dalefrancis88!
Thank you so much for your time and dedication getting this through.

image

dalefrancis88:
No problem, thanks for your limitless patience :)

PR #367, opened 2018-12-28
Update setPath to differentiate Array/Object by Integer/String

evilsoft:
## For :corn:sistancy
image

This PR is a follow up to [this older PR](https://github.com/evilsoft/crocks/pull/331) and updates `setPath` to only allow Positive `Integer`s for `Array`s and `String`s for `Object`s. Going forward `setPath` will throw if the key type does not align with the type being referenced.

Also updated `setProp`to also only work with Positive `Integer`s, otherwise a negative integer property is added to the array.

Also cleaned up some specs.


dalefrancis88:
Not sure, i thought we did but i found it in my results work, i don't remember where i started the results md from but it's likely in there, looking now it doesn't seem to be anywhere but there, i've likely started this before the deprecation was merged :D

evilsoft:
image

PR #370, opened 2019-01-17
More changes to setPath and unsetPath

evilsoft:
image
As part of this [this issue](https://github.com/evilsoft/crocks/issues/329), there are some opportunities to clean some bits up around the Path functions. This PR implements those changes by doing the following:
* Add `unset` function to `core/array`
* Add `unset function to `core/object`
* Remove a couple needless checks on keys in `setPath`
* Report key based errors in `setPath` outside of next being an Array or Object.
* update `unsetPath` to use new `unset` functions in `array` and `object`
* update `unsetPath` to report errors more like `setPath`

### DIFF
--- a/src/helpers/setPath.js
+++ b/src/helpers/setPath.js
@@ -13,12 +13,13 @@ const object = require('../core/object')
 const isValid = x =>
   isObject(x) || isArray(x)
 
+const pathErr =
+  'setPath: Non-empty Array of non-empty Strings and/or Positive Integers required for first argument'
+
 // setPath :: [ String | Integer ] -> a -> (Object | Array) -> (Object | Array)
 function setPath(path, val, obj) {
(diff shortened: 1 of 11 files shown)


src/helpers/setPath.js
```diff
@@ -30,6 +31,10 @@ function setPath(path, val, obj) {
   const key = path[0]
   let newVal = val
 
+  if(!(isString(key) && !isEmpty(key) || isInteger(key) && key >= 0)) {
```
dalefrancis88: This appears in a few spots, do we want to put it somewhere common or we don't care?

evilsoft: We will more than likely abstract this when we move `prop` -> `getProp` and `propPath` -> `getPath`.

evilsoft:
image

PR #386, opened 2019-03-16
Add `unsetProp` and deprecate `dissoc` in favor of the new function

evilsoft:
## Get rid of the crusties
image

This PR addresses part of [this issue](https://github.com/evilsoft/crocks/issues/329) and adds the `unsetProp` function while also depreciating `dissoc`. The behavior for argument handling matches `unsetPath` in that:
* Only allows non-empty `String`s and positive `Integer`s for referencing keys/indices on a given `Object`/`Array`
* Will echo the provided target when it is not an `Object` or `Array`

### DIFF
--- a/src/helpers/dissoc.spec.js
+++ b/src/helpers/dissoc.spec.js
@@ -2,41 +2,91 @@ const test = require('tape')
 const helpers = require('../test/helpers')
 
 const bindFunc = helpers.bindFunc
-const unit = require('../core/_unit')
+
+const isFunction = require('../core/isFunction')
+const unit =  Function.prototype
 
 const dissoc = require('./dissoc')
 
-test('dissoc', t => {
+const err =
+  /dissoc: Non-empty String required or Positive Integer required for first argument/
+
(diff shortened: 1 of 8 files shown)


src/helpers/dissoc.spec.js
```diff
@@ -2,41 +2,91 @@ const test = require('tape')
...
+    x => dissoc(x, data)
+
+  t.equals(noInteger('string'), data, 'does nothing with truthy string as first argument')
+
+  const undefs = [ undefined, undefined ]
+  t.same(dissoc(0, undefs), [ undefined ], 'does not remove undefined values')
+
+  const defs = [ 34, 99, 'string' ]
+  t.same(dissoc(3, defs), defs, 'keeps items when index does not exist')
+  t.notEqual(dissoc(3, defs), defs, 'returns a new object')
```
dalefrancis88: nice work, wanted to make sure that this check was in and 💥 there it is! 👍

evilsoft:
image

PR #390, opened 2019-04-07
Update `src` dependencies to fixup audit

evilsoft:
## Rags to Riches
image

Been a little lax in keep the deps updated and a few vulnerabilities started to slip into the dev stack. This PR brings everything up to date, including `nodemon` which has fixed the issue it had on Linux for a wee bit there.

As part of the upgrade process, I went a head and re-ran the `all-contributors` script to apply the breaking changes to the generated output on the README.


evilsoft:
@dalefrancis88 if you get a chance, could you make sure this does not impact your Windows development, it seems to work for both Mac OSX and Linux.

package.json
```diff
@@ -7,7 +7,6 @@
     "contributors:add": "all-contributors add",
     "contributors:generate": "all-contributors generate",
     "setup": "node ./bin/setup.js",
-    "precommit": "lint-staged",
```
dalefrancis88: I assume this was just not being used?

PR #395, opened 2019-05-05
Move `propPathOr` to `getPathOr`

evilsoft:
## A Path To Enlightenment (or not)
image

As part of [this issue][issue], this PR addresses the item to move `propPathOr` to the new `getPathOr`. As with all of these moves, `propPathOr` is still available to maintain back-compat until we release `v1.0.0`.

Also fixed an error with the type siggy as part of this. We state we want a foldable, but due to the early exit optimization we added a while back, we need for the path to be specified with an `Array`.


docs/src/pages/docs/functions/helpers.md
```diff
@@ -340,6 +340,59 @@ you provide an `undefined` values for the second, that `Pair` will not be
...
+
+def([ 'foo', 'nan' ], data)
+//=> NaN
+
+def([ 'foo', 'undef' ], data)
+//=> "default"
+
+def([ 'arr', 'length' ], data)
+//=> 2
+```
```
dalefrancis88: The is no additional changes to the `#### propPathOr` section, would we want to mention there that this is now deprecated and reference the new name?

dalefrancis88: Also in that example, would we want to switch it to using the new name as well? These are just curiosity questions

docs/src/pages/docs/functions/index.md
```diff
@@ -172,7 +174,6 @@ type: `Pred a` and vice-versa
 [pipes]: helpers.html#pipes
 [prop]: ../crocks/Maybe.html#prop
 [proppath]: ../crocks/Maybe.html#proppath
-[proppathor]: helpers.html#proppathor
```
dalefrancis88: So if you take this out, wouldn't you want to remove the section in helpers?

src/helpers/getPathOr.js
```diff
@@ -0,0 +1,59 @@
...
+      return def
+    }
+
+    let value = target
+    for(let i = 0; i < keys.length; i++) {
+      const key = keys[i]
+
+      if(!(isString(key) && !isEmpty(key) || isInteger(key))) {
+        throw new TypeError(
+          `${name}: Array of Non-empty Strings or Integers required for second argument`
```
dalefrancis88: I'm sure that this is pedantic but with strings repeated twice, is the intention to always return the same error? if so would it be worth putting this outside the main function as a small function that takes the name and returns the error?

src/helpers/getPathOr.js
```diff
@@ -0,0 +1,59 @@
...
+        `${name}: Array of Non-empty Strings or Integers required for second argument`
+      )
+    }
+
+    if(isNil(target)) {
+      return def
+    }
+
+    let value = target
+    for(let i = 0; i < keys.length; i++) {
```
dalefrancis88: Is the point of avoiding a recursive function here to not grow the stack?

docs/src/pages/docs/functions/helpers.md
evilsoft: Ohhh great catches. Rebase issue with a stale branch.
Example will be switched and `propPathOr` will be removed from page entirely.
Will update

src/helpers/getPathOr.js
evilsoft: yep and for early exit.

src/helpers/getPathOr.js
evilsoft: I like it, will look into it and update!

evilsoft: Much better! Great feedback!

docs/src/pages/docs/functions/index.md
evilsoft: Section removed. Good 👀

docs/src/pages/docs/functions/helpers.md
evilsoft: Section removed.
Great Catch!

evilsoft:
@dalefrancis88 Okay I think I addressed your 🌽cerns, please throw some more 👀 at this if you wanna. And Thanks again!

docs/src/pages/docs/functions/index.md
```diff
@@ -70,6 +70,7 @@ need to account for for the rest of your flow.
 | [`fanout`][fanout] | `(a -> b) -> (a -> c) -> (a -> Pair b c)` | `crocks/Pair/fanout` |
 | [`find`][find] | <code>Foldable f => ((a -> Boolean) &#124; Pred) -> f a -> Maybe a</code> | `crocks/Maybe/find` |
 | [`fromPairs`][frompairs] | `Foldable f => f (Pair String a) -> Object` | `crocks/helpers/fromPairs` |
+| [`getPathOr`][getpathor] | <code>a -> f [ (String &#124; Integer) ] -> b -> a</code> | `crocks/helpers/getPathOr` |
```
evilsoft: need to remove that `f`

dalefrancis88:
image

PR #396, opened 2019-05-05
Remove `warning` option for uglifyjs to fix node 6 builds

evilsoft:
## Picard is Disappoint.
image

Looks like when upgrading we broke the build stack for node 6 testing due to how uglifyjs handles unrecognized options. This is cause a build failure for the node 6 stacks. This PR should resolve the failed builds.


evilsoft:
Once this passes I am going to override the approval process as it is a hotfix.

evilsoft:
image

PR #400, opened 2019-05-20
Move `prop` to `getProp`

evilsoft:
## Get Some Prop-er Etiquette 
image

As part of [this issue][issue], this PR addresses the item to move `prop` to the new `getProp`. It still allows for use of `prop` for now, but changes all documentation to speak only of `getProp`. . As with all of these moves, `prop` is still available to maintain back-compat until we release `v1.0.0`.


evilsoft:
image

PR #402, opened 2019-05-31
Update omit function signature

zhentian-wan:
Looks like `omit` should take an array of string, instead of just string


docs/src/pages/docs/functions/helpers.md
```diff
@@ -743,7 +743,7 @@ yourself constantly concatenating the result of this function into another
 `crocks/helpers/omit`
 
 ```haskell
-omit :: Foldable f => f String -> Object -> Object
+omit :: Foldable f => f ([ String ] | List String) -> Object -> Object
```
evilsoft: We should remove the `Foldable f => f` portion. At one time that working or was going to work with any foldable structure. Also `List` should be removed as this function only applies to Array at the moment. `List` does not have the required `indexOf`

evilsoft:
@zhentian-wan sorry I missed this on my first review, but the siggy should also be changed [here](https://github.com/evilsoft/crocks/blob/master/docs/src/pages/docs/functions/index.md) on the Functions index page.

Also if you want to add yourself to the contributors list, you should run the following in the project root and then commit the changes:
```sh
yarn contributors:add zhentian-wan doc
```

or using npm:
```sh
npm run contributors:add zhentian-wan doc
```

zhentian-wan:
> @zhentian-wan sorry I missed this on my first review, but the siggy should also be changed [here](https://github.com/evilsoft/crocks/blob/master/docs/src/pages/docs/functions/index.md) on the Functions index page.
> 
> Also if you want to add yourself to the contributors list, you should run the following in the project root and then commit the changes:
> 
> ```shell
> yarn contributors:add zhentian-wan doc
> ```
> 
> or using npm:
> 
> ```shell
> npm run contributors:add zhentian-wan doc
> ```

Thanks :)

evilsoft:
Merging!

zhentian-wan:
> Thank you so much for getting this in! Here have a squirrel for all your hard work!
> image

Thanks for accepting it :) Thanks for your hard work on crock.js and videos on Youtube and Egghead.io, I have learned a lot from those.

PR #403, opened 2019-06-02
Move `propPath` to `getPath`

evilsoft:
## Another Path to Traverse
image

This PR addresses the `propPath`portion of [this issue][issue] by deprecating `propPath` and replacing providing the same functionality, just in a new function to match the others called `getPath`


docs/src/pages/docs/crocks/Maybe.md
```diff
@@ -990,6 +990,65 @@ find(isEven, [ 1, 2, 3, 4, 5 ])
...
+
+```haskell
+getPath :: [ (String | Integer) ] -> a -> Maybe b
+```
+
+While [`getProp`](#getprop) is good for simple, single-level structures, there
+may come a time when you have to work with nested POJOs or Arrays. When you run
+into this situation, just pull in `getPath` (formally known as `propPath`) and
+pass it a left-to-right traversal path of keys, indices or a combination of
+both. This will kick you back a function that behaves just
```
bennypowers:
```suggestion
both as an array. This will kick you back a function that behaves just
```

unless we plan on supporting any foldable, why not be explicit?

docs/src/pages/docs/crocks/Maybe.md
```diff
@@ -990,6 +990,65 @@ find(isEven, [ 1, 2, 3, 4, 5 ])
...
+like [`getProp`](#getprop). You pass it some data, and it will attempt to
+resolve your provided path. If the path is valid, it will return the value
+residing there (`null` and `NaN` included!) in a `Just`. But, if at any point,
+that path "breaks", it will give you back a `Nothing`.
```
bennypowers:
```suggestion
that path traversal "breaks", it will give you back a `Nothing`.
```

src/Maybe/getPath.js
```diff
@@ -0,0 +1,55 @@
...
+const isDefined = require('../core/isDefined')
+const isEmpty = require('../core/isEmpty')
+const isInteger = require('../core/isInteger')
+const isNil = require('../core/isNil')
+const isString = require('../core/isString')
+
+function fn(name) {
+  function getPath(keys, target) {
+    if(!isArray(keys)) {
+      throw new TypeError(`${name}: Array of Non-empty Strings or Integers required for first argument`)
```
bennypowers:
```suggestion
      throw new TypeError(`${name}: First argument must be an Array of non-empty Strings or Integers`)
```

We use a mix of voices across the project in these errors. I prefer moving away from passive forms like "is required"

src/Maybe/getPath.js
```diff
@@ -0,0 +1,55 @@
...
+    if(isNil(target)) {
+      return Nothing()
+    }
+
+    let value = target
+    for(let i = 0; i < keys.length; i++) {
+      const key = keys[i]
+
+      if(!(isString(key) && !isEmpty(key) || isInteger(key))) {
+        throw new TypeError(`${name}: Array of Non-empty Strings or Integers required for first argument`)
```
bennypowers:
```suggestion
        throw new TypeError(`${name}: First argument must be an Array of non-empty Strings or Integers`)
```

src/Maybe/getPath.spec.js
```diff
@@ -0,0 +1,111 @@
...
+  t.equals(empty(null), 'nothing', 'returns Nothing when data is null')
+  t.equals(empty(NaN), 'nothing', 'returns Nothing when data is NaN')
+
+  t.end()
+})
+
+test('getPath errors', t => {
+  const fn = bindFunc(x => getPath(x, {}))
+
+  const err = /getPath: Array of Non-empty Strings or Integers required for first argument/
```
bennypowers:
```suggestion
  const err = /getPath: First argument must be an Array of non-empty Strings or Integers/
```

dalefrancis88:
I think the first two suggestions by @bennypowers are not important however I agree with the comments on the voice and tone of the error messages, i think they're good suggestions

evilsoft:
I like them all @bennypowers and @dalefrancis88 will implement tonight.
Thanks for dem 👀 !!

evilsoft:
Alrighty @bennypowers got dem changes in!
Thanks again for dem :eyes: you all are the best, I gotta say!

evilsoft:
I agree with the this voicing? should we make an issue to make error messaging :corn:sistant?
What are ya'lls thoughts on that?

bennypowers:
> I agree with the this voicing? should we make an issue to make error messaging 🌽sistant?
> What are ya'lls thoughts on that?

ack. recommend.

bennypowers:
https://github.com/evilsoft/crocks/issues/404

PR #405, opened 2019-06-05
Rewording Errors

### LINKED ISSUE #404: Errors should use a consistent voice
**Describe the bug**
Following on https://github.com/evilsoft/crocks/pull/403#discussion_r290143440, all errors should use a :corn:sistent voice, preferring active over passive voices where possible

Good:
"First argument must be a Number" 

Bad:
"Number is required for first argument"

This is also an opportunity to fix capitalization and other grammatical nits in error messages.

**To Reproduce**
Filed as bug since changing errors is a change to API, however there's no repro- it's strings.

bennypowers:
**This PR is a WIP, not intended to merge yet.** Rather it's intended to demonstrate the direction I intend to head in and to seek review and critique before committing to these changes on a larger scale.

Closes #404 

## Rationales

These new error messages follow an [SVO word order](https://www.wikiwand.com/en/Subject%E2%80%93verb%E2%80%93object), which is likely to be easier to understand. While more verbose, it is also clearer, explaining that the *function argument* is the thing that is required to be the `${REQUIRED_THING}`

In the case of unary functions I opted for brevity, so not "Unary argument must be..." or "The single argument..." etc.

I opted to capitalize type names like `Async` and `Function`, even in the plural case like `Function` and even for types which don't exist in either JavaScript or Crocks like `Integer`

I also noticed a few regexps in the tests that were underspecified, and corrected them. No big shakes there, really, but why not.


dalefrancis88:
I think this is a good direction to go in and a good tone

dalefrancis88:
@bennypowers can you think of a reason this was never merged other than a lack of review?

bennypowers:
nope should still be GTG

src/Async/index.js
```diff
@@ -131,7 +131,7 @@ function Async(fn) {
 
   function fork(reject, resolve, cleanup) {
     if(!isFunction(reject) || !isFunction(resolve)) {
-      throw new TypeError('Async.fork: Reject and resolve functions required')
+      throw new TypeError('Async.fork: Resolve and reject arguments must both be Functions')
```
dalefrancis88: What is the purpose in switching the order here. I can see that originally it was Reject then Resolve in the word order because that matches the order of the arguments

dalefrancis88 (commented):
I think that this is actually GTG @bennypowers but just have the one nitpick, lmk what you think?

src/Async/index.js
bennypowers: no preference, IIRC my original issue was SVO phrasing

dalefrancis88: Lets go ahead and change the order, if you like i'm happy to make the change myself then we can merge this in

bennypowers: Yes, please, and thank you :)

PR #408, opened 2019-06-11
Fix Either constructor when called with object containing a circular reference

jonwhelan:
Introduced in release 0.9.4 with commit https://github.com/evilsoft/crocks/commit/6bc5377cf4a47a927ff3fa407073d17aa3b0cdbe#diff-d70a7c87b679a4d796e121bd30e12fd1 the `inspect` function recursively stringifies objects which will blow the callstack when given an object containing a circular reference.

This bug is triggered when constructing an Either due to its `inspect` method being eagerly evaluated during construction.

### DIFF
--- a/src/Either/Either.spec.js
+++ b/src/Either/Either.spec.js
@@ -25,6 +25,14 @@ const applyTo =
 const constant = x => () => x
 const identity = x => x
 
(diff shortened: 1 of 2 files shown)


evilsoft:
Ohhh this is nice. 👀

PR #409, opened 2019-06-16
add init and last to pointfree

RichardForrester:
Hi, enjoying Crocks here.

I had need of `init` and `last` and I wanted to get more familiar with the Crocks library so I forked and used the existing `head` and `tail` as templates and tried to stay true to the existing programming style.

Just like Haskell and Ramda:

`last` is the opposite of `head`
`init` is the opposite of `tail`

Lines 21 - 38 of `last` are not exactly my style so a review of that may be desirable - I don't think there's a problem, but because of the need to cycle through `next` until the iterator returns `done` there is at least the potential for a nasty bug.


src/pointfree/last.js
```diff
@@ -0,0 +1,43 @@
+/** @license ISC License (c) copyright 2016 original and current authors */
+/** @author Ian Hofmann-Hicks (evil) */
+
+const isArray = require('../core/isArray')
+const isFunction = require('../core/isFunction')
+const isIterable = require('../core/isIterable')
+const isString = require('../core/isString')
+const cloneIterable = require('../core/cloneIterable')
```
dalefrancis88: I get this is copied from head, but please move clone to the top of the list require/imports are done in alphabetical order. Feel free to change head too if you like :)

docs/src/pages/docs/functions/pointfree-functions.md
```diff
@@ -2,7 +2,7 @@
 title: "Point-free Functions"
 description: "Point-free Functions API"
 layout: "notopic"
-functions: ["alt", "ap", "bimap", "both", "chain", "coalesce", "comparewith", "concat", "cons", "contramap", "either", "empty", "equals", "extend", "filter", "first", "fold", "foldmap", "head", "map", "merge", "option", "promap", "race", "reduce", "reduceright", "run", "runwith", "second", "sequence", "swap", "tail", "traverse", "valueof"]
+functions: ["alt", "ap", "bimap", "both", "chain", "coalesce", "comparewith", "concat", "cons", "contramap", "either", "empty", "equals", "extend", "filter", "first", "fold", "foldmap", "head", "init", "last", "map", "merge", "option", "promap", "race", "reduce", "reduceright", "run", "runwith", "second", "sequence", "swap", "tail", "traverse", "valueof"]
 weight: 50
 ---
```
dalefrancis88: You need to add the signatures in this section

src/pointfree/index.js
```diff
@@ -18,6 +18,8 @@ module.exports = {
   fold: require('./fold'),
   foldMap: require('./foldMap'),
   head: require('./head'),
+  init: require('./init'),
+  last: require('./last'),
```
dalefrancis88: These need to be added to the index.spec.js for the pointfree index. 
You'll also need to add these to the tests for the root `src/index.spec.js`

src/pointfree/init.js
```diff
@@ -0,0 +1,25 @@
+/** @license ISC License (c) copyright 2016 original and current authors */
+/** @author Ian Hofmann-Hicks (evil) */
```
bennypowers: why not give yourself credit for creating the file

src/pointfree/init.js
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    if(isFunction(m.slice)) {
+      return m.length < 2
+        ? Nothing()
+        : Just(m.slice(0, -1))
+    }
+  }
+
+  throw new TypeError('init: Array, String, or List required')
```
bennypowers: please see #405  WRT to the preferred voice for errors

```suggestion
  throw new TypeError('init: Argument must be an Array, String, or List')
```

src/pointfree/init.spec.js
```diff
@@ -0,0 +1,80 @@
...
+const constant = x => () => x
+
+const init = require('./init')
+
+test('init pointfree', t => {
+  const f = bindFunc(init)
+
+  t.ok(isFunction(init), 'is a function')
+
+  const err = /init: Array, String, or List required/
```
bennypowers:
```suggestion
  const err = /init: Argument must be an Array, String, or List/
```

src/pointfree/last.js
```diff
@@ -0,0 +1,43 @@
...
+    let val
+    while (!curr.done) {
+      val = curr.value
+      curr = iterator.next()
+    }
+
+    return Just(val)
+  }
+
+  throw new TypeError('last: List, String, or Iterable required')
```
bennypowers:
```suggestion
  throw new TypeError('last: Argument should be a List, String, or Iterable')
```

src/pointfree/last.spec.js
```diff
@@ -0,0 +1,90 @@
...
+const constant = x => () => x
+
+const last = require('./last')
+
+test('last pointfree', t => {
+  const f = bindFunc(last)
+
+  t.ok(isFunction(last), 'is a function')
+
+  const err = /last: List, String, or Iterable required/
```
bennypowers:
```suggestion
  const err = /last: Argument should be a List, String, or Iterable/
```

src/pointfree/init.js
dalefrancis88: Good catch @bennypowers 😄 @RichardForrester as a standard we accredit the original author for each file at the top

RichardForrester:
Thanks for all the good feedback. I was hoping to wrap this up today but got busy at work. I’m now on a camping trip until Thursday, so I’ll incorporate all soon after that. Thanks!

RichardForrester:
Okay, it should be ready. Let me know if there's anything else.

Thanks for letting me contribute.

RichardForrester:
Forgive me if this has already been discussed, but I don't understand why Crocks is throwing errors.

Take `head` for example:

```javascript
function head(m) {
  // if m is acceptable type
  // do stuff and return a Just or Nothing

  // ERROR
  throw new TypeError('head: List or iterable required')
}
```
I see that `throw` and my first thought is that I really do not want that in my production code, especially where I may not have control over input types, e.g., fetches from DB or web API. In fact, that's exactly the reason I'm reaching for a Maybe type to begin with. Of course, a type error is not the same as an empty array, and I *do* want to be notified that something has gone far awry, but I don't want my whole app to crash or my dev cycle to be halted. I would much prefer something like:

```javascript
function head(m) {
  // if m is acceptable type
  // do stuff and return a Just or Nothing

  // ERROR
  console.error('head: Argument must be an Array, Iterable, or String')
  return Nothing()
}
```

Even better would be something like:

```javascript
function head(m) {
  // if m is acceptable type
  // do stuff and return a Just or Nothing

  // ERROR
  if (process.env.NODE_ENV === 'production') {
    reportProdError('head: Argument must be an Array, Iterable, or String', Error())
  } else {
  reportDevError('head: Argument must be an Array, Iterable, or String', Error())
  }

  return Nothing()
}
```
In a compiled language like Haskell it is true that the compiler will throw errors at you for calling a function with an argument of the wrong type, but assuming you get the code to compile, you wouldn't intentionally cause a runtime error. 

Is there something simple I'm missing here?

evilsoft:
> Is there something simple I'm missing here?

Nope you are not missing anything. This is all part of defining a good edge in your program. Then you can put that env check specific to your do main at the edge by using `tryCatch` This is not a compiled language and there are many ways that people want to deal with these exceptions, depending on environment. So as to not lock anyone into a way of handling errors, we throw.

So if the code is done "right" there should be no reason that the type requirements  are not met in production. The only thing of variance (hopefully) is IO into your flow. Also if you are using things like Async, their are Natural Transformations from `Result` to `Async`, so they can be used when chaining in things like reading from a file, or making a web request, etc.

Hope that makes sense.

Short answer: We leave it up to the end user and their domain on how they want to deal with the errors, instead of making the decision for them.

src/pointfree/last.js
```diff
@@ -37,7 +37,7 @@ function last(m) {
     return Just(val)
   }
 
-  throw new TypeError('last: List, String, or Iterable required')
+  throw new TypeError('last: Argument should be a List, String, or Iterable')
```
bennypowers: s/should/must/g

src/pointfree/last.spec.js
```diff
@@ -17,7 +17,7 @@ test('last pointfree', t => {
 
   t.ok(isFunction(last), 'is a function')
 
-  const err = /last: List, String, or Iterable required/
+  const err = /last: Argument should be a List, String, or Iterable/
```
bennypowers: same

docs/src/pages/docs/functions/pointfree-functions.md
RichardForrester: Also to the Datatypes correct?

.all-contributorsrc
```diff
@@ -241,6 +241,17 @@
...
+    {
+      "login": "RichardForrester",
+      "name": "RichardForrester",
+      "avatar_url": "https://avatars0.githubusercontent.com/u/12902182?v=4",
+      "profile": "https://github.com/RichardForrester",
+      "contributions": [
+        "code",
+        "doc",
+        "test"
+      ]
```
dalefrancis88: 👍

src/pointfree/last.js
```diff
@@ -0,0 +1,43 @@
+/** @license ISC License (c) copyright 2016 original and current authors */
```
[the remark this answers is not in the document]
dalefrancis88: Good catches on these, great 👁 for detail

RichardForrester:
@evilsoft I'm sure it would have taken you about 15 mins, but thanks for letting me contribute. I like the library a lot and am learning quite a bit by checking out the setup.

PR #411, opened 2019-06-23
add docs.crocks.dev hostname to docs

bennypowers:
I purchased the domain [crocks.dev](https://crocks.dev). It's currently pointed to Github's servers. This PR should (I think) let `docs.crocks.dev` point to the docs page.

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+docs.crocks.dev


bennypowers:
open questions:
- is this even a good idea?
- is `crocks.dev` preferable to `docs.crocks.dev`
- should this point to somewhere other than github? i.e. a netlify instance etc.?

dalefrancis88:
hmmmm if we did `docks.crocks...` we could do `blog.crocks...` and `guides.crocks...`
but as a counter argument what problem are we solving here?

bennypowers:
two problems as i see them
1. evilsoft.github.io/crocks is more to type
2. evilsoft.github.io/crocks has two nouns before it gets to crocks, and crocks is a subdirectory
  - so, for 'seo' and to communicate more clearly that 'this is the crocks site', `crocks.dev`is better

non-medicinal ingredients:
- .dev domains are cool

evilsoft:
Wow, thanks for this @bennypowers I agree this would be much better.
I do not think we need to do the CNAME change, github pages takes care of a lot of that. We may need to :corn:figure your ALIAS or ANAME record on the DNS and I need to look into getting an SSL Cert for the domain.

I will have some time this week to look into this. This is amazing, do you want some reimbursement or anything? I am happy to help or absorb this cost!

bennypowers:
SSL comes built in at GitHub, but you have to tick something off in settings.

tl;dr: dm me (Gitter? Twitter? I accept any social media outlet that ends in 'itter') and I'll set you up on the Google domains account for the name.

dalefrancis88:
@bennypowers  Should we still keep this open? it appears to be done?

bennypowers:
LGTM nice work everyone

PR #412, opened 2019-06-24
Add maybeToList and maybeToArray transformations

evilsoft:
## Just two handy Transformers...well Transformations...
image

When working through a couple example of how to filter and transform without using reduce on `Array`s and `List`s, I found that a lot of value could be adding with some `maybeToX` transformations. The basically will take a `Maybe` to either a `List` or an `Array`. A `Just` instance will return a single element structure, containing the value in the original `Just`.  While a `Nothing` returns an empty structure.

When used with `chain` this can be used to filter and transform a structures values in one go, without having to set up a `mapReduce` with `alt` to get the same result.

In addition to adding these functions and their documentation, Also added a couple links to the `Result` NTs and cleaned up some specs that suffered from :policeman:-y :spaghetti: artifacts.


evilsoft:
image

PR #414, opened 2019-07-05
Update CI to remove node-6 and add node-12

evilsoft:
## I guess ☝️
image

What it says on the tin. We only support the last three LTS versions. This drops 6 and adds 12.


evilsoft:
What no squirrel?

PR #417, opened 2019-07-08
Adding new `hasProps` predicate

dalefrancis88:
This will allow us to pass an array of keys instead of a single key to check the structure of an object.

Have some things to point out

1. Are we ok with the use of `every`? I'm not married to it and am happy to convert it to reduce
2. Right now it doesn't support taking either the Array or the single value, do we want to do that?
3. Is that the correct way to specify the expected values in an Array? I see that everywhere else we're using just Array but it'd be nice to be more specific in the signature

### DIFF
--- a/docs/src/pages/docs/functions/predicate-functions.md
+++ b/docs/src/pages/docs/functions/predicate-functions.md
@@ -16,7 +16,7 @@ Below is a list of all the current predicates that are included with a
 description of their truth:
 
 * `hasProp :: (String | Integer) -> a -> Boolean`: an `Array` or `Object` that contains the provided index or key
-* `hasProps :: Array (String | Integer) -> a -> Boolean`: an `Array` or `Object` that contains the provided indexs or keys
+* `hasProps :: Foldable f => f (String | Integer) -> a -> Boolean`: an `Array` or `Object` that contains the provided indexs or keys
 * `hasPropPath :: [ String | Integer ] -> a -> Boolean`: an `Array` or `Object` that contains the provided index path
 * `isAlt :: a -> Boolean`: an ADT that provides `map` and `alt` methods
 * `isAlternative :: a -> Boolean`: an ADT that provides `alt`, `zero`, `map`, `ap`, `chain` and `of`methods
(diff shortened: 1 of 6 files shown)


src/predicates/hasProps.js
```diff
@@ -0,0 +1,29 @@
...
+
+const curry = require('../core/curry')
+const isArray = require('../core/isArray')
+const isDefined = require('../core/isDefined')
+const isEmpty = require('../core/isEmpty')
+const isInteger = require('../core/isInteger')
+const isNil = require('../core/isNil')
+
+const isValid = keys =>
+  isArray(keys) && !isEmpty(keys) && keys.every(key => !isEmpty(key) || isInteger(key))
```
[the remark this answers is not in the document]
evilsoft: Also think this should check for Positive Integers, like other path/prop functions do.

dalefrancis88:
Yeah i am much more leaning towards the foldable. Expecially since we're trying to increase support for foldable over arrays. Thanks for the suggestion, i'll make the updates and resubmit

dalefrancis88:
Do we want to accept an empty array and have it return true? I think that's perfectly acceptable

src/predicates/hasProps.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  return isDefined(obj[key])
+}
+
+// every :: (a -> Boolean) -> ((Null | Boolean), a) -> Boolean
+const every = fn => (acc, x) =>
+  (acc === null ? true : acc) && fn(x)
+
+// hasProps :: Foldable f -> a -> Boolean
```
evilsoft: Should be something like:
```js
// hasProps :: Foldable f => f (String | Integer) -> a -> Boolean
```

docs/src/pages/docs/functions/predicate-functions.md
```diff
@@ -16,6 +16,7 @@ Below is a list of all the current predicates that are included with a
 description of their truth:
 
 * `hasProp :: (String | Integer) -> a -> Boolean`: an `Array` or `Object` that contains the provided index or key
+* `hasProps :: Array (String | Integer) -> a -> Boolean`: an `Array` or `Object` that contains the provided indexs or keys
```
evilsoft: This signature should be:
```js
// hasProps :: Foldable f => f (String | Integer) -> a -> Boolean
```

Issue #360, opened 2018-12-13
CurryN could support curried and manually curried functions


dalefrancis88:
**Is your feature request related to a problem? Please describe.**
The feature is un-related to a problem, however was noticed when playing around with `nAry` functions like `binary` and `unary` that they don't support functions that are curried

**Describe the solution you'd like**
I think it would be very possible to update the `curryN` implementation to support curried functions by applying undefined until a non-func is returned

**Describe alternatives for how you do this now**
I think the only alternative at the moment is to create a non-curried wrapper on the curried function, however i'm unsure as this issue has arisen in my main code-base 

**Code**
Example code for a potential solution can be found here [Repl](https://repl.it/@DaleFrancis/CurryN-Changes)

dalefrancis88:
After a discussion it was decided not to pursue this. We will re-address if anyone else raises an issue

PR #421, opened 2019-07-15
Adding in support for Auto and manually curried functions unary/binary/nary

dalefrancis88:
This is work to better add support for all function types.

This will address this issue, https://github.com/evilsoft/crocks/issues/360

This will sequentially apply the arguments and then recursively apply undefined until a non-function value is returned.

@evilsoft this is the work i promised a while back, i think it's the nicest i can make it for what its doing. What i'd like to know is could i make use of Function.bind or fn.apply here to clean it up more? googleing did not shed much light on the difference
Also i can't get the test to pass but i think it may be more around how the test is written, thoughts?

### DIFF
--- a/src/core/curryN.js
+++ b/src/core/curryN.js
@@ -3,22 +3,25 @@
 
 const isFunction = require('./isFunction')
 
-const reduceApply = (fn, args) =>
-  args.reduce((f, arg) => isFunction(f) ? f(arg) : f, fn)
+const applyArgs = (fn, args) => {
+  const toApply = args.splice(0, fn.length || args.length)
 
-const finalizeResult = a =>
-  isFunction(a) ? finalizeResult(a()) : a
+  return isFunction(fn) && toApply.length ? applyArgs(fn.apply(null, toApply), args) : fn
+}
+
+const applyUntil = a =>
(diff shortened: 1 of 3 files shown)


dalefrancis88:
@evilsoft check this out if you like but I realised as I was falling asleep that I broke this for variadic functions. Will fix it

src/core/curryN.js
```diff
@@ -1,17 +1,27 @@
...
 
+const isFunction = require('./isFunction')
+
+const applyArgs = (fn, args) => {
+  const toApply = args.splice(0, fn.length || args.length)
+
+  return isFunction(fn) && toApply.length ? applyArgs(fn.apply(null, toApply), args) : fn
+}
+
+const applyUntil = a =>
```
dalefrancis88: not sure on the name for this, `applyUndefinedUntil` ?

dalefrancis88:
kk @evilsoft this one was painful, i understand your original hesitation now, this is committed to start discussion, going to let my mind think through ways to improve but would love some 👀

dalefrancis88:
Looks like they'res some issues with liftN i'm going to have to address

dalefrancis88:
We're going to approach this a different way

Issue #338, opened 2018-10-27
Have all `compose`/`pipe` functions return only unary functions


evilsoft:
### Description/Location

**location:** './helpers/aThing.js'
**description:**
The current implementation allows the ability to apply any number of arguments to the head function in the composition/pipe. While this is very :corn:venient at times, it has been know to lead to hard to reason about compositions and many time unnecessary  argument juggling. It also provides limits on how a composed function is curried.

By limiting this to just a unary function, it removes the ability to get into a juggling mess and can also allow us to properly curry the function that is returned for cases like:

```js
// pluck :: String -> [ a ] -> Maybe b
const pluck = curry(
  key => map(prop(key))
)

pluck('id', [ { id: 2 } ])
//=> [ Just 2 ]
```

becomes:
```js
// pluck :: String -> [ a ] -> Maybe b
const pluck = 
 compose(map, prop)
```
While is maybe different then what we are used to from the standard in JS in most libs, if you look at the code and remember that `x => f(g(x))` is the same thing as `compose(f, g)` see how these are actually equivalent. Now people who for some reason have not adopted compose and STILL use pipe for one reason or another, it may be harder for this to click when writing and maintaining code. But as shown above, these are equivalent and should be allowed to be implemented that way.

This is more of a proposal to get peoples opinions and thoughts. Would love to get some emoji reactions to this, and if possible a comment on why you feel the way you do.

### Task List

* [ ] Build out a task list if enough people understand why and the implications.

For more information, see the `CONTRIBUTING` guide.

dalefrancis88:
I think to help with upgrading and reducing the pain a little, could we keep the previous implementation with some other name? we could have a `composeN` that takes the number of initial arguments to pass in, then the rest get fed as normal. So if people want to keep there behaviour it won't require a large refactor because it's highly likely people have done a lot of args juggling

Signature (this sucks, i don't know these sigs well)
`composeN :: n -> [fns] -> * -> * -> a`

karthikiyengar:
@dalefrancis88 - I don't think that this change would affect the people that have done args juggling as it's only enforcing that the head to be unary. 

However, it does get in the way of some other use cases. A contrived example follows:

```javascript
 compose(ap(Maybe.of(10)), Maybe.of, add)(5)
//=> Just 15 (current implementation)
```
Our new implementation however, would result in an exception. 

Personally, I dig the simplification, but I am not sure if I completely understand how it still remains lawful. Could you please elaborate @evilsoft? 

[...]

PR #422, opened 2019-07-15
Adding `curry` to `pipe` and `compose`

dalefrancis88:
Nearly all of our functions export a curried function when complete except for our `compose` and `pipe` functions. If we `curry` the result then we get https://github.com/evilsoft/crocks/issues/338

If we do something like this we get the power of these functions as is, along with the expected behaviour of a few people in the order of argument application with non-unary invocation of the resultant composition

Also testing this implementation has it not appear to be a breaking change


dalefrancis88:
With this addition of `compose2` we've decided to hold off on this for now until other issues come up

PR #429, opened 2019-07-17
Adding in the powerful P combinator

dalefrancis88:
This combinator, once i wrapped my head around it, has become one of my go to tools since. Like `converge` it allows for some really nice/neat compositions for inputs into those `binary` functions we all love :)


dalefrancis88:
@evilsoft I'm hoping to get this into the next release along with `bichain` but if there are issues, don't hesitate to release `bichain` this can wait

docs/src/pages/docs/functions/combinators.md
```diff
@@ -309,6 +309,90 @@ with this library. It quite simply is just a function that when you pass it
 something, it returns that thing right back to you. So simple, I will leave it
 as an exercise to reason about why this is so powerful and important.
 
+#### psi
```
bennypowers: for the other combinators, we've in the past prefered descriptive names over conventional ones. (i.e. 'flip' vs 'kestrel' or 'C')

in this case, maybe we'd prefer 'on' ?

docs/src/pages/docs/functions/combinators.md
```diff
@@ -309,6 +309,90 @@ with this library. It quite simply is just a function that when you pass it
...
+
+`crocks/combinators/psi`
+
+```haskell
+psi ::  (b -> b -> c) -> (a -> b) -> a -> a -> c
+```
+
+`psi` is a function that can be considered the sister of [`converge`](#converge).
+Where [`converge`](#converge) takes one argument an maps it through two functions,
```
bennypowers:
```suggestion
Where [`converge`](#converge) takes one argument and maps it through two functions,
```

docs/src/pages/docs/functions/combinators.md
```diff
@@ -309,6 +309,90 @@ with this library. It quite simply is just a function that when you pass it
...
+merging the resulting values with an accumulating function, `psi` takes two
```
bennypowers:
```suggestion
merging the resulting values with an binary combination function, `psi` takes two
```

docs/src/pages/docs/functions/combinators.md
```diff
@@ -309,6 +309,90 @@ with this library. It quite simply is just a function that when you pass it
...
+arguments and runs them through the same function individually before merging
```
bennypowers:
```suggestion
arguments and runs each through the same unary function before merging
```

docs/src/pages/docs/functions/combinators.md
```diff
@@ -309,6 +309,90 @@ with this library. It quite simply is just a function that when you pass it
...
+them with the given accumulating function.
```
bennypowers:
```suggestion
them with the given binary combination function.
```

docs/src/pages/docs/functions/combinators.md
bennypowers: `binary` is important. If there are other instances in docs where `accumulating` is preferred over `combining`, or if there is some other term, use that.

docs/src/pages/docs/functions/combinators.md
```diff
@@ -309,6 +309,90 @@ with this library. It quite simply is just a function that when you pass it
...
+
+`psi` is often used to easily [`compose`][compose] equality checking functions or when
```
bennypowers:
```suggestion
`psi` is often used to [`compose`][compose] equality checking functions or when
```

docs/src/pages/docs/functions/combinators.md
```diff
@@ -309,6 +309,90 @@ with this library. It quite simply is just a function that when you pass it
...
+import and from 'crocks/logic/and'
+import equals from 'crocks/pointfree/equals'
+import isNumber from 'crocks/predicates/isNumber'
+import liftA2 from 'crocks/helpers/liftA2'
+import safe from 'crocks/Maybe/safe'
+
+// isNonZero :: Number -> Boolean
+const isNonZero = x => x !== 0
+
+// isValid :: Number -> Boolean
```
bennypowers:
```suggestion:+2
// isValidDivisor :: Number -> Boolean
const isValidDivisor = and(isNumber, isNonZero) 
```

docs/src/pages/docs/functions/combinators.md
```diff
@@ -309,6 +309,90 @@ with this library. It quite simply is just a function that when you pass it
...
+const isValid = and(isNumber, isNonZero)
+
+// divideBy :: Number -> Number -> Number
+const divideBy = x => y => y / x
+
+// safeDivide :: Number -> Number -> Maybe Number
+const safeDivide =
+  psi(liftA2(divideBy), safe(isValid))
```
bennypowers:
```suggestion
  psi(liftA2(divideBy), safe(isValidDivisor))
```

docs/src/pages/docs/functions/combinators.md
```diff
@@ -309,6 +309,90 @@ with this library. It quite simply is just a function that when you pass it
...
+safeDivide('0.5', 21)
+//=> Nothing
+
+safeDivide(0.5, '21')
+//=> Nothing
+
+safeDivide(29, 0)
+//=> Nothing
+
+// toTitleCase :: String -> String
```
bennypowers:
```suggestion:+2
// capitalize :: String -> String
const capitalize = str => 
```

docs/src/pages/docs/functions/combinators.md
```diff
@@ -309,6 +309,90 @@ with this library. It quite simply is just a function that when you pass it
...
+const toTitleCase = str =>
+  `${str.charAt(0).toUpperCase()}${str.slice(1)}`
+
+// join :: String -> String -> String -> String
+const join = delim => right => left =>
+  `${left}${delim}${right}`
+
+// createName :: String -> String -> String
+const createName =
+  psi(join(', '), toTitleCase)
```
bennypowers:
```suggestion
  psi(join(', '), capitalize)
```

docs/src/pages/docs/functions/combinators.md
```diff
@@ -309,6 +309,90 @@ with this library. It quite simply is just a function that when you pass it
...
+
+createName('sara', 'smith')
+//=> Smith, Sara
+
+// toLowerCase :: String -> String
+const toLowerCase = str =>
+  str.toLowerCase()
+
+// equalsIgnoreCase :: String -> String -> Boolean
+const equalsIgnoreCase =
```
bennypowers: great example this really drives the use case home!

docs/src/pages/docs/functions/combinators.md
dalefrancis88: Yeah i know `on` is what this is called in Haskell and Sanctuary but i don't find it to be that descriptive honestly, I got the original reference and name from [here](https://gist.github.com/Avaq/1f0636ec5c8d6aed2e45) we seem to have matched the names with most everything else except `thrush`

docs/src/pages/docs/functions/combinators.md
dalefrancis88: So this is actually similiar to what i wrote originally, my preference is `binary` over `accumulating` however i changed this to match the phrasing of `converge` and have a common voice. 
Should we make a point of changing the docs for `converge`?

dalefrancis88: I might also drop combination from your suggestion, i'm not sure it's neccessary

docs/src/pages/docs/functions/combinators.md
dalefrancis88: Yeah equality checking is really where `psi` earns it's stripes

docs/src/pages/docs/functions/combinators.md
bennypowers: looks like a solid precedent. Thanks for addressing this.

docs/src/pages/docs/functions/combinators.md
bennypowers: sounds good to me. To my ear, binary is clearer than accumulating, so i'd prefer to see that jargon adopted globally @evilsoft ?

dalefrancis88: Yeah i would agree 100%

PR #430, opened 2019-07-20
Changing toPairs to return an array

dalefrancis88:
For merging into 0.13

This work is to start the journey to complete https://github.com/evilsoft/crocks/issues/336

i'm unsure if the type sig for arr is `[ String ]` or `Array String`

### DIFF
--- a/src/Pair/toPairs.js
+++ b/src/Pair/toPairs.js
@@ -1,21 +1,20 @@
 /** @license ISC License (c) copyright 2017 original and current authors */
 /** @author Ian Hofmann-Hicks (evil) */
 
-const List = require('../core/List')
 const Pair = require('../core/Pair')
 const isObject = require('../core/isObject')
 
-// toPairs : Object -> List (Pair String a)
+// toPairs : Object -> Array (Pair String a)
 function toPairs(obj) {
   if(!isObject(obj)) {
     throw new TypeError('toPairs: Object required for argument')
   }
 
-  return Object.keys(obj).reduce(
-    (acc, key) => obj[key] !== undefined
-      ? acc.concat(List.of(Pair(key, obj[key])))
+  return Object.entries(obj).reduce(
+    (acc, [ key, value ]) => value !== undefined
+      ? acc.concat([ Pair(key, value) ])
       : acc,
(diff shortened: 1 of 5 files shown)


src/Pair/toPairs.js
```diff
@@ -1,21 +1,20 @@
...
+// toPairs : Object -> Array (Pair String a)
 function toPairs(obj) {
   if(!isObject(obj)) {
     throw new TypeError('toPairs: Object required for argument')
   }
 
-  return Object.keys(obj).reduce(
-    (acc, key) => obj[key] !== undefined
-      ? acc.concat(List.of(Pair(key, obj[key])))
+  return Object.entries(obj).reduce(
```
dalefrancis88: entries is supported by most, if not all, browsers and Node these days, but let me know if this should go back to keys

src/Pair/toPairs.js
```diff
@@ -1,21 +1,20 @@
 /** @license ISC License (c) copyright 2017 original and current authors */
 /** @author Ian Hofmann-Hicks (evil) */
 
-const List = require('../core/List')
 const Pair = require('../core/Pair')
 const isObject = require('../core/isObject')
 
-// toPairs : Object -> List (Pair String a)
+// toPairs : Object -> Array (Pair String a)
```
dalefrancis88: `[ (Pair String a) ]` or `Array (Pair String a)`?

src/Pair/toPairs.js
evilsoft: Thinking we should go back to keys. JIC

src/Pair/toPairs.js
evilsoft: I am good with how you have it, reads better IMO

dalefrancis88: I agree

dalefrancis88: I agree

dalefrancis88:
The idea of the 0.13 is to get the work for #336 completed and any other breaking changes but still merging/closing the branches. I think that work may take a bit of time so breaking it up incase multiple people want to work on pieces.

Issue #428, opened 2019-07-16
Adding in the `compose2`/`binaryCompose` combinator


dalefrancis88:
**Feature Request**
Adding in the `compose2`/`binaryCompose` combinator would add some real nice power to the system. I think we could also overload it to support taking a binary array or a `Pair` 

https://hackage.haskell.org/package/util-0.1.13.0/docs/Util.html#v:compose2

**Describe alternatives for how you do this now**
Have it in my local codebase

**Code**
```
const compose2 = f => g => h => x => y => f(g(x))(h(y))
```

**Additional context**

evilsoft:
I am all for the addition of this, just as is, we can think about the Pair/Array application later.

dalefrancis88:
Closing as will be implemented on #431, can continue any discussion there as needed :)

PR #431, opened 2019-07-21
Adding in the `compose2` combinator

dalefrancis88:
per the discussion on gitter and https://github.com/evilsoft/crocks/issues/428 this is to add in the `compose2` combinator

This will allow for some nice composition


docs/src/pages/docs/functions/combinators.md
```diff
@@ -64,6 +64,119 @@ safeYell(42)
...
+
+```haskell
+compose2 :: (c -> d -> e) -> (a -> c) -> (b -> d) -> a -> b -> e
+```
+
+`compose2` allows for composition between a `binary` function and
+two `unary` functions. `compose2` takes a `binary` function followed by
+two `unary` functions and returns a `binary` function that maps the first
+argument with the first `unary` and the second with the second, passing
+the results to the given `binary` and returning the result.
```
dalefrancis88: Really not married to this description, it feels a little flat, lmk if anyone can make it better

docs/src/pages/docs/functions/combinators.md
```diff
@@ -64,6 +64,119 @@ safeYell(42)
...
+//=> Nothing
+
+// Item :: { id: Integer }
+// Items :: Array Item
+const items =
+  [ { id: 2 }, { id: 1 } ]
+
+// pluck :: String -> Array Object -> Maybe a
+const pluck =
+  compose2(applyTo, prop, flip(map))
```
dalefrancis88: This is to show usage in relation to https://github.com/evilsoft/crocks/issues/338

docs/src/pages/docs/functions/combinators.md
evilsoft: I think this is good as is, gets to the point and provides the info someone would need to wrap their head around it.

docs/src/pages/docs/functions/combinators.md
```diff
@@ -64,6 +64,119 @@ safeYell(42)
 //=> Left 42
 ```
 
+#### compose2
```
evilsoft: This should come before `composeB`

docs/src/pages/docs/functions/combinators.md
```diff
@@ -64,6 +64,119 @@ safeYell(42)
...
+import hasProp from 'crocks/predicates/hasProp'
+import isNumber from 'crocks/predicates/isNumber'
+import liftA2 from 'crocks/helpers/liftA2'
+import map from 'crocks/pointfree/map'
+import prop from 'crocks/Maybe/prop'
+import safe from 'crocks/Maybe/safe'
+import safeLift from 'crocks/Maybe/safeLift'
+
+// isNonZero :: Number -> Boolean
+const isNonZero = x => x !== 0
```
evilsoft: These should all be multi-line
```js
// isNonZero :: Number -> Boolean
const isNonZero = x =>
  x !== 0

// isValidDivisor :: Number -> Boolean
const isValidDivisor =
  and(isNumber, isNonZero)

// divideBy :: Number -> Number -> Number
const divideBy = x => y =>
  y / x

// safeDivide :: Number -> Number -> Maybe Number
const safeDivide = compose2(
  liftA2(divideBy),
  safe(isValidDivisor),
  safe(isNumber)
)
```

docs/src/pages/docs/functions/combinators.md
```diff
@@ -64,6 +64,119 @@ safeYell(42)
...
+const createSummary = compose2(
+  liftA2(summarize),
+  prop('name'),
+  getLength
+)
+
+createSummary({
+  name: 'Sam Smith'
+}, items)
+// => Just "Sam Smith purchased 2 items"
```
evilsoft: extra space after comment

src/combinators/compose2.spec.js
```diff
@@ -0,0 +1,59 @@
...
+  const fn = bindFunc(compose2)
+  const f = x => y => x * y
+  const g = x => x - 1
+  const h = x => x + 1
+  const x = 22
+  const y = 1
+
+  t.ok(isFunction(compose2), 'is a function')
+
+  const err = /TypeError: compose2: First, second and third arguments must be functions/
```
evilsoft: Interesting, including `TypeError`. Thinking we should do this for all cases.
should this include the `^` at the start, JIC.
```js
const err = /^TypeError: compose2: First, second and third arguments must be functions/
```

dalefrancis88: It doesn't hurt to add the ^ and i think being as specific as possible without being limiting is the best call

Issue #201, opened 2018-02-10
Update Writer To Take an Instance instead of a Value


evilsoft:
## Mistakes were Made

When `Writer` was first created, I thought it would be neat to let the `Writer` take care of lifting raw values into the underlying `Monoid`. This created many problems with this Datatype:

* Invalid values for the `Monoid` would throw deep in the `Writer` and out of context from where the value was "lifted". Makes for a nasty stack trace at best, and at worst you share way to much information between the type and your specific domain.
* Lose all power of the extra strength of the `Monoid`. So if an `Array` is used, you can never use the empty as there is no way other than and empty array to have a `void`-ish value.

This will be a breaking change.

dalefrancis88:
Was just reading this, I've yet to use `Writer` so this will obviously add to my confusion but I'm not sure what the update is that needs to be made. Could we add a [Repl](www.repl.it/languages/babel) that shows the issue?

kierans:
I just had this problem with trying to use a `Writer (List a) b`

Passing in a `List` instance result in a list of lists. Not what I wanted and at odds with the documentation.

cf: https://github.com/evilsoft/crocks/blob/master/src/Writer/index.js#L54, https://github.com/evilsoft/crocks/blob/master/src/Writer/index.js#L60

PR #433, opened 2019-07-22
Require Monoid Instance for Log Entry When Constructing A Writer

evilsoft:
## Captains Log
image

This PR is a direct answer to implementing the breaking change as defined in [this issue][issue]. Before this change we provided the value to be wrapped in the `Monoid` and the `Writer` would wrap it for us. This created issues when dealing with empty values on types like `Array` or other structure-ish type Monoids.

Going forward, each log entry will now have to provide an instance of the `Monoid`

```js
const MaxWriter =
  Writer(Max)

const oldWay =
  MaxWriter(34, 'string')

const newWay =
  MaxWriter(Max(34), 'string')
```

[issue]: https://github.com/evilsoft/crocks/issues/201

### DIFF
--- a/src/Writer/Writer.spec.js
+++ b/src/Writer/Writer.spec.js
@@ -58,11 +58,11 @@ test('Writer', t => {
   t.ok(isFunction(Writer.type), 'provides a type function')
   t.ok(isString(Writer['@@type']), 'provides a @@type string')
 
-  const err = /^TypeError: Writer: Must be contructed with both a log entry and a value/
+  const err = /^TypeError: Writer\(Last\): Must be contructed with both a log entry and a value/
   t.throws(f(), err, 'throws with no parameters')
   t.throws(f(0), err, 'throws with one parameter')
 
-  const entryErr = /^TypeError: Writer: Log enrty must be an instance of Last/
+  const entryErr = /^TypeError: Writer\(Last\): Log enrty must be an instance of Last/
   t.throws(f(undefined, 0), entryErr, 'throws with undefined entry')
   t.throws(f(null, 0), entryErr, 'throws with null entry')
   t.throws(f(0, 0), entryErr, 'throws with falsy number entry')
@@ -200,7 +200,7 @@ test('Writer equals properties (Setoid)', t => {
 test('Writer map errors', t => {
   const map = bindFunc(Writer(Last(0), 0).map)
 
-  const err = /^TypeError: Writer.map: Function required/
+  const err = /^TypeError: Writer\(Last\).map: Function required/
   t.throws(map(undefined), err, 'throws with undefined')
   t.throws(map(null), err, 'throws with null')
   t.throws(map(0), err, 'throws with falsey number')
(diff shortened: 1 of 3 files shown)


src/Writer/Writer.spec.js
```diff
@@ -27,7 +27,7 @@ const Writer = _Writer(Last)
 test('Writer construction', t => {
   const w = bindFunc(_Writer)
 
-  const err = /Writer: Monoid required for construction/
+  const err = /^TypeError: Writer: Argument must be a Monoid Constructor/
```
dalefrancis88: Nice :)

src/Writer/index.js
```diff
@@ -20,21 +20,25 @@ const constant = x => () => x
...
     `${_typeString}( ${Monoid['@@type']} )`
 
   function Writer(entry, val) {
     if(arguments.length !== 2) {
-      throw new TypeError('Writer: Log entry and a value required')
+      throw new TypeError(`${_type()}: Must be contructed with both a log entry and a value`)
+    }
+
+    if(!isSameType(Monoid, entry)) {
+      throw new TypeError(`${_type()}: Log enrty must be an instance of ${Monoid.type()}`)
```
dalefrancis88: enrty -> entry

src/Writer/Writer.spec.js
```diff
@@ -46,27 +46,39 @@ test('Writer construction', t => {
...
   t.ok(isFunction(Writer.of), 'provides an of function')
   t.ok(isFunction(Writer.type), 'provides a type function')
   t.ok(isString(Writer['@@type']), 'provides a @@type string')
 
-  const err = /Writer: Log entry and a value required/
+  const err = /^TypeError: Writer\(Last\): Must be contructed with both a log entry and a value/
   t.throws(f(), err, 'throws with no parameters')
   t.throws(f(0), err, 'throws with one parameter')
 
+  const entryErr = /^TypeError: Writer\(Last\): Log enrty must be an instance of Last/
```
dalefrancis88: enrty -> enrty

evilsoft:
@dalefrancis88 all fixed up. Good 👁

src/Writer/index.js
```diff
@@ -20,21 +20,25 @@ const constant = x => () => x
...
   }
 
-  const _of =
-    x => Writer(Monoid.empty().valueOf(), x)
+  const _of = x =>
+    Writer(Monoid.empty(), x)
 
   const _type =
-    constant(`${__type}( ${Monoid.type()} )`)
+    constant(`${__type}(${Monoid.type()})`)
```
dalefrancis88: Not required for review but I'm curious, why make this a func instead of a variable?

dalefrancis88:
image

src/Writer/index.js
evilsoft: Everything except for the sanctuary type is a function. Mostly to keep the API :corn:sistant. But it has the benefit of not accidentally overriding a value that we depend on for most of our type operations. If that makes sense?

PR #434, opened 2019-07-22
Add isMap predicate

### LINKED ISSUE #427: isMap Predicate
**Describe the solution you'd like**
We've got `isArray` and `isPromise`, why not `isMap`?

**Describe alternatives for how you do this now**
implement on my own

**Code**
```js
function isMap(x)
  return x instanceof Map;
}
```

bennypowers:
Closes #427

Notes to reviewers: please advise if I missed any opportunities to document this predicate.


src/predicates/isMap.js
```diff
@@ -0,0 +1,5 @@
+/** @license ISC License (c) copyright 2017 original and current authors */
```
evilsoft: This should be `2019`

src/core/isMap.js
```diff
@@ -0,0 +1,8 @@
+/** @license ISC License (c) copyright 2016 original and current authors */
```
evilsoft: Should be `2019`

bennypowers:
Comments addressed by an amended commit

PR #438, opened 2019-07-28
Micro-optimization to curry

dalefrancis88:
adding in a propery to check if the passed in function has already been run through our curry method

### DIFF
--- a/src/core/curry.js
+++ b/src/core/curry.js
@@ -5,12 +5,15 @@ const isFunction = require('./isFunction')
 
 function applyCurry(fn, arg) {
   if(!isFunction(fn)) { return fn }
+
   return fn.length > 1 ? fn.bind(null, arg) : fn.call(null, arg)
 }
 
 // curry : ((a, b, c) -> d) -> a -> b -> c -> d
 function curry(fn) {
-  return function(...xs) {
+  if(fn.isCurried) { return fn }
+
+  const curried = function(...xs) {
     const args =
       xs.length ? xs : [ undefined ]
 
@@ -28,6 +31,14 @@ function curry(fn) {
 
     return val
   }
+
+  Object.defineProperty(curried, 'isCurried', {
+    enumerable: false,
+    writable: false,
+    value: true
+  })
+
+  return curried
 }
 
 module.exports = curry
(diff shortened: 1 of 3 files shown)


src/core/curry.js
```diff
@@ -5,12 +5,15 @@ const isFunction = require('./isFunction')
...
+
   return fn.length > 1 ? fn.bind(null, arg) : fn.call(null, arg)
 }
 
 // curry : ((a, b, c) -> d) -> a -> b -> c -> d
 function curry(fn) {
-  return function(...xs) {
+  if(fn.isCurried) { return fn }
+
+  const curried = function(...xs) {
```
evilsoft: Could this become a function as opposed to assignment.

src/core/curry.js
```diff
@@ -28,6 +31,14 @@ function curry(fn) {
 
     return val
   }
+
+  Object.defineProperty(curried, 'isCurried', {
```
evilsoft: Lets go with `@@crocks/curried` for the signifier

src/core/curry.js
dalefrancis88: Not sure what you mean by ' function as opposed to assignment', do you mean `function curried(...xs)` ?

evilsoft: yeppers, named function I should of said.

src/core/curry.js
```diff
@@ -5,12 +5,15 @@ const isFunction = require('./isFunction')
...
 function applyCurry(fn, arg) {
   if(!isFunction(fn)) { return fn }
+
   return fn.length > 1 ? fn.bind(null, arg) : fn.call(null, arg)
 }
 
 // curry : ((a, b, c) -> d) -> a -> b -> c -> d
 function curry(fn) {
-  return function(...xs) {
+  if(fn['@@crocks/curried']) { return fn }
```
evilsoft: How about we create a constant for the signal and use a multiline `if` (should probably add that to lint soon)
```js
const CURRY_SYMB =
  '@@crocks/curried'

if(fn[CURRY_SYMB]) {
  return fn
}
```

src/helpers/unary.js
```diff
@@ -2,16 +2,15 @@
 /** @author Ian Hofmann-Hicks (evil) */
 
 const isFunction = require('../core/isFunction')
+const nAry = require('./nAry')
```
evilsoft: Do not think this change belongs in this PR. Thoughts?

PR #442, opened 2019-09-05
Correct project example of Tuple

JacopKane:


JacopKane:
@evilsoft Sure I would be flattered, thanks. Just pushed that as well. Hopefully, I can add more solid contributions in future.

PR #394, opened 2019-04-26
Working on adding bichain

dalefrancis88:
Per the discussion [here](https://github.com/evilsoft/crocks/issues/393), adding in `bichain`

Issue #393, opened 2019-04-25
Chain over the "rejected" side of an `Async`


mgreystone:
I would like to request the ability to chain over the left or "rejected" side of an `Async`. This would be helpful to move values from the left side of the `Async` to the right.

```js
getAsync()
  .chainRej(ifElse(isErrorOk, constant(Async.of('Error Ignored')), Async.Rejected))
```

I'm currently doing this with a helper function.
```js
// chainRej :: (a -> Async e a) -> Async a b -> Async e a|b
export const chainRej = curry((fn, a) =>
  Async((rej, res) => a.fork(x => fn(x).fork(rej, res), res))
)
```

dalefrancis88:
Hey @mgreystone have you checked out [`Coalesce`](https://evilsoft.github.io/crocks/docs/crocks/Async.html#coalesce) or [`Swap`](https://evilsoft.github.io/crocks/docs/crocks/Async.html#swap) I think that they may be more what you're looking for? Coalesce takes two functions one for the left and one for the right to map the values to the same codomain and swap moves left into right and right into left

dalefrancis88:
Product types will always only `map`, `chain` etc over the 'right' value except where the function itself is designed to consider both parts i.e. `coalesce` and `swap. There is also another common method called `bimap` similiar to `swap` but each function could have a different co-domain and they each remain in there respective sides

mgreystone:
Hi @dalefrancis88 . I don't think `coalesce`, `swap`, `map`, `chain` or `bimap` will do what i'm looking for.
* `coalesce` will always move the value from left-to-right, whereas i'm looking to do this conditionally.
* `swap` moves the left to the right & the right to to the left. I'm afraid i'm not sure how i could conditionally move from left-to-right with `swap`.
* `map` only affects the right side
* `chain` is the reverse of what i want. I can move a value from right-to-left.
* `bimap` does not exchange values between sides at all.

evilsoft:
I like it. Think we should figure this one out.
we could add it as a special method on the type instance and then provide an Async specific pointfree function. Would like to see this in on the next patch release. we will be having `0.12.0` coming soon, and we will follow it up in `0.12.1`.
Unless @dalefrancis88 or @karthikiyengar or @bennypowers sees any issue with that. But nothing comes to mind that could be a show stopper for this VERY helpful function.

Then as people request it for other Sum types, we can address, but I would say starting with `Async` will get the best value.

dalefrancis88:
[...]

PR #443, opened 2019-09-08
Add `bichain` to remaining Sum Types

evilsoft:
## 2 chains are better than 1
image

This PR responds to [this issue][issue] and adds `bichain` to the remaining (3) Sum Types as it was already added on to `Async` on [this PR][pr]:
* `Either`
* `Maybe`
* `Result`

For each example in the documentation, we show both just a simple function to swap between "left"/"right" instances and a more in depth example. The exception to this is `Maybe`, which only shows the simple swap example. This is because, there is not really a solid use case that is not better implement with `alt`. Although, if anyone has any suggested, real-world use cases for the `Maybe` example, I am all :ear: 

[issue]: https://github.com/evilsoft/crocks/issues/393
[pr]: https://github.com/evilsoft/crocks/pull/394

### DIFF
--- a/docs/src/pages/docs/crocks/Async.md
+++ b/docs/src/pages/docs/crocks/Async.md
@@ -18,8 +18,8 @@ that can be composed together.
 
 Depending on your needs, an `Async` can be constructed in a variety of ways. The
 typical closely resembles how a `Promise` is constructed with one major
-difference, the arguments used in the function that is passed to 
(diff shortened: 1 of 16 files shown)


evilsoft:
Self Reviewed and gonna merge to get a release out!
image

dalefrancis88:
Apologies for not reviewing this, i thought i did :/

evilsoft:
Oh goodness @dalefrancis88 don't sweat it at all!

Issue #454, opened 2020-01-17
README.md example confusion with `curry`


sunwukonga:
**Is your feature request related to a problem? Please describe.**
Based on my understanding of what `curry` should do, I would expect it to behave like this:
```
const uncurriedAdd = (x, y) = x + y
const curriedAdd = curry( add )
console.log(curriedAdd(20, 4))
// [Function]
```
Just as it would if I did:
```
const properlyCurriedAdd = x => y => x + y
console.log(properlyCurriedAdd(20, 4))
// [Function]
```

**Describe the solution you'd like**
This is very convenient behavior. I'm only requesting to change the README.md to explicitly use the curry'd form first and then explain the convenience.
i.e. 
```
safeDivide(20)(0)
//=> Nothing

// for convenience, the function returned by curry() can also be called with
safeDivide(20, 0)
//=> Nothing
```

evilsoft:
I like this idea!
It can be 🌽fusing, especially if a person is new to doing Functional Programming with JS.
I do not see that this could hurt at all. We could probably also make an example for the docs as well that includes this information: https://crocks.dev/docs/functions/helpers.html#curry

If you want to contribute, please feel free to submit a PR for review, otherwise we can make sure to get the README and docs updated to make it a bit more clear.

sunwukonga:
Yes, of course. I asked because I'd like to contribute.

sunwukonga:
I'm working some examples for the `curry` docs, and I'm having trouble with the type signature of `curriedAdd`.

```
// add :: (Number, Number, Number) => Number
const add = (x, y, z) =>
  x + y + z

// curriedAdd :: (Number, Number, Number) => Number
// curriedAdd :: (Number, Number) => Number => Number
// curriedAdd :: Number => (Number, Number) => Number
// curriedAdd :: Number => Number => Number => Number
const curriedAdd = curry( add )
```
Haskell would not be happy with the redefinition, but beyond that, it would also be unhappy with the change in the number of clauses. I think it's impossible to write the definition of this function in Haskell notation.

sunwukonga:
A four argument curry would have eight type signatures. Five arguments, sixteen. So on and so forth.
Generally, there are `2^(n-1)` type signatures, where `n` is the number of arguments in the function to be curried.

dalefrancis88:
[...]

PR #456, opened 2020-01-18
Edit README (#454)

sunwukonga:
- remove repeated content
- add documentation references
- add extra example


README.md
```diff
@@ -183,7 +103,17 @@ import curry from 'crocks/helpers/curry'
 import isNumber from 'crocks/predicates/isNumber'
 import liftA2 from 'crocks/helpers/liftA2'
 import safe from 'crocks/Maybe/safe'
+```
 
+### Example
+
+Documentation references: [and](https://crocks.dev/docs/functions/logic-functions.html#and),
```
evilsoft: The references should use the reference bracket notation and have the link moved down below.

docs/src/pages/docs/functions/helpers.md
```diff
@@ -283,6 +283,66 @@ that can be called in any combination, such as: `f(x, y, z)`, `f(x)(y)(z)`,
...
+```javascript
+import curry from 'crocks/helpers/curry'
+
+// add :: (Number, Number, Number) -> Number
+const add = (a, b, c) =>
+  a + b + c
+
+// partial application impossible
+add(1)
+// => 1 + undefined + undefined => NaN
```
evilsoft: To keep it 🌽 sistant, these results should just show the result. I do like the comentarty tho. so maybe that should do in a comment below like:

```js
//=> NaN
// 1 + undefined + undefined => NaN
```

docs/src/pages/docs/functions/helpers.md
```diff
@@ -283,6 +283,66 @@ that can be called in any combination, such as: `f(x, y, z)`, `f(x)(y)(z)`,
...
+
+// NOT the only possible type signature as highlighted below
+// curriedAdd :: Number -> Number -> Number -> Number
+const curriedAdd = curry( add )
```
evilsoft: should be in the newline format, no spaces around `()`:
```js
const curriedAdd =
  curry(add)
```

docs/src/pages/docs/functions/helpers.md
```diff
@@ -283,6 +283,66 @@ that can be called in any combination, such as: `f(x, y, z)`, `f(x)(y)(z)`,
...
+
+curriedAdd(1)
```
evilsoft: Would be better to show this as a signature, by assignment

```js
// appliedAdd :: Number -> Number -> Number
const appliedAdd =
  curriedAdd(1)
```

docs/src/pages/docs/functions/helpers.md
```diff
@@ -283,6 +283,66 @@ that can be called in any combination, such as: `f(x, y, z)`, `f(x)(y)(z)`,
...
+// => [Function: curried]
+
+// Number -> Number -> Number -> Number
```
evilsoft: Only put signature on assignments. This should be removed.

docs/src/pages/docs/functions/helpers.md
```diff
@@ -283,6 +283,66 @@ that can be called in any combination, such as: `f(x, y, z)`, `f(x)(y)(z)`,
...
+curriedAdd(1)(2)(3)
+// => 6
```
evilsoft: Extra space between comment and `=>`, should be `\\=>`

docs/src/pages/docs/functions/helpers.md
```diff
@@ -283,6 +283,66 @@ that can be called in any combination, such as: `f(x, y, z)`, `f(x)(y)(z)`,
...
+
+// (Number, Number) -> Number -> Number
```
evilsoft: Should be a comment under the result, maybe using simple variable names instead of types like:
// (a, b)(c)

docs/src/pages/docs/functions/helpers.md
```diff
@@ -283,6 +283,66 @@ that can be called in any combination, such as: `f(x, y, z)`, `f(x)(y)(z)`,
...
+add(1)
+// => 1 + undefined + undefined => NaN
+
+// NOT the only possible type signature as highlighted below
+// curriedAdd :: Number -> Number -> Number -> Number
+const curriedAdd = curry( add )
+
+curriedAdd(1)
+// => [Function: curried]
+
```
evilsoft: Thinking these next bites of code could be "grouped" with a comment explaining the 🌽-venience of partial application using the `curry` function.

docs/src/pages/docs/functions/helpers.md
```diff
@@ -283,6 +283,66 @@ that can be called in any combination, such as: `f(x, y, z)`, `f(x)(y)(z)`,
...
+// (Number, Number, Number) -> Number
+curriedAdd(1, 2, 3)
+// => 6
+
+// optionalAdd :: (Number -> Number -> Number) -> Number
+const optionalAdd = (a, b, c = 2) =>
+  a + b + c
+
+// curriedOptionalAdd :: Number -> Number -> Number
+const curriedOptionalAdd = curry( optionalAdd )
```
evilsoft: use the no space, multi-line as stated above.

docs/src/pages/docs/functions/helpers.md
```diff
@@ -283,6 +283,66 @@ that can be called in any combination, such as: `f(x, y, z)`, `f(x)(y)(z)`,
...
+// Number -> (Number -> Number) -> Number
+curriedAdd(1)(2, 3)
+// => 6
+
+// (Number, Number, Number) -> Number
+curriedAdd(1, 2, 3)
+// => 6
+
+// optionalAdd :: (Number -> Number -> Number) -> Number
+const optionalAdd = (a, b, c = 2) =>
```
evilsoft: I am thinking this is too much information for this. example, maybe just add something to the paragraph above, warning of this case. Seems to cloud the intent of the function here.

README.md
sunwukonga: Understood. Would you like me to convert the existing links (besides the badges at the top), that do not follow this pattern?

docs/src/pages/docs/functions/helpers.md
sunwukonga: This was added in response to a [comment](https://github.com/evilsoft/crocks/issues/454#issuecomment-575936386) made by @dalefrancis88. I'll wait on clarification, it may be that I didn't express his idea well.

docs/src/pages/docs/functions/helpers.md
sunwukonga: > ```js
> const add = curry(
>   x => y => x + y
> )
> ```
> 
> and it will "auto"-curry it to be able to be called like `add(1, 2)`. You cannot do that with things like `ramda`,

This is an uncurry, not a curry. In addition, would you like me to write something like:
> The crocks curry fulfills the role of both `curry` and `uncurry`, returning a function that can be used both as a curried function, an uncurried function, or any combination of argument applications in between.
> ```
> const crocksCurriedFunc =
>   curry((a, b, c, d, e) => {/* something great */})
>
> crocksCurriedFunc(a)(b)(c)(d)(e)  // [Fully curried]
> crocksCurriedFunc(a)(b)(c)(d, e)  // + three permutations
> crocksCurriedFunc(a)(b)(c, d, e)  // + two permutations
> crocksCurriedFunc(a)(b, c, d, e)  // + one permutation, i.e. (a, b, c, d)(e)
> crocksCurriedFunc(a, b, c, d, e)  // [Fully uncurried]
> crocksCurriedFunc(a, b)(c, d, e)  // + one permutation, i.e. (a, b, c)(d, e)
> crocksCurriedFunc(a)(b, c)(d, e)  // + two permutations
> ```
> A five argument crock's curry has sixteen distinct type signatures. In general, a crock's curry with `n` arguments has  type signatures.

README.md
evilsoft: Sure if you wanna.
Unless you feel it will add too much noise to the PR. In that case we can do a PR later, focused on that specifically. Your call totally on that.

docs/src/pages/docs/functions/helpers.md
evilsoft: > This is an uncurry, not a curry. 
👍 

> The crocks curry fulfills the role of both curry and uncurry, returning a function that can be used both as a curried function, an uncurried function, or any combination of argument applications in between.

Do not think we need to go all the way to 5, we do not want to encourage people to make such large function unless they really need to. Thinking just going up 3 would be 👌. But I like

docs/src/pages/docs/functions/helpers.md
evilsoft: Gotcha. Yeah, I think it is good to mention like @dalefrancis88 said.
I think we should frame it more as a warning not to use default args, but it is hard to explain why without launching into some white board discussion on how Function Inputs can be considered Product Types and how Products are nothing more than Limits, and because JS does not use Parametric Polymorphism, we cannot accurately represent a Limit without implementing a case for EVERY single type in JS.

So short of that maybe we can show how to deal with an existing function/method that uses argument defaults instead of showing them creating one from scratch. Kinda steer them away from the idea. And mention that they should be avoid them, without a real reason, other than it is hard to use curry with them.

Then if people ask about it, we can explain it in more depth in future issues or chat room discussions.

docs/src/pages/docs/functions/helpers.md
sunwukonga: > ```js
> // pluck :: String -> [ a ] -> Maybe b
> const pluck = curry(
>   compose(map, prop)
> )
> 
> const data = [
>   { a: 'nice' },
>   { a: 'great', b: 'nice' },
>   { b: 'nice' },
> ]
> 
> pluck('a', data).map(x => `${x}`)
> //=> [ Just 'nice', Just 'great', Nothing ]
> ```

Should I include the signatures for `map`, `prop`, and `compose` and show the contrast of using `compose(map, prop)` directly without the aid of `curry` first?

docs/src/pages/docs/functions/helpers.md
sunwukonga: Can I negotiate to four? :) The case of three is trivial:
```
(a)(b)(c)
(a,b)(c)
(a)(b,c)
(a,b,c)
```
whereas four begins to show the explosion of signatures a little more clearly:
```
(a)(b)(c)(d)
(a)(b)(c,d) -- (a)(b,c)(d) -- (a,b)(c)(d)
(a)(b,c,d) -- (a,b,c)(d)
(a,b,c,d)
(a,b)(c,d)
```

docs/src/pages/docs/functions/helpers.md
sunwukonga: OK. Off the top of my head, I tested `curry`-ing `console.log` and `JSON.stringify` and the results were consistent with my understanding for `console.log`, but strange for `JSON.stringify`.

```
const curriedConsolelog =
  curry(console.log)

curriedConsolelog("Hello ", "me Matey!")
//=> Hello
// As expected, the optional arguments after the first are ignored

const curriedStringify =
  curry(JSON.stringify)

curriedStringify({'head': 1})
//=> [Function curried]
// Unexpected. Apparently JSON.stringify uses another mechanism for optional parameters

curriedStringify({'head': 1}, null, '')
//=> {'head': 1}

curriedStringify({'head': 1})(null)('')
//=> {'head': 1}
```

sunwukonga: Looking at the code for `JSON.stringify` @ [crockford's](https://github.com/douglascrockford/JSON-js/blob/master/json2.js) the `stringify` method doesn't use optional parameters as such. It allows them to be `undefined` and recovers with checks and balances.

dalefrancis88: Apologies for the confusion, my comment was more as @evilsoft said, just a bit of a warning so that it sits in peoples subconscious for when they accidentally run into the issue. 

We've run into that issue with `stringify` and `parse` before, it's quite annoying. The way we get round that is to wrap it with `unary` and then it passes `undefined` to the fake "optional" parameters

docs/src/pages/docs/functions/helpers.md
evilsoft: Nah, should be fine as it is. We can iterate on it, if need be.
We should remove the `.map` bit though, so sorry I thought I removed that.

docs/src/pages/docs/functions/helpers.md
```diff
@@ -280,8 +280,97 @@ Pass this function a function and it will return you a function that can be
...
+between. The returned function has `2^(n-1)` type signatures, where `n` is the
+number of parameters.
+
+```javascript
+import compose from 'crocks/helpers/compose'
+import curry from 'crocks/helpers/curry'
+import map from 'crocks/pointfree/map'
+import prop from 'crocks/maybe/prop'
+
+const crocksCurriedFunc =
```
dalefrancis88: We need to add a signature

```javascript
// crocksCurriedFunc :: Number -> Number -> Number -> Number -> Number
```

docs/src/pages/docs/functions/helpers.md
```diff
@@ -280,8 +280,97 @@ Pass this function a function and it will return you a function that can be
...
+  curry((a, b, c, d) => a + b + c + d)
+
+crocksCurriedFunc(1)(2)(3)(4)  // [Fully curried]
+crocksCurriedFunc(1)(2)(3, 4)
+crocksCurriedFunc(1)(2, 3)(4)
+crocksCurriedFunc(1, 2)(3)(4)
+crocksCurriedFunc(1)(2, 3, 4)
+crocksCurriedFunc(1, 2, 3)(4)
+crocksCurriedFunc(1, 2, 3, 4)  // [Fully uncurried]
+crocksCurriedFunc(1, 2)(3, 4)
```
dalefrancis88: These all need there output shown however given there structure and purpose maybe we can just put the output after the final line

docs/src/pages/docs/functions/helpers.md
```diff
@@ -280,8 +280,97 @@ Pass this function a function and it will return you a function that can be
...
+// add :: (Number, Number, Number) -> Number
+const add = (a, b, c) =>
+  a + b + c
+
+// partial application impossible
+add(1)
+//=> NaN
+// 1 + undefined + undefined => NaN
+
+const curriedAdd =
```
dalefrancis88:
```javascript
// curriedAdd :: Number -> Number -> Number -> Number
```

docs/src/pages/docs/functions/helpers.md
```diff
@@ -280,8 +280,97 @@ Pass this function a function and it will return you a function that can be
...
+//=> 6
+
+// strictCurriedPluck :: String -> [ a ] -> Maybe b
+const strictCurriedPluck =
+  compose(map, prop)
+
+const crockCurriedPluck =
+  curry(strictCurriedPluck)
+
+const data = [
```
dalefrancis88:
```javascript
// data :: [ { a: String, b: String } ]
```

docs/src/pages/docs/functions/helpers.md
```diff
@@ -280,8 +280,97 @@ Pass this function a function and it will return you a function that can be
...
+//=> [ Just 'nice', Just 'great', Nothing ]
+
+```
+
+An important caveat when using `curry` with functions containing optional parameters,
+is that the defaults are applied immediately, reducing the number of partial
+applications to just the number of required parameters. Adding optional parameters
+to functions may not be a good choice if the intention is to use them with `curry`,
+as the ability to change the defaults is lost. Worse yet, any parameter after the
+first optional parameter, without an explicit default value, defaults to `undefined`.
```
dalefrancis88: I personally wouldn't say "worse yet" here. I do get what you're trying to say though. Maybe it could be reworded to say something like `curry only applies the paramaters up until the first optional parameter, from there default values or undefined are passed in`

docs/src/pages/docs/functions/helpers.md
```diff
@@ -280,8 +280,97 @@ Pass this function a function and it will return you a function that can be
...
+  curry(strictCurriedPluck)
+
+const data = [
+  { a: 'nice' },
+  { a: 'great', b: 'nice' },
+  { b: 'nice' }
+]
+
+strictCurriedPluck('a')(data).map(x => `${x}`)
+//=> [ Just 'nice', Just 'great', Nothing ]
```
dalefrancis88: While we use single quotes for input strings we you double quotes in the outputs

docs/src/pages/docs/functions/helpers.md
```diff
@@ -280,8 +280,97 @@ Pass this function a function and it will return you a function that can be
...
+
+crockCurriedPluck('a', data).map(x => `${x}`)
+//=> [ Just 'nice', Just 'great', Nothing ]
```
dalefrancis88: Same here

docs/src/pages/docs/functions/helpers.md
```diff
@@ -280,8 +280,97 @@ Pass this function a function and it will return you a function that can be
...
+number of parameters.
+
+```javascript
+import compose from 'crocks/helpers/compose'
+import curry from 'crocks/helpers/curry'
+import map from 'crocks/pointfree/map'
+import prop from 'crocks/maybe/prop'
+
+const crocksCurriedFunc =
+  curry((a, b, c, d) => a + b + c + d)
```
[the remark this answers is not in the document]
sunwukonga: I've seen the light on this. There's no need for it at all. I rename the smaller `add` function below to `crocksCurriedAdd` and delete this one.

docs/src/pages/docs/functions/helpers.md
sunwukonga: I can put that in. Despite what you said on the issue I raised, I am really resisting labeling function with a single type signature that actually has four. For the original definition of `curry` I understand the reasoning, but I'm making one last plea that it shouldn't be repeated too often when it isn't accurate. My feeling is that coder's in general, and functional coder's in particular, will assume the definition is perfectly correct and that they're looking at it wrong.

README.md
```diff
@@ -183,7 +103,17 @@ import curry from 'crocks/helpers/curry'
...
+
+Documentation references: [and][and-logic-crocks],
+[curry][curry-helpers-crocks],
+[predicates (isNumber)][predicate-docs],
+[liftA2][lifta2-helpers-crocks],
+[safe][safe-maybe-crocks].
+
+```javascript
 // divide :: Number -> Number
 const divide =
```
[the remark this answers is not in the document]
sunwukonga: What about the one below:
```
const safeDivide = curry(
   (x, y) => liftA2(divide, safeNumber(x), notZero(y))
)
```
Should it be change to:
```
const safeDivide = 
  curry((x, y) => liftA2(divide, safeNumber(x), notZero(y)))
```
Or:
```
const safeDivide = 
  curry((x, y) =>
    liftA2(divide, safeNumber(x), notZero(y))
  )
```

docs/src/pages/docs/functions/helpers.md
evilsoft: I get that, mostly we are just matching the way we represent these handy auto-curry functions in JS. If you look at libs like Ramda, Fluture, Sanctuary (because of their ties to Ramda) they use this same signature to represent an auto-curried function.

I agree with you, other people coming from Haskell, Purescript, etc they may be a little unnerved by this. But people coming from FP from a JS side of things are already familiar with this short hand convention.

README.md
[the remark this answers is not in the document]
evilsoft: It is good as is.
separates the body (function) from the modifier (curry)

docs/src/pages/docs/functions/helpers.md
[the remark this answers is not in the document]
evilsoft: @dalefrancis88 
I like to use whitespace to separate things. If we are modifiing a function with curry and the function is a one-liner, that I think something like:
```js
const crocksCurriedFunc = curry(
  (a, b, c, d) => a + b + c + d
)
```
keeps the "body" from the "modifier"

docs/src/pages/docs/functions/helpers.md
sunwukonga: I went with this:
> Parameters are applied up until the first optional parameter, at which point subsequent parameters either receive the declared default or go undefined.

docs/src/pages/docs/functions/helpers.md
sunwukonga: I put it in. I really want to put a caveat on it, but I resisted the urge.

docs/src/pages/docs/functions/helpers.md
sunwukonga: Without the map, the proper output is gigantic. The output comment below the call to pluck would be figurative, not literal. Are you sure?

evilsoft: The console output is not what is displayed in those results (`\\=>`), but a representation of the result.
We would have to put that on every result, so it is best to leave that off and just go with:

```js
pluck('a', data)
//=>  [ Just "nice", Just "great", Nothing ]
```

sunwukonga: OK

docs/src/pages/docs/functions/helpers.md
dalefrancis88: So something we've been playing with the idea for a while now is having a section that's articles/tip & tricks/examples/blogs etc on the help file to guide people in starting out with crocks and potentially FP. Maybe when we have this section ready you could write something in more detail that really outlines everything in a way that shines the light on how to use it

dalefrancis88:
Just want to leave a message to let you know this has not been forgotten about, I'll get a final review and squirrel on this shortly

README.md
```diff
@@ -183,10 +103,20 @@ import curry from 'crocks/helpers/curry'
...
+### Example
 
+Documentation references: [and][and-logic-crocks],
+[curry][curry-helpers-crocks],
+[predicates (isNumber)][predicate-docs],
+[liftA2][lifta2-helpers-crocks],
+[safe][safe-maybe-crocks].
+
+```javascript
 // divide :: Number -> Number
```
dalefrancis88: This signature is wrong, but lets not block this PR for it

dalefrancis88:
Thanks for doing this @sunwukonga I think it's going to really remove some of the confusion and create a great reference for us when the issue pops up

README.md
sunwukonga: I missed that, sorry.

The `safeDivide` signature is wrong too.

PR #465, opened 2020-03-16
Update Deps and Remove Dev Support for Node:8

evilsoft:
# So Secure
image

Just a simple PR to get our dev stack up to date and remove a vulnerability. The updates to `husky`, `lint-staged` and `sinon` removed support for node:8, so it makes sense to move on and only test against versions `10` and `12` of node. node:8` is coming to an end on its life cycle anyway.

This should clear up the dependabot vulnerability when merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: node_js
 node_js:
   - '12'
   - '10'
-  - '8'
 
 install:
   - npm install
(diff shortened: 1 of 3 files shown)


evilsoft:
Just going to squash and merge this in.
Wanna clear up all that vulnerability.
image

Issue #105, opened 2017-03-07
Allow `String` and `Array` constructors anywhere a Monoid is accepted


evilsoft:
So there is no reason we cannot just accept `String` and `Array` JS Constructors as `Monoid`. They will need to be implemented in the following places:

- [ ] `mconcat`
- [ ] `mconcatMap`
- [ ] `mreduce`
- [ ] `mreduceMap`
- [ ] `Writer` constructor

dalefrancis88:
Hey, There is a small issue with `Array` having two constructors, one of which is when taking a number this'll  define the length of the `Array`. 

I have an idea but not sure if it's the best, what are your thoughts? 

https://repl.it/@DaleFrancis/String-Array-Monoid

evilsoft:
Hmmm. You may wanna wait on this until I finish up the changes I am doing to `Writer`, have some ideas about getting the empty for JS Types.

Also this core function will need to be changes to remove the `foldWith` and the `compose` we should do it in place to avoid the addition of stack frames.  Was thinking something like:
```js
const fl = require('crocks/core/flNames')

// comming soon in core
const primatives = T => ({
  ['String']: () => '',
  ['Array']: () => []
})[T.name]

// export coming soon in core
const getEmpty = T =>
  T[fl.empty] || T.empty || primatives(T)

// core implementation, all validation happens in implementing functions
// no errors or composition or anything here.
function _mconcatMap(Rep, fn, xs) {
  const empty = getEmpty(Rep)

  const M = Rep === Array
    ? x => [ x ]
    : Rep

  return xs.reduce(
    (acc, x) => acc.concat(M(fn(x))),
    empty()
  )
}

// Example
const { List, curry } = require('crocks')

const mconcatMap =
  curry(_mconcatMap)

const data =
  List.fromArray([ 1, 3, 5, 6 ])

const mapListToArray = curry(
  fn => mconcatMap(Array, fn)
)
```

evilsoft:
And I think we may have to dispatch `concat` to one of the core files, string or array. We want to put an error in the case of something like:
```js
mconcat(String, [ 1, 2, 3, 4 ])
//=> '1234'
```
it should throw an error instead of doing the default concat because of Semigroup laws. they need to be instances of the same type.

dalefrancis88:
Cool we seem to be on the same page here, albeit yours being nicer, i'm glad that's the case. 

Happy to take this on once the new things are in core

PR #468, opened 2020-04-04
Add Writer docs, minor corrections to Pair docs

karthikiyengar:
Long overdue PR, very happy to strike this off my to do list! Please take a look guys.

### DIFF
--- a/docs/src/pages/docs/crocks/Pair.md
+++ b/docs/src/pages/docs/crocks/Pair.md
@@ -153,11 +153,12 @@ equals(Pair([ 1, 2 ], ''), [ 1, 2 ])
 Semigroup s, t => Pair s t ~> Pair s t -> Pair s t
 ```
 
-When both underlying values of a given `Pair` are fixed to a `Semigroup`,
-`concat` can be used to concatenate another `Pair` instance with underlying
-`Semigroup`s of the same type and structure. Expecting a `Maybe` wrapping a
-`Semigroup` of the same type, `concat` will give back a new `Pair` instance
-wrapping the result of combining the underlying `Semigroup` instances.
+When both underlying values of a given `Pair` are fixed to
+a `Semigroup`, `concat` can be used to concatenate another `Pair` instance
+with underlying `Semigroup`s of the same type and structure. Expecting
+a `Maybe` wrapping a `Semigroup` of the same type, `concat` will give back a
+new `Pair` instance wrapping the result of combining the
+underlying `Semigroup` instances.
 
 ```javascript
 import Pair from 'crocks/Pair'
@@ -365,16 +366,17 @@ Semigroup s => Pair s a ~> (a -> Pair s b) -> Pair s b
 ```
 
 Combining a sequential series of transformations that allows for custom
-accumulation in addition to transforming a value. `chain` requires a `Pair`
-returning function that contains a `Semigroup` in its first position. As an
-additional requirement, is that instances of the same `Semigroup` must occupy
-the first position of the source `Pair` and the `Pair` returned by the function.
+accumulation in addition to transforming a value. `chain` requires
+a `Pair` returning function that contains a `Semigroup` in its first position.
+An additional requirement, is that instances of the same `Semigroup` must
+occupy the first position of the source `Pair` and the `Pair` returned by the
+function.
 
 ```javascript
-const Pair = require('crocks/Pair')
+import Pair from 'crocks/Pair'
 
-const setProp = require('crocks/helpers/setProp')
-const omit = require('crocks/helpers/omit')
+import setProp from 'crocks/helpers/setProp'
+import omit from 'crocks/helpers/omit'
 
 // addTmp :: (String, a, Object) -> Pair [ String ] Object
 const addTmp = (key, value, x) =>
@@ -704,7 +706,7 @@ for `merge`.
 This method comes in handy when using a `Pair` as a means to run parallel
 computations and combine their results into a final answer. Typically this
 method works hand in hand with the either the [`branch`](#branch) or
-[`fanout`][fanout] helper functions.
+[`fanout`](#fanout) helper functions.
 
 ```javascript
 import Sum from 'crocks/Sum'
@@ -1032,7 +1034,7 @@ a function will be returned that takes a given value and returns an `Pair`.
 import Pair from 'crocks/Pair'
 import Sum from 'crocks/Sum'
 import Writer from 'crocks/Writer'
-import fanout from 'crocks/helpers/fanout'
+import fanout from 'crocks/Pair/fanout'
 
 import writerToPair from 'crocks/Pair/writerToPair'
 
--- a/docs/src/pages/docs/crocks/Writer.md
+++ b/docs/src/pages/docs/crocks/Writer.md
@@ -11,10 +11,11 @@ weight: 160
 Monoid m => Writer m a
 ```
 
-The `Writer` monad makes it possible to represent and compose computations that produce some 
-additional data along with their computed values. `Writer` is a `Product` type that consists 
-of a `Monoid` as the first value that typically holds metadata, while the 
-second value holds the result of a computation. 
+The `Writer` monad makes it possible to represent and compose computations 
+that produce some additional data along with their computed values. `Writer` 
+is a `Product` type that consists of a `Monoid` as the first value that 
+typically holds metadata, while the second value holds the result of a 
+computation. 
 
 <article id="topic-implements">
 
@@ -31,7 +32,9 @@ second value holds the result of a computation.
 Writer :: MonoidTypeRep M, Monoid m => M -> (m, a) -> Writer m a
 ```
 
-`Writer` is a type constructor that takes a `Monoid` constructor and returns a new constructor. The new constructor is a binary function that accepts a `Monoid` instance and a value `a`, and returns a `Writer` instance.
+`Writer` is a type constructor that takes a `Monoid` constructor and returns 
+a new constructor. The new constructor is a binary function that accepts a 
+`Monoid` instance and a value `a`, and returns a `Writer` instance.
 
 ```javascript
 import Writer from 'crocks/Writer'
@@ -58,7 +61,9 @@ StringWriter('adding inputs', (a, b) => a + b)
 Writer.of :: a -> Writer m a
 ```
 
-`of` is a construction helper that can be used to construct an `Writer` with any given value, with an empty instance of the `Monoid` that was passed to the type constructor.
+`of` is a construction helper that can be used to construct an `Writer` with 
+any given value, with an empty instance of the `Monoid` that was passed to 
+the type constructor.
 
 ```javascript
 import Writer from 'crocks/Writer'
@@ -84,7 +89,12 @@ ListWriter.of(42)
 Writer m a ~> b -> Boolean
 ```
 
-Used to compare the underlying values of two `Writer` instances for equality by value, `equals` takes any given argument and returns `true` if the passed arguments are a `Writer` with an underlying value are equal to the underlying value of the `Writer` the method is being called on. If the passed argument is not a `Writer` instance or if the underlying values are not equal, `equals` will return false.
+Used to compare the underlying values of two `Writer` instances for equality 
+by value, `equals` takes any given argument and returns `true` if the passed 
+argument is a `Writer` with an underlying value equal to the underlying 
+value of the `Writer` the method is being called on. If the passed argument is 
+not a `Writer` instance or if the underlying values are not equal, `equals` 
+will return false.
 
 ```javascript
 import Assign from 'crocks/Assign'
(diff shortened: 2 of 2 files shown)


karthikiyengar:
@evilsoft - I just dropped a commit that makes the docs comply with 0.13. The docs assume that #201 and #105 are merged.

dalefrancis88:
@karthikiyengar i think you were right in pointing this to 0.13

karthikiyengar:
@dalefrancis88 - There we go.

docs/src/pages/docs/crocks/Pair.md
```diff
@@ -1032,7 +1034,7 @@ a function will be returned that takes a given value and returns an `Pair`.
 import Pair from 'crocks/Pair'
 import Sum from 'crocks/Sum'
 import Writer from 'crocks/Writer'
-import fanout from 'crocks/helpers/fanout'
+import fanout from 'crocks/Pair/fanout'
```
dalefrancis88, line 1037: nice catch

docs/src/pages/docs/crocks/Writer.md
```diff
@@ -0,0 +1,352 @@
...
+`Functor`, `Apply`, `Chain`, `Applicative`, `Monad`
+
+</article>
+
+<article id="topic-construction">
+
+## Construction
+
+```haskell
+Writer :: MonoidTypeRep M, Monoid m => M -> (m, a) -> Writer m a
```
dalefrancis88, line 31: What is MonoidTypeRep? i've never seen this notation before

docs/src/pages/docs/crocks/Writer.md
```diff
@@ -0,0 +1,352 @@
...
+```
+
+`Writer` is a type constructor that takes a `Monoid` constructor and returns a new constructor. The new constructor is a binary function that accepts a `Monoid` instance and a value `a`, and returns a `Writer` instance.
```
dalefrancis88, line 34: The lines should be a max of 80 characters

docs/src/pages/docs/crocks/Writer.md
```diff
@@ -0,0 +1,352 @@
...
+
+## Constructor Methods
+
+#### of 
+
+```haskell
+Writer.of :: a -> Writer m a
+```
+
+`of` is a construction helper that can be used to construct an `Writer` with any given value, with an empty instance of the `Monoid` that was passed to the type constructor.
```
dalefrancis88, line 61: Same here

docs/src/pages/docs/crocks/Writer.md
```diff
@@ -0,0 +1,352 @@
...
+
+## Instance Methods
+
+#### equals
+
+```haskell
+Writer m a ~> b -> Boolean
+```
+
+Used to compare the underlying values of two `Writer` instances for equality by value, `equals` takes any given argument and returns `true` if the passed arguments are a `Writer` with an underlying value are equal to the underlying value of the `Writer` the method is being called on. If the passed argument is not a `Writer` instance or if the underlying values are not equal, `equals` will return false.
```
dalefrancis88, line 87: Same here

dalefrancis88, line 87: this is equality by strict comparison, is that what you're saying in the first part?

> ...n underlying value are equal to the...

I think there is an extra 'are' here

docs/src/pages/docs/crocks/Writer.md
```diff
@@ -0,0 +1,352 @@
...
+  .equals(AssignWriter(Assign({ a: 30 }), null))
+//=> true
+
+Writer(String)('here is 42', { value: 42 })
+  .equals(Writer(String)('also 42', { value: 42 }))
+//=> true
+
+Writer(String)('hello', 25)
+  .equals(Writer(String)('hello', 26))
+//=> false
```
dalefrancis88, line 112: You made an` AssignWriter` for consitency should we have a `StringWriter` as well

docs/src/pages/docs/crocks/Writer.md
```diff
@@ -0,0 +1,352 @@
...
+
+equals(Writer(Array).of('result'), 'result')
+//=> false
+```
+
+#### map
+
+Used to apply transformations to the value that a `Writer` holds, `map` takes a function that it will lift into the context of the `Writer` and apply to it the wrapped value.
```
dalefrancis88, line 120: same here

docs/src/pages/docs/crocks/Writer.md
```diff
@@ -0,0 +1,352 @@
...
+#### map
+
+Used to apply transformations to the value that a `Writer` holds, `map` takes a function that it will lift into the context of the `Writer` and apply to it the wrapped value. 
+
+```javascript
+import Sum from 'crocks/Sum'
+import Writer from 'crocks/Writer'
+
+// toLower :: String -> String
+const toLower = x => x.toLowerCase()
```
dalefrancis88, line 127: the for functions is to have the parameters on the same line as the function name and the body following

docs/src/pages/docs/crocks/Writer.md
```diff
@@ -0,0 +1,352 @@
...
+//=> Writer( Sum 0 "hello!" )
+
+Writer(String)('file was loaded', 'important text from document')
+  .map(contents => contents.length)
+//=> Writer("file was loaded"  28)
+```
+
+#### ap
+
+Short for apply, `ap` is used to apply a `Writer` instance containing a value to another `Writer` instance that contains a function, resulting in new `Writer` instance comprising of the result of concatenating the monoid instances, and the value derived from applying the function. `ap` requires that it is called on a writer that wraps a curried polyadic function as its value.
```
dalefrancis88, line 140: is polyadic the right word here? i had to look it up so not sure either way, just checking

docs/src/pages/docs/crocks/Writer.md
```diff
@@ -0,0 +1,352 @@
...
+const add = curry((x, y) => x + y)
+
+// addPercentage :: Number -> Number -> Number
+const addPercentage = curry((percentage, amount) => amount + percentage * amount)
+
+// includeVatWithLog :: Number -> Writer [Number]
+const includeVatWithLog = (amount) => LogWriter([ 'adding vat' ], addPercentage(0.19, amount))
+
+// includeVatWithLog :: Number -> Writer [Number]
+const includeExpressShippingWithLog = (amount) => LogWriter([ 'adding express shipping' ], add(3.99, amount))
```
dalefrancis88, line 176: I don't personally mind these on one line but for consistencies sake should we mulitline them? maybe @evilsoft can weigh in here. i don't want to continue something just because "that's how we've always done it" but i also think the multi-line makes it easier for a FP newbie to follow

docs/src/pages/docs/crocks/Writer.md
karthikiyengar: I'm not entirely sure that's a bad pattern, I just left it in to demonstrate that it's possible to use it this way.

docs/src/pages/docs/crocks/Writer.md
karthikiyengar: Basically refers to the constructor of a monoid instance. You can find a similar use case in `ReaderT`.

docs/src/pages/docs/crocks/Writer.md
karthikiyengar: Nice catch.

karthikiyengar: Nice catch.

docs/src/pages/docs/crocks/Writer.md
karthikiyengar: It's been consistently used across other docs and it's descriptive imho.

docs/src/pages/docs/crocks/Writer.md
karthikiyengar: I really don't intend to be blunt, but it's really unproductive and demoralizing to have comments in most PRs about line lengths, how a function looks and the like. 

I wouldn't mind fixing these for this PR, but if it's really a project level constraint, it would be great to have this enforced by lint rules, or not at all.

docs/src/pages/docs/crocks/Writer.md
dalefrancis88: Sounds good to me, i just hadn't seen it

docs/src/pages/docs/crocks/Writer.md
dalefrancis88: I agree, i hate it myself. That's why i loved it when i seen you making the new rule. I would prefer that the main focus of a PR is about the value add not the formatting of it

fabfianda: Guys, love the whole Crocks library (using Async in production a lot).

I noticed that on line 175 the signature has not been updated to match the following function (// includeVatWithLog ... vs //includeExpressShippingWithLog).

Also on line 159 and following I think I spotted a typo... "dobuleAttack" should be "doubleAttack" yes? 

Thanks! Cannot wait to add Writer to the toolbox.

docs/src/pages/docs/crocks/Writer.md
evilsoft: We want to break up those lines. Makes code review easier in a lot of cases.

docs/src/pages/docs/crocks/Writer.md
evilsoft: Yeah, I feel you, we need to figure out a way to communicate things.
To avoid this, until we can come up with a way to communicate/enforce these styles/formatting bit. I can go in after and make those changes. things like the multiline and using `()` around single item. That feels a little shifty to me though, as if I am going back and redoing work, so I am open to other suggestions.

The importance of it, is the styling is 🌽sistant between docs. Or as 🌽sistant as we can get it. 
The reason for the multi line definition is for mobile and tablet views, it sucks to have to scroll over to see the definition and then go back and forth to do the composition in your head. 

So there are reasons for them, but you are right, without some way to define that rule in eslint, it can make these demoralizing.

evilsoft: But there is value in the formatting, especially with documentation.
How information is presented in documentation does make an impact on the cognitive load required of the Learner and to account for a11y and responsive 🌽cerns.

In normal code I would say that could be debated, but for documentation it is really important to think about things like screen width, spacing between expressions, not adding unnecessary characters (like () on single arguments), or the like. This is mostly a failing on my part for not having this a documented enough.

docs/src/pages/docs/crocks/Writer.md
karthikiyengar: What do you think of https://eslint.org/docs/rules/max-len? It does the job pretty well, for files including markdown.

karthikiyengar: P.S: I ran a lint check with this rule activated across the codebase, and it flags around 2500 cases. So I'm a bit lost here. IMHO, this open-to-interpretation nature of this requirement is not ideal.

docs/src/pages/docs/crocks/Writer.md
karthikiyengar: If this is a hard requirement, could we please consider enabling https://eslint.org/docs/rules/implicit-arrow-linebreak and fixing this across the codebase?

docs/src/pages/docs/crocks/Writer.md
karthikiyengar: @fabfianda - Thanks for the corrections! Definitely missed that.

docs/src/pages/docs/crocks/Writer.md
evilsoft: Thinking, at the least, in the documentation examples.

docs/src/pages/docs/crocks/Writer.md
evilsoft: Really just care about the markdown in the docs/src/pages folder for these block descriptions.

dalefrancis88:
@karthikiyengar is this ready for another review? just want to make sure it's not lost

karthikiyengar:
@dalefrancis88 - I've addressed comments from the previous review.

docs/src/pages/docs/crocks/Writer.md
```diff
@@ -0,0 +1,384 @@
...
+## Construction
+
+```haskell
+Writer :: MonoidTypeRep M, Monoid m => M -> (m, a) -> Writer m a
+```
+
+`Writer` is a type constructor that takes a `Monoid` constructor and returns 
+a new constructor. The new constructor is a binary function that accepts a 
+`Monoid` instance and a value `a`, and returns a `Writer` instance.
+
```
dalefrancis88, line 38: @evilsoft can comment on this but i believe there is an issue starting a line with `this`, not going to block this for that but just putting a note here to get teh feedback

PR #478, opened 2020-05-24
Allow fn passed to Async.fromNode to be partially applied

JamieDixon:
`Async.fromNode` returns a function that eventually resolves to an `Async` type once it has been applied to all of its parameters.

This PR is a first-pass at allowing the function returned from `Async.fromNode` to be partially applied.

Dependent on https://github.com/evilsoft/crocks/pull/477


JamieDixon:
@evilsoft @dalefrancis88 Just letting you know that this one is now ready for review 🙂

dalefrancis88:
Thanks mate, i'll get some 👀on this tomorrow or overmorrow

src/Async/index.js
```diff
@@ -48,14 +48,22 @@ function fromNode(fn, ctx) {
...
-    Async((reject, resolve) => {
+  const _fn = curry(fn)
+
+  return (...args) => {
+
+    if (args.length < _fn.length - 1) {
+      return fromNode(_fn(...args), ctx)
+    }
+
+    return Async((reject, resolve) => {
```
dalefrancis88, line 59: @evilsoft I know there is the issue with `Async` swallowing errors in one particular scenario, for this change, will wrapping this in the `curry` and the recursive call to `fromNode` create a similiar scenario?

src/Async/index.js
```diff
@@ -48,14 +48,22 @@ function fromNode(fn, ctx) {
...
 
-  return (...args) =>
-    Async((reject, resolve) => {
+  const _fn = curry(fn)
+
+  return (...args) => {
+
+    if (args.length < _fn.length - 1) {
+      return fromNode(_fn(...args), ctx)
+    }
```
dalefrancis88, line 57: Per your documentation and our discussions their is a chance that the given function has a length of 0. Would this be a case to return an error?

docs/src/pages/docs/crocks/Async.md
```diff
@@ -379,8 +379,17 @@ As such, the need for binding may arise. `fromNode` provides a second, optional
...
+The function returned from `fromNode` will be automatically curried, allowing you
+to partially apply the function up to its penultimate parameter<sup>[1]</sup>, so long as the 
+arity of the original function given to `fromNode` can be determined via its `.length` property.
+
+In practice this means that functions defined via `compose` or those making use
+of `arguments`, spread args (`...args`) or default values for parameters, will not be 
+good candidates for partial application.
+
+<sup>[1]</sup> The final parameter to the incoming function is provided to you by
+`fromNode` rather than being part of the parameters that can be partially applied.
```
dalefrancis88, line 391: We should avoid starting and ending lines with a `

src/Async/index.js
JamieDixon: My understanding at this point is that if the function has a length of 0 then the safest thing to do is assume that it isn't being partially applied, and apply all of the parameters as per usual. This will mean that any code that currently passes a `compose`d or `...args` or `argument` function will continue to work in exactly the same way as it does today.

docs/src/pages/docs/crocks/Async.md
JamieDixon: Thanks! Fixed 😃

src/Async/index.js
dalefrancis88: That's true, sounds good to me

src/Async/index.js
evilsoft: Nah, we should be good with this.
The Async "issue" is actually the desired behavior when dealing with `Promises`. And it is a behavior of Promises, so we should be good here!

PR #479, opened 2020-05-25
Upgrade all the things

evilsoft:
image
Noticed we are a little out of date on a few development dependencies. This PR just upgrades them all in anticipation for an upcoming release.

We have not fully upgraded `tape` to the `5.0.0` version yet, we will need to work on redoing all `plan` based specs for that. BUT they are maintaining `4.x`, so that is super awesome!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,22 +14,23 @@
       }
     },
     "@babel/core": {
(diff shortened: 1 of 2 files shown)


evilsoft:
Gonna merge this on in.
image

PR #486, opened 2020-07-16
Upgrading packaged to remove vulnerability in Lodash and bringing some up to date

dalefrancis88:
Package update time!


dalefrancis88:
@evilsoft i can't self approve PR's are you able to chuck one on and we can get this out? it may be worth getting a release out too

PR #492, opened 2020-09-04
Correct getPropOr type annotation

jakubbarczyk:


dalefrancis88:
I know we discussed this on gitter, but actually i'm wondering if `getPathOr` has the wrong sig
`a -> (String | Integer) -> b -> a` i think is that way because the first param is a default value for the property, b is the object you pulling from and then the final value is the value of that property or the default
so an argument could be made that the default and the value are from the same set a, correct?

With `getPathOr` having a sig of `a -> [ (String | Integer) ] -> b -> c` really the only way for that to work would be that if `a` was considered always subset of `c`

What are your thoughts @evilsoft?

evilsoft:
> With getPathOr having a sig of a -> [ (String | Integer) ] -> b -> c really the only way for that to work would be that if a was considered always subset of c

That is :100: correct, `getPathOr` is the faulty siggy here. For exactly the reasons you specified! Good :eyes: !

